Issue #5, opened 2020-05-29
Glitching border on FAQ page


EricLambrecht:
This is only a very minor CSS bug that is easy to fix. In fact I already have a fix prepared that I can open a PR for.

## Where to find the issue
See bottom border of each closed FAQ item on https://www.coronawarn.app/de/faq/.
It seems to happen in every browser.

## Describe the issue
![cwa_css_bug1](https://user-images.githubusercontent.com/1144699/83295061-ad4cc780-a1ee-11ea-9f2b-26c29e6f7644.png)

See screenshot above: the bottom border in the inactive state of each item is not styled correctly.

## Suggested change
Changing two lines of CSS. I can create a PR for a fix of this issue. 🙂

tkowark:
Great catch! Thank you very much for spotting that glitch and we'd be more than happy to accept your PR.

MalteJ:
Thank you!
I have merged your PR and it will be deployed via GitHub Actions within the next few minutes :)

EricLambrecht:
Sure! Only a minor thing, but I'm glad I could help :)

PR #7, opened 2020-05-29
Fix accordion body border glitch

EricLambrecht:
Fixes issue #5 

Before:
cwa_css_bug1

After:
cwa_css_bug2


MalteJ:
Thanks a lot!

PR #8, opened 2020-05-29
align package.json with the repository license

CCFenner:
The repository is flagged to be licensed under `Apache v2`, but the license in the `package.json` is set to `MIT`. Although the package is not published, this looks quite odd and should be fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "build": "gulp build --production"
   },
   "author": "ZURB <foundation@zurb.com>",
-  "license": "Apache 2.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "build": "gulp build --production"
   },
   "author": "ZURB <foundation@zurb.com>",
-  "license": "MIT",
+  "license": "Apache 2.0",
```
MalteJ, line 11: Shouldn't it be "Apache-2.0"?

Issue #6, opened 2020-05-21
Website lacks many Security- and Privacy Features


TomTeeJay:
Disclaimer: I am not sure if the website issues for https://www.coronawarn.app should be post here. So please feel free to forward this to whom is responsible for)

Though the website claims to be privacy-friendly, there are still these issues:

- Unsafe TLS Encryption Configuration
  TLS 1.3 is not enabled, TLS 1.0 is still active (!)
  Please check Ciphers aswell.
- CSP (Content Security Policy) is missing
- "no-referrer" tag is missing. Referers are told to 3rd parties.
- Several X-Headers (X-Content-Type, X-Frame, X-XSS-Protection) are missing

Please use common test suites to check website for security and privacy:

- https://observatory.mozilla.org/
- https://webbkoll.dataskydd.net/de
- https://www.ssllabs.com/ssltest/

Julian-B90:
The following resource can´t be found
https://www.coronawarn.app/assets/css/ajax-loader.gif

<img width="451" alt="Bildschirmfoto 2020-06-02 um 22 07 39" src="https://user-images.githubusercontent.com/4377396/83564817-a3dc9b80-a51d-11ea-9a7b-09c944b70459.png">

tkowark:
@Julian-B90 I'll create a new issue from your last comment

SebastianWolf-SAP:
Sorry that it took a while, but finally all mentioned issues should be fixed. See also the results on:
- https://observatory.mozilla.org/analyze/coronawarn.app
- https://webbkoll.dataskydd.net/de/results?url=http%3A%2F%2Fcoronawarn.app
- https://www.ssllabs.com/ssltest/analyze.html?d=coronawarn.app
- https://www.ssllabs.com/ssltest/analyze.html?d=www.coronawarn.app

Mit freundlichen Grüßen/Best regards, 
SW
Corona Warn-App Open Source Team

PR #9, opened 2020-05-30
display Mozillas Obersovatory status

CCFenner:
This change adds a badge for Mozilla's Observatory status to display the website's security status discussed in #6.

Bildschirmfoto 2020-05-30 um 07 10 53

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,10 @@
 </p>
 
 <hr />
+
+<p align="center">
+    <a href="https://observatory.mozilla.org/analyze/coronawarn.app" title="Latest Results"><img src="https://img.shields.io/mozilla-observatory/grade/coronawarn.app" alt="Mozilla HTTP Observatory Grade"></a>
+</p>
 <p align="center">
     <a href="#about-this-repository">About this Repository</a> • 
     <a href="#development">Development</a> •


MalteJ:
Sorry, but we do not include any assets loaded from external sources.

MalteJ:
By the way, we are not vulnerable to XSS as this is a static website and no authorized requests are made against a backend. So the security rate "D+" is not justified for this setup.

MalteJ:
Ups, just realized this badge will be displayed on the GitHub Readme, only, and not on the website itself. So I'll reopen this PR. Still the rating "D+" is not justified and has to be adjusted.

uniquoo:
> By the way, we are not vulnerable to XSS as this is a static website and no authorized requests are made against a backend. So the security rate "D+" is not justified for this setup.
Just because you're serving a statically rendered website does not mean that X-(Content | Frame | XSS)-Headers (sh | c)ould be skipped. Because endusers do not know in which way the website will receive additions in functionality these headers should be placed either way. So D+ rating is fine at this point.

SebastianWolf-SAP:
Well, we already accepted #6 so it was just a question of time and configuration to get this fixed. I would still say they could honor meta tags, but don't need to be nitpicking. Now we proudly show our A+ :)

Mit freundlichen Grüßen/Best regards, 
SW
Corona Warn-App Open Source Team

PR #10, opened 2020-05-30
Fix repetition of DP-3T in German index site

daniel0611:


SebastianWolf-SAP:
Good catch, thanks! :)

PR #20, opened 2020-05-31
Paragraph breaks added in faq_de.json

jostschmithals:
1. Suggestion to split up the text in the first accordions with additional paragraph breaks (in the same way as it has already been done in the last accordion), to increase the readability, especially in mobile view.
2. Removed the only `<br>` in one paragraph.

### DIFF
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -22,22 +22,29 @@
                 },{
                     "title" : "Welche konkreten Ziele verfolgt die Bundesregierung mit der Entwicklung einer Kontaktverfolgungs-App?",
                     "textblock": [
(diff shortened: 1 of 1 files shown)


src/data/faq_de.json
```diff
@@ -77,7 +88,7 @@
                     "title": "Wird die Smartphone-Batterie nicht schnell leerlaufen mit der App im Hintergrund?",
                     "textblock": [
                         "Die Corona-Warn-App nutzt die vorhandene Bluetooth-Technologie der Endgeräte. Durch die Einbeziehung der Exposure Notification API von Apple und Google kann die App die energieeffiziente Bluetooth-Technologie BLE (Bluetooth Low Energy) verwendet werden."
-					]
+          ]
```
Bananeweizen, line 91: indentation seems wrong here

jostschmithals: Thanks! - Should be fixed now,

SebastianWolf-SAP:
Sorry for the late reply. As there was a conflict in the meantime and as the English version was missing, we included the fix in an own commit. 

Thanks for reporting! We will close this issue now.

Mit freundlichen Grüßen/Best regards, 
SW
Corona Warn-App Open Source Team

PR #37, opened 2020-06-07
Avoid the word "person" in connection with cwa-server 

mh-:
Avoid the (incorrect) impression that cwa-server keeps a list of infected _persons_.


mh-:
> Could you please add the same fix for the English version?

The English version "If confirmed, the server adds the user’s keys to the SARS-CoV-2 confirmed list, which is regularly broadcasted to all apps." is already ok IMHO.

PR #45, opened 2020-06-11
Disable browser caching

ambiti0us0x01:
adds Cache-Control Header to prevent HTTP/1.1 request caching


MalteJ:
Why do you want to disable this?

ambiti0us0x01:
Good practice, especially if sensitive data are handeld. The browser will not store them.

hangy:
> Good practice, especially if sensitive data are handeld. The browser will not store them.

Is there any handling of user data in the website? I think it's just a static info site.

tkowark:
Indeed, the page contains no user data at all. Hence, we'll leave the caching enabled

PR #46, opened 2020-06-13
Redirect to german page when lang_de is set

thomas-hoer:
The inline script is nonsence because of the Content-Security-Policy. It prints out a warning and redirects to the english version via meta at any time. 
`Refused to apply inline style because it violates the following Content Security Policy directive: "default-src 'self' *.coronawarn.app". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.`
The javascript did not work on any of my browsers. But of curse, I neither use Safari nor Internet Explorer

### DIFF
--- a/src/assets/js/redirect.js
+++ b/src/assets/js/redirect.js
@@ -1,0 +1,6 @@
+if((navigator.language || navigator.userLanguage).toLowerCase().indexOf('de') >= 0) {
+  window.location.href = "./de/";
+}
+else {
+  window.location.href = "./en/";
+}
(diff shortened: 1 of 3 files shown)


tkowark:
Thank you very much for spotting the issue and providing the fix, @thomas-hoer !

PR #74, opened 2020-06-16
Use navigator.languages for redirect

hangy:
Improves support for browsers that may request another language before German or English.

### DIFF
--- a/src/assets/js/static/redirect.js
+++ b/src/assets/js/static/redirect.js
@@ -1,23 +1,21 @@
-var href = "./en/";
+var lang = "en";
+
 if (
   typeof navigator.languages !== "undefined" &&
   navigator.languages.length > 0
 ) {
   for (var i = 0; i < navigator.languages.length; ++i) {
-    var lowerCaseValue = navigator.languages[i].toLowerCase();
-    if (lowerCaseValue.indexOf("de") >= 0) {
-      href = "./de/";
-      break;
-    } else if (lowerCaseValue.indexOf("en") >= 0) {
-      href = "./en/";
+    var tag = navigator.languages[i].slice(0, 2).toLowerCase();
+    if (tag === "de" || tag === "en") {
+      lang = tag;
       break;
     }
   }
 } else if (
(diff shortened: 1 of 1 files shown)


src/assets/js/static/redirect.js
```diff
@@ -1,6 +1,23 @@
-if((navigator.language || navigator.userLanguage).toLowerCase().indexOf('de') >= 0) {
-  window.location.href = "./de/";
+var href = "./en/";
+if (
+  typeof navigator.languages !== "undefined" &&
+  navigator.languages.length > 0
+) {
+  for (var i = 0; i < navigator.languages.length; ++i) {
+    var lowerCaseValue = navigator.languages[i].toLowerCase();
```
mroderick, line 7: The `lowerCaseValue` will look like `de-DE`, `en-US`, `en-GB`, `de`, `en`. These values are [`BCP 47` language tags](https://tools.ietf.org/html/bcp47).

What we will be looking for is `en` or `de` in the [Primary Language Subtag](https://tools.ietf.org/html/bcp47#section-2.2.1), not anywhere else in the string.

Therefore, we can look at just the first two characters and see if they match our supported languages

I'd suggest trimming the string, to only contain the first two characters. Then you can simplify the body of the loop with something simpler that uses `===` instead of `>=` comparisons

hangy: Thanks for your review. You're absolutely correct that the language tag should be easier to match. I had actually checked this before, and noticed that the `navigator.languages` property, which was checked as the preferred value in the original version, is also [defined as BCP 47](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language#Value). In either case, slicing/trimming the individual value should be a more precise match, as the user might only have `de-DE` enabled, but not `de`.

hangy: @mroderick I've updated the PR accordingly. I think this new version is a bit simpler in general.

tkowark:
Thanks a lot, that looks great

PR #77, opened 2020-06-16
Fix typos

llamadeus:
Replaced "Reproducable" with "Reproducible" and added a missing "e".


OlMue:
Well spotted, thank you!

PR #80, opened 2020-06-16
Update packages

MartijnCuppens:
Package updates & minor tweaks:

### Commit the `package-lock.json` file

This file should be committed, as mentioned by npm when `npm install` is executed. The lockfile prevents devs from installing different packages during install.

### Package updates

- All packages are up to date now
- Packages that weren't used are removed (protip: use [depcheck](https://www.npmjs.com/package/depcheck) to detect those packages)

### README edits

- `npm` is installed with nodejs, no reason to mention that as 2 separate things.
- It's better to use the gulp version which is included in the `node_modules` folder since you don't know if the globally installed gulp is compatible with the project. npm scripts can than be used to run gulp
- Browsersync already spins up a http server, no reason to mention `http-server`.

### Run `npm ci` instead of `npm install`

- See https://docs.npmjs.com/cli/ci.html


README.md
```diff
@@ -25,29 +25,32 @@ This repository contains the source files of the official website for the Corona
...
-After a successful build, you'll have a new folder ```public``` in the repository's base directory. It contains the generated files for the complete website. To test the generated content, simply use a local web server such as http-server by executing the command 
+After a successful build, you'll have a new folder `public` in the repository's base directory. It contains the generated files for the complete website.
+
+To test the generated content, simply use a local web server such as http-server by executing the command
 
 ```bash
-  http-server
+npm start
 ```
-It will automatically use ```public``` as base directory, go to ```localhost:8080``` to view the website.
```
ingomc, line 44: isnt it `:8000`

gulpfile.js
```diff
@@ -5,14 +5,11 @@ const yargs = require('yargs');
 const browser = require('browser-sync');
 const gulp = require('gulp');
 const panini = require('panini');
-//const rimraf = require('rimraf');
```
ingomc, line 8: if you remove the import, you should also delete the commented out use of it in build task.
But before every Build, the dist Folder should be completely wiped

README.md
MartijnCuppens: Indeed, fixed that

gulpfile.js
MartijnCuppens: I just removed the comment (see `//`)

> But before every Build, the dist Folder should be completely wiped

This is not the case at this moment, this might need to be tackled in another PR?

ingomc: ok for me

MartijnCuppens: Moved the issue to https://github.com/corona-warn-app/cwa-website/issues/98

tkowark:
Thank you very much for these fixes!

PR #91, opened 2020-06-17
Correcting comment

ingomc:
It should be config, since it actually loaded from config.yml not settings.yml which is confusing


tkowark:
Thanks for spotting :)

PR #96, opened 2020-06-17
Remove redundant whitespace in bash syntax

zdroid:
This way commands can be copy-pasted easily (without selecting the whitespace).


tkowark:
Thank you for the contribution. With accepting #80 , these changes unfortunately became obsolete.

PR #103, opened 2020-06-17
Unfold FAQ accordion on deeplink by default

### LINKED ISSUE #100: FAQ deeplinks: Answer does not unfold by default
Deeplinks were added yesterday to the FAQs as requested in #71 

Unfortunately, when one clicks on a deeplink, it doesn't unfold the answer, which is disorienting to a reader who expects to see the answer and not have to find the correct question and click on the answer manually. This is clearly unexpected behaviour, thus a bug.

davidbucka:
Resolves #100


tkowark:
Thanks a lot for this fix!

PR #105, opened 2020-06-18
tiny wording improvement in "Can I be identified as COVID-19-positive…

sneumann:
… by network traffic monitoring"

Hi, people might get worried by the words "false notifications", 
so my addition mentions that they have no consequences on the server side. 
The diff is a bit difficult to grasp first sight, I added:

... and sending false notifications *that will be discarded on the server side*. 
and
... and upload of *real* infection data cannot be distinguished from this background noise. 

Yours, Steffen


OlMue:
Good point, thanks very much!

PR #107, opened 2020-06-18
Fix typos in FAQ & Privacy Statement

christiankrug:
Fixed four small typos in English texts.


src/data/faq.json
```diff
@@ -107,7 +107,7 @@
                         "The further development of the Corona-Warn-App will also be closely accompanied by scientific research."
                     ]
                 },{
-                    "title" : "What specific constraints has the German government imposed on the contractors, SAP and Deutsche Telekom, for the development?",
+                    "title" : "What specific constraints have the German government imposed on the contractors, SAP and Deutsche Telekom, for the development?",
```
OlMue, line 110: "government" is a collective noun, but I'd say that it takes the singular in this case. The government acted as a unity when deciding about this, so there's no need to use the plural. The plural would be better if we wanted to emphasize the individual members of the government, which isn't the case here.
Source: [https://www.dailywritingtips.com/collective-nouns-singular-or-plural/](https://www.dailywritingtips.com/collective-nouns-singular-or-plural/) (which cites several established style guides on this topic)

christiankrug: You are right, good explanation. It has been reverted.

OlMue:
Excellent, thank you!

PR #118, opened 2020-06-21
Remove jQuery and slickSlider

maurer2:
Hello, it looks like jQuery is only used for some class toggling and slickSlider. In order to reduce the page weight those things could probably be done via classList, addEventListener, queryselector and a vanillaJS slider plugin like glideJS (https://glidejs.com/). Additionally document.ready could also be replaced with a small library like domready https://github.com/ded/domready).
Here is a quick test for the vanillaJS slider part.


mojoaxel:
@maurer2  I think reducing the footprint of this page is a very good idea! Maybe you could cleanup/split this pull-request!? AT the moment it is hard to review.

tkowark:
Thanks for the proposal. The multitude of (suggested) changes makes it rather tough to review, though. 

Also, we're a bit hesitant to simply replace multiple libraries (at once) without knowing the concrete benefits (e.g., in kB saved on the wire) as we have to weigh these benefits against the additional efforts in maintenance.

We will therefore close this PR but would, of course, accept incremental improvements to the page in the future.

maurer2:
Okay, makes sense. I will create a pr with non-slider stuff converted to vanillaJS as a intermediate step.

PR #119, opened 2020-06-21
Remove unnecessary font types

maurer2:
Hello, eot, ttf and svg-fonts are not really necessary anymore, since woff is supported in IE9+ and woff2 in evergreen browsers, so those can probably be removed. More info about browser support here (https://css-tricks.com/snippets/css/using-font-face/). I also added a woff2 version of the icon font.


maurer2:
Updated to latest master.

Incognito:
It would be good to know the browser visit statistics for the website before approving. Is there a sizable chunk of visitors that this impacts?

maurer2:
Browserlist is set to IE9+, although layout is slightly restricted in IE9 due to the flexbox grid. It should be fine in most major browser that got released in the last and this decade, aside from Chrome 4, Safari 4.3 and the stock Android 4.3 browser.

maurer2:
No, Browserlist is set to IE9+, which supports woff1.

Incognito:
Sure, there's a list. But real living people in Germany don't match a text file that developers find convenient. The entire purpose is to get as many people as possible to download the app, not snub someone who doesn't even know what it means to be 2 versions behind.

maurer2:
Okay, makes sense.

Incognito:
😞 my comment was "It would be good to know the browser visit statistics for the website before approving", not "don't do this".

maurer2:
Don't worry. It's only like 10 lines of code, that would be removed. Not really worth the effort to be honest, especially since the existing code still works and changes would probably only save a few bytes.

PR #139, opened 2020-06-29
Adding of WebP support with PNG/JPEG as Fallback

tkowark:
Reverts corona-warn-app/cwa-website#136.

Now that libpng-dev is added, this should work.


tkowark:
Thanks again to @dannylinden !

PR #150, opened 2020-07-02
accessibility: increase color contrast ratio to >7

mojoaxel:
_this fixes #93_ 
and increeses the lighthouse score!

Lighthouse requires a color contrast ration of >7. The current value is 3.38.
I'll changed the clolor `blue` to be above 7 (Level AAA) - and still look good :wink:. 

see [TR/WCAG21#contrast-minimum](https://www.w3.org/TR/WCAG21/#contrast-minimum)

image


tkowark:
Looks ok and fixes the contrast ratio. Thanks a lot!

PR #151, opened 2020-07-03
Update FAQ (EN/DE): days active

MarlisFriedl:
Update link to country app stores
days_active_Apple
days_active_Android


src/data/faq_de.json
```diff
@@ -11,12 +11,19 @@
...
...tauscht und die Risiko-Ermittlung wird weiterhin durchgeführt. Es gehen keine Daten verloren. Technisch gesehen handelt es sich hier um einen Rundungsfehler, der auftritt, wenn die Anzahl aktiver Tage der Risiko-Ermittlung nicht mehr zunimmt, sondern statisch bei 14 Tagen bleibt. Ab diesem Zeitpunkt wird die Zeit, in der die Risiko-Ermittlung in irgendeiner Form nicht aktiv war, von den 14 Tagen abgezogen und die Zahl der aktiven Tage entsprechend abgerundet.", "Mögliche Ursachen für eine Deaktivierung der Risiko-Ermittlung:". "<ol><li>Bluetooth wurden deaktiviert, auch wenn es nur ein kurzer Zeitraum ist</li><li>Die Hintergrundaktualisierung für die App war nicht aktiv, so dass nicht immer eine Risiko-Ermittlung durchgeführt wurde</li><li>Der Flugmodus wurde eingeschaltet, auch wenn es nur über einen kurzen Zeitraum war</li><li>Das Smartphone war ausgeschaltet</li><li>Das Smartphone wurde neu gestartet</li></ol>", "Der Fehler ist in iOS behoben.", "Detaillierte Informationen zu diesem Thema gibt es auf Github: <a href='https://github.com/corona-warn-app/cwa-app-android/issues/796' target='_blank' rel='noopener noreferrer'>github.com/corona-warn-app/cwa-app-android/issues/796</a>."
```
MichaelPFrey: The anchor is "days_active_Apple". The linked github issue is https://github.com/corona-warn-app/cwa-app-android/issues/796 from the android-app. I do understand that this is from a UI perspective and core-root perspective it is the same issue, but on an implementation level and from deploying a fixed version to end-users, the issues are and have to traced separately, so linking to https://github.com/corona-warn-app/cwa-app-ios/issues/804 or https://github.com/corona-warn-app/cwa-app-ios/issues/815 would be more helpful to iOS users. (The iOS and Android issues can be/are allready cross linked via github)

_wheter or not link from the faq to github is a separate question, to which I do not want to voice an opinion eitherway_

tkowark: Thanks for spotting. What we really wanted to link to, is this comment: https://github.com/corona-warn-app/cwa-app-android/issues/796#issuecomment-653061512 , which explains in great details how the number of active days is being calculated. We'll think about whether we find a better home for this info, but for the time being, I'll adopt the link

PR #185, opened 2020-07-14
feat: replaced running-app.png with a svg image

mojoaxel:
I replaced the largest image (`running-app.png`) with a SVG alternative.

**This reduces the footprint by ~55KB (almost 10% af the whole page).** :rocket: 

I also replaced the "source-code" of the image with some [real source-code from the cwa-server](https://github.com/corona-warn-app/cwa-server/blob/master/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/DiagnosisKeyBundler.java#L158-L180). To me it feels like an improvement!?

If that's a problem let me know than I try to use the original "source code".

### Old
image

### New
image


tkowark:
I really like this PR, especially the touch with the real code.

The only thing I would like to change is to increase the font size by 2 points. The code is barely readable right now. Maybe we can find a method with shorter lines where this increase would not cause line breaks?

mojoaxel:
> The only thing I would like to change is to increase the font size by 2 points.

I updated the font-size and the code. (see above).

tkowark:
Thanks a lot, looks very good to me :)

PR #241, opened 2020-07-31
Better Search

tkowark:
Instead of matching the entire string, now every word is looked up and if they all appear in the answer, it is returned.

Previous: "Google log" would yield no result
Now:"Google log" yields all FAQ entries containing `google` and `log`

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -97,10 +97,19 @@ $(document).ready(function(){
             let text = faq[anchor];
             let anchorDiv = "div[id='" + anchor + "-div']"
             // text and header does not match or the search string is empty
-            if(searchString.length === 0 || text.search(searchString) !== -1) {
+            if(searchString.length === 0) {
                 show.push(anchorDiv);
             } else {
-                hide.push(anchorDiv);
+                // remove double spaces, split by space, and filter out empty strings
+                let searchTerms = searchString.replace(/ +(?= )/g,'').split(" ").filter(x => x.length > 0);
+                // perform the search for each term
+                let allHits = searchTerms.every(term => text.search(term) !== -1);
+                // put to hide, if no hit was found
+                if(!allHits) {
+                    hide.push(anchorDiv);
+                } else {
+                    show.push(anchorDiv);
+                }
             }
         });
 


UGatgithub:
Didn‘t know where the right place is to report this issue, so I send it by mail. Thanks for picking up so quickly. Will check once again when the change is implemented on the website.

tkowark:
The change should be live noe. Maybe try reloading the page a couple of times, but merged PRs are published immediately.

tkowark:
Also, if there are further issues with the search, just open an issue in this repo, so it doesn't get lost in the mailbox

UGatgithub:
Strange, I reloaded multiple times the DE and EN pages but no success. Some examples...
https://www.coronawarn.app/de/faq/?search=Google
https://www.coronawarn.app/en/faq/?search=Apple
https://www.coronawarn.app/de/faq/?search=Apple+iOS

UGatgithub:
Found Type forward search works and the issue happening when the user presses return ...

tkowark:
ahhh, found the issue. the problem is case sensitivity. Fix will be available, soon.

UGatgithub:
I can confirm it works with small letters :-)

tkowark:
And with https://github.com/corona-warn-app/cwa-website/pull/243 it will also work with uppercase searches. When I introduced the url query parameter, I forgot to downcase the search string (the faq.json is all lowercase).

tkowark:
Fix is now live. Thanks again for testing!

UGatgithub:
Works well better now, hanks for this fix!

I understand results are different due to DE / EN FAQ contents but maybe there is a possibility to improve e.g. by joining results from both languages for some important search words like: log, test or key which are also getting used in German. Currently there are significant more results for these words searching FAQ in English.

UGatgithub:
Hallo Thomas,
fehlender Buchstabe im letzten Absatz, am Ende des vorletzten Wortes steht „... neue Ideen!“ statt „... neuen Ideen!“

URL: https://www.coronawarn.app/de/community/

IST: Wir freuen uns auf alle neue Ideen!
SOLL: Wir freuen uns auf alle neuen Ideen!

Gruß Uwe

Sent using the 1&1 mail app

tkowark:
Danke für den Hinweis! Ist jetzt gefixed. Wenn Dir noch was ähnliches auffällt, dann bitte als neuen Issue oder vielleicht gleich als Pull Request. Kommentare auf geschlossene Issues/PRs sieht man meist nur noch durch Zufall.

PR #286, opened 2020-08-14
Legal notice ios android

maxmarkus:
- for german and english pages


.github/workflows/build.yml
```diff
@@ -7,6 +7,8 @@ name: CI
 on:
   push:
     branches: [ '**', '!master' ]
+  pull_request:
+    branches: [ '**', '!master' ]
```
tkowark, line 11: Github Actions do not permit to build external Pull Requests, as it could allow you to, e.g., add a print statement somewhere to get access to website secrets, etc.

PR #288, opened 2020-08-16
Adding Statistic Feature

SEAL369:
### **Feature description**
1. Since this open source project is related to corona cases in Germany, therefore an initiative is made to link the visitors of the website to the world corona statistic information. 

### **Problem and motivation**
1. In order to save visitors’ time, the new added feature could allow visitors to access and read the latest information of corona cases such as confirmed case, recovered cases and death cases all over the world with just a simple click. Apart from general information, it is also providing the data, statistic and research regarding corona virus from big organization such as WHO. With this information, people can be aware of the pandemic and make a good decision when they are planning to travel to other country.

### **Is this something you're interested in working on**
1. Yes, because I found that this feature is useful for visitors.

### DIFF
--- a/src/partials/global-nav.html
+++ b/src/partials/global-nav.html
@@ -17,7 +17,9 @@
         {{#each language.nav}}
         {{> navlist-item item=.}}
         {{/each}}
+        
         <li class="nav-item lang">
+          <a href="https://www.google.com/search?client=firefox-b-d&q=world+corona+statistic" style="color:rgb(51, 51, 146);text-decoration: underline;"> Corona Virus Statistic</a>
           <a {{#if
             lang_de}}href="/en/{{#ifblogdetail}}blog/{{/ifblogdetail}}{{#if
             @root.page-name_en}}{{@root.page-name_en}}{{else}}{{#if


src/partials/global-nav.html
```diff
@@ -17,7 +17,9 @@
         {{#each language.nav}}
         {{> navlist-item item=.}}
         {{/each}}
+        
         <li class="nav-item lang">
+          <a href="https://www.google.com/search?client=firefox-b-d&q=world+corona+statistic" style="color:rgb(51, 51, 146);text-decoration: underline;"> Corona Virus Statistic</a>
```
MichaelPFrey, line 22: Focusing on the link target and not the idea as such:
- please remove add least the `client=firefox-b-d` - this does not belong here
- why link to google search, when google search is sourcing its data from wikipedia?
Bildschirmfoto von 2020-08-16 16-54-26
- why not link to the RKI? (e.g. [Fallzahlen](https://www.rki.de/DE/Content/InfAZ/N/Neuartiges_Coronavirus/Fallzahlen.htmll) or [COVID-19-Dashboard](https://corona.rki.de/))

As you can see in your own screenshot:
- the color you are setting has little contrast to the background

PR #296, opened 2020-08-19
Mitigate wrong links in app error message

tkowark:
It seems the URL for the error message of the 39508 error is wrong within the app (https://github.com/corona-warn-app/cwa-app-android/issues/1021).
This PR **mitigates** that by taking the wrong link and redirecting to the proper FAQ entry. 

## Disclaimer!
This should not be a long-living solution. If the link in the app is fixed, remove the two `(39508).html` files (I'll create a separate issue as a reminder). You can keep the faq-redirect.js and faq-redirect.html files just in case ^^

### DIFF
--- a/src/assets/js/static/faq-redirect.js
+++ b/src/assets/js/static/faq-redirect.js
@@ -1,0 +1,6 @@
+// get the target faq anchor from the script attributes
+const targetFaq = document.currentScript.getAttribute('target-faq');
+// get the current url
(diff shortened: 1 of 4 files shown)


tkowark:
Argh, this does not work with the current server config. The server does not even try to find a (39508).html file when the last part of the url is (39508) ...

tkowark:
Next update: works now after setting the mime type on the FAQ documents manually

PR #303, opened 2020-08-20
Extending FAQ entry for exposure check log

tkowark:


tkowark:
This closes https://github.com/corona-warn-app/cwa-documentation/issues/409

Issue #317, opened 2020-08-26
Improve 9002 (Timeout)


tkowark:
## Feature description

@vaubaehn noted some improvements for the FAQ entry he created: https://github.com/corona-warn-app/cwa-website/pull/308#issuecomment-680300008

OlMue:
Closed with #318

PR #308, opened 2020-08-23
Add entry 'Cause 9002: Something went wrong (timeout)' as requested in #998

vaubaehn:
@GPclips kindly [invited to create a PR](https://github.com/corona-warn-app/cwa-app-android/issues/998#issuecomment-678142777) to add an entry to the FAQ concerning error message 'Cause 9002: Something went wrong. timeout' as discussed in #998.

Thanks to the support of @Jo-Achim and @Gladdi causes of this exception could have been identified and resolutions can be provided.

As requested, I added the entry to section "id": "notification", just below 'Cause 9002: ... (sqlite). You might consider to move the new entry down to section "resolved", as not too many users are affected and also possible resolutions are already provided. Instead, I'd suggest to move up 'Cause 9002: ... (sqlite)' to section 'common problems', as many users are affected from this exception.

If you find the text fitting to your needs, you'd also be able to use the already provided German translation.
Otherwise, feel free to adjust everything according to your needs.

Additional fix: In the English version of the FAQ, for entry notifications: Cause 9002 (sqlite) the 'Android/Google' tag is missing in the title. I complemented it, according to the corresponding German entry.

Closing remark: 'timeout' is written in small letters, according to the original error message. It is suggested to not use a capital letter at the beginning of the word, in order to avoid (even more) confusion, if at a later point in time another 'Cause: 9002 - Timed Out waiting for 60000 ms' might be added to the FAQ.

PR #318, opened 2020-08-26
Improve FAQ cause9002 timeout

OlMue:
According to #317 and #308


vaubaehn:
Hi @OlMue and @tkowark , now it looks very nice! :+1:  Thanks very much again!

PR #325, opened 2020-08-27
Added smooth scrolling to anchor tag when clicking anchor link

SpyderScript:
I added an scrolling animation when clicking an anchor link.

## Anchor links
image


tkowark:
I really like that behavior as it makes the page much less "jumpy". Thanks a lot for that addition.

PR #338, opened 2020-08-28
fixed regexp errors

vitkolos:
Searching a query with one parenthesis led to an uncaught unterminated parenthetical SyntaxError, as JavaScript translates the parameter to RegExp.
> If a non-RegExp object regexp is passed, it is implicitly converted to a RegExp with new RegExp(regexp). [(source)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search)
However, this disables the possibility of using RegExp to filter the items.


tkowark (approved):

tkowark:
Thanks a lot for spotting and fixing that error!

vitkolos:
You're welcome!

Issue #339, opened 2020-08-29
FAQ log limit of 100 entries


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/faq/#exposure_check

## Describe the issue

In 1. of the FAQ entry [My exposure log shows checks for less than 14 days](https://www.coronawarn.app/en/faq/#exposure_check) it says:

"The log has a fixed limit of 100 entries. If there are more entries, only the latest are kept. We're in contact with Apple and Google so that the complete log will be displayed."

Android users have reported that the limit has been removed. See the closed issue [Exposure Checks only show 100 entries in the log (Android)](https://github.com/corona-warn-app/cwa-app-android/issues/941) https://github.com/corona-warn-app/cwa-app-android/issues/941 from @daimpi. @harambasicluka wrote in https://github.com/corona-warn-app/cwa-app-android/issues/941#issuecomment-674089365 that the issue had been brought up (presumably with Google and Apple), however we didn't hear about any feedback from the team. There is no mention of the issue in the [Google API for Exposure Notifications Release Notes](https://developers.google.com/android/exposure-notifications/release-notes) dated 2020-08-19.

## Suggested change
If the fixed limit of 100 entries has been reliably removed then the FAQ can be changed accordingly.

mtb77:
@MikeMcC399 thanks for reporting this. As far as I can see the limit has been changed for Android but iOS is still having this limit. 

I suggest to modify the FAQ entry from:

> The log has a fixed limit of 100 entries. If there are more entries, only the latest are kept. We're in contact with Apple and Google so that the complete log will be displayed.

to

> The log has a fixed limit of 100 entries on iOS Devices. If there are more entries, only the latest are kept. We're in contact with Apple so that the complete log will be displayed. The log limit of 100 entries on Android Devices has been recently removed by Google since ENF Version 16203302004.

@daimpi can you confirm the ENF version?

Thanks, 

Sascha
Corona-Warn-App Open Source Team

MikeMcC399:
@mtb77 
Thank you for picking up this issue!

> I suggest to modify the FAQ entry to
> 
> > The log has a fixed limit of 100 entries on iOS Devices. If there are more entries, only the latest are kept. We're in contact with Apple so that the complete log will be displayed. The log limit of 100 entries on Android Devices has been recently removed by Google since ENF Version 16203302004.

[...]

PR #342, opened 2020-08-30
Updated the FAQ according to #339 

mtb77:
as Android devices are now exporting more then 100 entries since end of August.


MarlisFriedl (commented):
see also text suggestion in / update also https://github.com/corona-warn-app/cwa-faq/issues/337 accordingly

MikeMcC399:
@MarlisFriedl 
> see also text suggestion in / update also [corona-warn-app/cwa-faq#337](https://github.com/corona-warn-app/cwa-faq/issues/337) accordingly

The link https://github.com/corona-warn-app/cwa-faq/issues/337 doesn't work for me (404 error). Maybe it's a private repository?

MikeMcC399:
Regarding the proposed text change:
"The log limit of 100 entries on Android devices has been recently removed by Google with a configuration change at the end of August 2020."

I'm not convinced that the limit was brought about by a configuration change, however I am unable to prove otherwise! Given that we don't really know,  I would remove the text "with a configuration change" making the text instead:

"The log limit of 100 entries on Android devices has been recently removed by Google at the end of August 2020."

If Google reveals at some time how they made the change, the FAQ entry could be updated accordingly.

MarlisFriedl:
@MikeMcC399 Hi Mike, yes you're right. We use it to prepare updates that we see coming. All changes are then done only publicly in cwa-website.

PR #369, opened 2020-09-13
Added information about QR code scanning

mtb77:
closes: #345


MikeMcC399:
I was asked in https://github.com/corona-warn-app/cwa-website/issues/345 to review this PR.

I think one person got into a problem by deleting a previously entered QR and then trying to re-enter it. That it is special more detailed error case. I don't know if that should be mentioned or not.

Otherwise https://github.com/corona-warn-app/cwa-website/pull/369/commits/4ff18a41f3cee28d2c87b41060175472d26b12bb looks very good to me as far as it stands. 


Thanks!

PR #381, opened 2020-09-21
Removed iOS 14 beta reference

mtb77:


mtb77:
@loki-cui removed the entry completely. Need to add it again with iOS 15 then :D

PR #386, opened 2020-09-23
new pdf

HannaHei:


vaubaehn (commented):
Hi @HannaHei ,
Nice information, thank you. The graph in the middle of the page 'Übermittelte Testergebnisse' lacks information on whether data is cumulative or they're daily numbers. Think, would be good to add to provide proper information. Thank you!

vaubaehn:
@HannaHei
One small addition: bottom graph, 'Testergebnisse bereit zum Teilen' : in the explanation "57% teilen..." you are using present tense, which suggests that this is a stable (continous) number. As this percentage is subject to change in future, past tense ("teilten") is more appropriate; other verbs for the two bullets accordingly.

PR #393, opened 2020-09-24
Updated battery information

mtb77:


mtb77:
@loki-cui @svengabr @tkowark please also check the last commit regarding the change on the start page (referencing iOS 14). I dont wanted to open a second PR for just this change. Thanks!

loki-cui (approved):

PR #617, opened 2020-11-25
Move FAQ entry reg. ExposureDetectionIsAlreadyRunning to solved section

Ein-Tim:
This PR moves the FAQ entry regarding the ExposureDetectionIsAlreadyRunning Error to the solved section and ads a comment to inform that this Issue has been solved with 1.7.1.


Ein-Tim:
I've added a commit which updates the current app version in gulpfile to keep them up to date. Hope this is ok with you 👍

loki-cui (approved):

MikeMcC399:
According to @d4rken rollout of CWA Android 1.7.0 has been interrupted and it will shortly be replaced by a hotfix rollout of CWA Android 1.7.1.

Possibly it would be better to revert the commit https://github.com/corona-warn-app/cwa-website/pull/617/commits/e03b04db1164ae82561830e104f4297cab9da4ce to change the latest iOS and Android versions in the [gulpfile.js](https://github.com/corona-warn-app/cwa-website/blob/master/gulpfile.js) and let the FAQ article update proceed independently of the exact version updates.

Ein-Tim:
Done, thanks for the hint @MikeMcC399. 
BTW what is the reason for the hotifx?

MikeMcC399:
@Ein-Tim 
> BTW what is the reason for the hotifx?

There are two hotfixes and these are listed in https://github.com/corona-warn-app/cwa-app-android/releases/tag/v1.7.1-RC1.

Issue #630, opened 2020-11-27
FAQ: Out-of-date references to current app version (1.5.x or 1.6.x)


MikeMcC399:
## Where to find the issue

FAQ Link
-- |
https://www.coronawarn.app/en/faq/#no_risk_update_ios
https://www.coronawarn.app/en/faq/#API39508
https://www.coronawarn.app/en/faq/#cause9002_timeout
https://www.coronawarn.app/en/faq/#cause9002
https://www.coronawarn.app/en/faq/#app_crash

and corresponding articles under https://www.coronawarn.app/de/faq/ for the German language.

## Describe the issue

Several articles have outdated references to earlier releases of CWA (1.5.x or 1.6.x) as being the current version, whereas the current version is now 1.7.x.

FAQ Link | Current text (partial excerpt)
-- | --
https://www.coronawarn.app/en/faq/#no_risk_update_ios | The latest version of the Corona   Warn App is currently 1.6.1.
https://www.coronawarn.app/en/faq/#API39508 | This bug has been fixed with the   currently available version 1.5 of the Corona Warn App.
https://www.coronawarn.app/en/faq/#cause9002_timeout | This bug has been fixed with the   currently available version 1.5 of the Corona Warn App.
https://www.coronawarn.app/en/faq/#cause9002 | This bug has been fixed with the   currently available version 1.5 of the Corona Warn App.
https://www.coronawarn.app/en/faq/#app_crash | This bug has been fixed with the   currently available version 1.5 of the Corona Warn App.

## Suggested change
Remove the assignment of latest or currently available version from the naming of the versions containing the relevant fixes.

Let me know if you would like a PR for these changes and I will be happy to submit.

heinezen:
Fixed in #631

MikeMcC399:
https://www.coronawarn.app/en/faq/#no_risk_update_ios still says
"The latest version of the Corona Warn App is currently **1.6.1**."

although the latest iOS version shown on https://apps.apple.com/de/app/corona-warn-app/id1512595757 is 1.7.1.

This is due to the out-of-date parameter [ios.current-app-version] in gulp.js which currently contains the value '1.6.1'. The version shown in https://www.coronawarn.app/en/faq/#no_risk_update_ios takes its value from [ios.current-app-version].

dsarkar:
see https://github.com/corona-warn-app/cwa-website/pull/635

PR #631, opened 2020-11-29
FAQ - Remove outdated references to current app version

heinezen:
Fixes https://github.com/corona-warn-app/cwa-website/issues/630


Ein-Tim:
@heinezen Idk but since this Pull request is about the App versions I think it would be nice if you could also quickly update the App versions in the [gulpfile](https://github.com/corona-warn-app/cwa-website/blob/master/gulpfile.js).

dsarkar (approved):

PR #664, opened 2020-12-13
fixed typo

JonasFluegge:
There was one comma too much.


Ein-Tim:
I'm not sure about that... 
I think the comma must stay there.
Test it via mentor.duden.de

svengabr:
I think @Ein-Tim is right and the comma needs to stay.

PR #668, opened 2020-12-15
Fix url for SAP Trust Center

kilbphilippSAP:
The SAP Trust Center is now reachable via https://www.sap.com/about/trust-center/security/incident-management.html (with a /security/).

At the end of the SECURITY.md the correct link is already used. Line 12 however, uses the old url.


heinezen (approved):
LGTM

PR #707, opened 2020-12-17
Define "since the update" more precisely 

Ein-Tim:
This PR defines "since the update" more precisely. ("since the update" -> "since the update to version 1.9")


heinezen:
If you don't mind an additional nitpick: We can be even more precise here in my opinion by specifying 

> since version 1.9 **of the Corona Warn App**

and 

> seit dem Update auf Version 1.9 **der Corona-Warn-App**

We have versions of other software floating around (ENF, iOS, Android), so it would not hurt to be extra specific.

Ein-Tim:
Good Idea @heinezen. Done!

dsarkar (approved):

dsarkar:
@Ein-Tim, @heinezen, NIce one....

PR #756, opened 2021-01-06
Belgien, finnland

dsarkar:
closes https://github.com/corona-warn-app/cwa-website/issues/754


src/data/faq.json
```diff
@@ -342,8 +342,8 @@
...
+                            "(Status: 6.01.2021)",
```
Ein-Tim:
```suggestion
                            "(Status: 06.01.2021)",
```

src/data/faq_de.json
```diff
@@ -351,8 +351,8 @@
...
+                            "(Stand: 6.01.2021)",
```
Ein-Tim:
```suggestion
                            "(Stand: 06.01.2021)",
```

Ein-Tim:
I left to suggestion, if we write the month 01 and not 1 we should also write the day 06 and not 6. 👍

dsarkar:
EN

image

image


DE

image


image

loki-cui (changes requested):
From what I understood is that Belgium might no go live today, so I propose to put it on "pending" until it's final.

Ein-Tim (approved):

dsarkar:
@Ein-Tim Ah... could have just accepted your commits.... next time. Thanks anyway.

dsarkar:
@loki-cui ok. blocked until further notice.

src/data/faq.json
dsarkar: ok

MikeMcC399:
The English dates on
/en/faq/#interoperability_countries
are not formatted with an English locale, and this is generally so on the https://www.coronawarn.app/ site.

In UK the date would be written 06/01/2021 with "/" delimiters instead of ".".
I actually prefer writing the date completely differently as Jan 6, 2021 in English because using the "/" notification you often don't know if it is an American format mm/dd/yyyy or a UK format dd/mm/yyyy with month and day interchanged. Another way which is unambiguous is 06-Jan-2021. That works quite well for German as well and a lot of the months have the same first three characters in English and German. Jan, Feb, (Mar/Mär), Apr, (May, Mai), Jun, Jul, Aug, Sep, (Oct/Okt), Nov, (Dec/Dez).

dsarkar:
@MikeMcC399 I see, next time we will change it. Thanks.

PR #787, opened 2021-01-15
Improve contact journal

thomasaugsten:


MikeMcC399:
The app calls it "Contact Journal" in English however the FAQ is referring to it as "Contact Diary". Would it be better to keep the terms consistent?

In German, the terms in the app and the FAQ are consistent, where both use the term "Kontakt-Tagebuch.

dsarkar (approved):

dsarkar (approved):

Issue #781, opened 2021-01-14
FAQ: [Google/Android]: Error when communicating with Google API (39508) - update needed?


MikeMcC399:
## Where to find the issue
https://www.coronawarn.app/en/faq/#API39508 
[Google/Android]: Error when communicating with Google API (39508)

## Describe the issue
The app no longer outputs the error code 39508 any more. It puts out a text error instead:

<string name="errors_risk_detection_limit_reached_title">"Limit already reached"</string>

<string name="errors_risk_detection_limit_reached_description">"No more exposure checks possible today, as you have reached the maximum number of checks per day defined by your operating system. Please check your risk status again tomorrow."</string>

The whole exposure checking went through a major update with the release of 1.9.1 according to the blog post https://www.coronawarn.app/en/blog/2020-12-16-corona-warn-app-version-1-9/ from Dec 16, 2020.

~~There haven't been any more comments made to the issue https://github.com/corona-warn-app/cwa-app-android/issues/1459 for over 5 weeks.~~ Now closed.
  

## Suggested change
Possibly change the FAQ to say fixed with 1.9.1 ~~and close https://github.com/corona-warn-app/cwa-app-android/issues/1459~~.

Edit:
https://github.com/corona-warn-app/cwa-app-android/issues/1459 was closed on Mar 9, 2021.

dsarkar:
Suggestion:


"

[Google/Android]: Limit already reached (Previously: Error when communicating with Google API (39508))
---

**UPDATE 16.01.2021:**

The notification reads now:

**'Limit already reached**

**No more exposure checks possible today, as you have reached the maximum number of checks per day defined by your operating system. Please check your risk status again tomorrow.'**

The whole exposure checking went through a major update with the release of 1.9.1, see also the blog post https://www.coronawarn.app/en/blog/2020-12-16-corona-warn-app-version-1-9/ from Dec 16, 2020, and since the release 1.9.1 no more reports on this issue have been submitted.

However, should you experience an issue related to this error notification, please submit your report here: corona-warn-app/cwa-app-android#1459.

---

_... the remainder of the original FAQ entry_
"

MikeMcC399:
@dsarkar 
Since the error message has changed, maybe the title of issue https://github.com/corona-warn-app/cwa-app-android/issues/1459 needs to be changed as well, if the old errors, and any new errors are to be captured in the same issue? The title is currently:
"🔥 Error 39508 after updating to 1.5 or later"

[...]

PR #789, opened 2021-01-16
Update faq 39508

dsarkar:
Closes https://github.com/corona-warn-app/cwa-website/issues/781

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -1000,10 +1000,15 @@
                         ]
                     },
                     {
-                        "title": "[Google/Android]: Error when communicating with Google API (39508)",
+                        "title": "[Google/Android]: Limit already reached (Previously: Error when communicating with Google API (39508))",
                         "anchor": "API39508",
                         "active": true,
                         "textblock": [
+                            "<b>UPDATE 16.01.2021</b><br/>The error notification reads now:",
+                            "<b>'Limit already reached'</b>",
+                            "<b>'No more exposure checks possible today, as you have reached the maximum number of checks per day defined by your operating system. Please check your risk status again tomorrow.'</b>",
+                            "The whole exposure checking went through a major update with the release of 1.9.1, see also the  <a href='https://www.coronawarn.app/en/blog/2020-12-16-corona-warn-app-version-1-9/' target='_blank' rel='noopener noreferrer'> blog post from Dec 16, 2020</a>,  and since the release 1.9.1 no more reports on this issue have been submitted.", 
+                            "However, should you experience an issue related to this error notification, please submit your report here: <a href='https://github.com/corona-warn-app/cwa-app-android/issues/1459' target='_blank' rel='noopener noreferrer'>associated GitHub issue 1459</a><hr/>",
                             "<b>UPDATE</b><br/>This bug has been fixed in version 1.5 of the Corona Warn App. We therefore ask you to update the app accordingly.<br/><b>IMPORTANT:</b><br/><b>After updating to version 1.5 it will take up to 24 hours until the error is gone.</b>",
                             "If you are still receiving the error after updating to 1.5 and waiting 24 hours, please report it in the <a href='https://github.com/corona-warn-app/cwa-app-android/issues/1459' target='_blank' rel='noopener noreferrer'>associated GitHub issue</a>.<hr/>",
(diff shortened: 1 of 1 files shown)


dsarkar:
EN

image

dsarkar:
DE version in progress ...

MikeMcC399:
@dsarkar 
> DE version in progress ...

The error message in German is:

<string name="errors_risk_detection_limit_reached_title">"Limit bereits erreicht"</string>

<string name="errors_risk_detection_limit_reached_description">"Heute sind keine weiteren Risiko-Überprüfungen möglich, weil das von Ihrem Betriebssystem festgelegte Limit von Risiko-Überprüfungen pro Tag bereits erreicht ist. Bitte überprüfen Sie Ihren Risikostatus morgen wieder."</string>

from strings
```
errors_risk_detection_limit_reached_title
errors_risk_detection_limit_reached_description
```

heinezen:
@dsarkar @MikeMcC399 Should this get its own FAQ entry? I don't see why the "limit reached" error necessarily has to be group together with a seemingly unelated issue.

MikeMcC399:
@heinezen 

The underlying Android error status is still 39508. The message from CWA is however no longer 39508, it has been changed to a reader-friendlier message which says:


<string name="errors_risk_detection_limit_reached_title">"Limit already reached"</string>

Here is the link to the Google documentation for the error code:
https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/ExposureNotificationStatusCodes?hl=en#public-static-final-int-failed_rate_limited.

heinezen:
I think from a user-perspective, the ammendment of 39508 is confusing (even though it makes technical sense).

I have made a counter-proposal to this approach https://github.com/corona-warn-app/cwa-website/commit/491faf68a96c4b6e08de803697845e41afd0dca1 where the limit reached error has a separate entry that is linked in the original 39508 entry. Didn't change much else.

MikeMcC399:
@heinezen 
That's a valid point and a good suggestion!

dsarkar:
@heinezen @MikeMcC399 
 491faf6 LGTM. 
 Usually we keep the history and track of changes of all entries. Suggestion: We move the old 39508 entry to the bottom of the FAQ, section "resolved" with a comment/link to your new article. Altogether this will look cleaner. I will discuss this internally if it is ok to proceed this way, and then, if approved, we can create finally a PR as you suggested. Thanks!

dsarkar:
@heinezen @MikeMcC399 
Internally discussed:
In order to harmonize, shorten, clean-up the FAQ page we now create a .md file, where we document outdated FAQ entries. We will link the old and new article mutually, so it is transparent but at the same time it should look cleaner.

MikeMcC399:
@dsarkar 
> In order to harmonize, shorten, clean-up the FAQ page we now create a .md file, where we document outdated FAQ entries. We will link the old and new articles mutually, so it is transparent but at the same time, it should look cleaner.

Good idea! The FAQ list is now very long and if someone is viewing on a mobile device it can be quite difficult to find the relevant information.

If the minimum versions of the app on iOS and Android are increased then the entries which only refer to problems with versions which are subject to forced update can be archived. So in the .md file split into a section for resolved issues and another section for resolved issues with versions no longer supported perhaps?

See your comments about minimum versions:
https://github.com/corona-warn-app/cwa-website/pull/852#issuecomment-775905999
https://github.com/corona-warn-app/cwa-website/pull/852#issuecomment-775979928

and server config:
https://github.com/corona-warn-app/cwa-server/blob/a22242bea548dea6f6e35c4fe4c8eba3d8b4fc64/services/distribution/src/main/resources/application.yaml#L112-L119

    app-versions:
      latest-ios: ${IOS_LATEST_VERSION:1.5.3}
      min-ios: ${IOS_MIN_VERSION:1.5.3}
      latest-android: ${ANDROID_LATEST_VERSION:1.5.1}
      min-android: ${ANDROID_MIN_VERSION:1.5.1}
      # With ENF v2 Android apps are versioned by Version Code and not by Semantic Versioning
      min-android-version-code: ${ANDROID_MIN_VERSION_CODE:48}
      latest-android-version-code: ${ANDROID_LATEST_VERSION_CODE:48}

dsarkar:
replaced in favour of https://github.com/corona-warn-app/cwa-website/pull/895

Issue #763, opened 2021-01-09
npm and webserver in README.md


MikeMcC399:
## Source Document
In https://github.com/corona-warn-app/cwa-website/blob/master/README.md#requirements it says:

### Requirements section from above document

"You need [Node.js](https://nodejs.org/en/) (which includes npm) to build the website. Optionally, you need an HTTP Server such as [http-server](https://github.com/http-party/http-server) to run and test it locally."

### Questions

1) As I found out through experience using npm on the https://github.com/corona-warn-app/cwa-documentation repo, using npm 7.x can have undesired effects on 
[package-lock.json](https://github.com/corona-warn-app/cwa-website/blob/master/package-lock.json) which contains the line
https://github.com/corona-warn-app/cwa-website/blob/d08165f048d6075eab97b50b4fc1eee459dd2afe/package-lock.json#L4
Running npm install using npm 7.x changes lockfileVersion to 2 in package-lock.json and changes the entire format of the file. The problem does not occur using npm 6.x.
<br /> Should README.md recommend to install the LTS (Long Term Support) version of node.js and npm, which includes version 6.x of npm? See https://nodejs.org/en/download/ which labels LTS as "Recommended for Most Users".

2) I don't understand the reference to [http-server](https://github.com/http-party/http-server) as a prerequisite to run a test locally. After installing node.js, the commands:
`npm ci`
`npm start`
are sufficient to get the cwa-website to be fired up on a local webserver `http://localhost:8000/en/`. The installation of HTTP Server doesn't seem necessary.
<br />Could the explicit reference to "HTTP Server" be removed from README.md?

heinezen:
> Should README.md recommend to install the LTS (Long Term Support) version of node.js and npm, which includes version 6.x of npm?

We can change the docs to mention LTS explicitely, but unless a changed `package-lock.json` is committed in a PR, it shouldn't make that much of a difference what is used. Changing `npm install` to `npm ci` in **Getting Started** should counter this behaviour.

> I don't understand the reference to http-server as a prerequisite to run a test locally.

If one would want to actually *deploy* the website somewhere else, then a dedicated web server is better than using node-http. nodejs is not really the most stable solution in that regard. I guess this could be worded better in the docs by saying that a web server is an optonal dependency for deployment.

---
Corona-Warn-App Open Source Team

MikeMcC399:
Should we change the README.md?

There don't seem to be any new people wanting to contribute here, so it's probably not so important.

I made a note that
[...]

PR #859, opened 2021-02-09
Recommend LTS version of node

heinezen:
Fixes https://github.com/corona-warn-app/cwa-website/issues/763


MikeMcC399 (approved):
That is OK for me. Thank you!
The important point was to standardize officially on the LTS version of npm to avoid compatibility issues.

dsarkar (approved):

Issue #848, opened 2021-02-06
Special case "Limit already reached" after app reset


MikeMcC399:
## Where to find the issue

[Google/Android]: Error when communicating with Google API (39508)
https://www.coronawarn.app/en/faq/#API39508

Also work-in-progress PR #789.

## Describe the issue

As described in "Reset App late in day causes "Limit already reached" error message" https://github.com/corona-warn-app/cwa-app-android/issues/2248 the error message "Limit already reached" can be caused by resetting the app late in the day after approximately 20:00 UTC (21:00 CET or 22:00 CEST) when the app has been operating correctly and is permanently connected to WiFi, so that it will have performed a risk evaluation every 4 hours. 

According to https://developers.google.com/android/exposure-notifications/exposure-notifications-api#modes provideDiagnosisKeys quota is Six per day (counter resets at 00:00 UTC) in ExposureWindow mode (aka v2). So if the first submission is done at 00:01 UTC, followed by 04:01, 08:01, 12:01, 16:01 at 20:01 the quota of six submissions for the day has already been used up. This should not normally matter, because the next submission will not be due until after 24:00 UTC. However, Reset App, could provoke an additional submission and thus exceed the quota.

This Reset App issue will not be fixed (see https://github.com/corona-warn-app/cwa-app-android/issues/2248#issuecomment-773900831) since it is considered an edge case and would be complex to work around.

The situation is not described in the FAQ https://www.coronawarn.app/en/faq/#API39508.

## Suggested change

In the draft PR #789, after the sentence:

"However, should you experience an issue related to this error notification, please submit your report here: [associated GitHub issue 1459](https://github.com/corona-warn-app/cwa-app-android/issues/1459)"

add

EN
If you have just carried out "Reset App" in the app Settings and you then receive the error message "Limit already reached", please follow the instructions in the message and wait until the next day for the limit to be automatically reset.

DE
Wenn Sie gerade "Anwendung zurücksetzen" in den Einstellungen der App durchgeführt haben, und danach die Fehlermeldung "Limit bereits erreicht" bekommen, dann warten Sie bitte bis zum nächsten Tag, dass das Limit automatisch zurückgesetzt wird, wie in der Meldung empfohlen.

heinezen:
Also relevant: https://github.com/corona-warn-app/cwa-website/pull/895

MikeMcC399:
This issue was resolved by PR #895, which was merged on May 3, 2021.

PR #895, opened 2021-02-16
FAQ entry - Limit reached (replaces API39508)

heinezen:
Alternative to https://github.com/corona-warn-app/cwa-website/pull/789


src/data/faq_de.json
```diff
@@ -267,6 +267,19 @@
...
+                        "textblock": [
+                            "Falls Sie nach dem Zurücksetzen der App die folgende Fehlermeldung erhalten haben, warten Sie bitte bis zum nächsten Tag, bis Sie die App wieder öffnen.",
+                            "<b>'Limit bereits erreicht'</b>",
+                            "<b>'Heute sind keine weiteren Risiko-Überprüfungen möglich, weil das von Ihrem Betriebssystem festgelegte Limit von Risiko-Überprüfungen pro Tag bereits erreicht ist. Bitte überprüfen Sie Ihren Risikostatus morgen wieder.'</b>",
+                            "Der Fehler sollte dann verschwunden sein und Ihr Risiko-Status sollte korrekt angezeigt werden.",
+                            "Dieser Fehler kann auftreten, wenn Sie die App zwischen 20:00 Uhr und 0:00 Uhr zurücksetzen und ihr Gerät den gesamten Tag über eine konstante WLAN-Verbindung hatte. Die Corona-Warn-App hat ein Tageslimit von 6 regelmäßigen Risiko-Überprüfungen. In diesem Fall führt das Zurücksetzen der App zu einer außerplanmäßigen Risiko-Überprüfung, bei der das Tageslimit überschritten werden kann. Die Funktionalität der App wird durch die fehlgeschlagene Überprüfung nicht beeinträchtigt.",
```
MikeMcC399, line 279: Suggest 0:00 Uhr => 00:00 Uhr
ihr Gerät => Ihr Gerät

src/data/faq.json
```diff
@@ -266,6 +266,19 @@
...
                     },
+                    {
+                        "title": "[Google/Android]: Error after resetting the app: Limit already reached",
+                        "anchor": "limit_reached",
+                        "active": true,
+                        "textblock": [
+                            "If you encounter the following error message after resetting the app, please wait until the next day before you open the app again.",
+                            "<b>'Limit already reached'</b>",
+                            "<b>'No more exposure checks possible today, as you have reached the maximum number of checks per day defined by your operating system. Please check your risk status again tomorrow.'</b>",
+                            "The error should have disappeared by then and the your risk status should be displayed correctly.",
```
MikeMcC399, line 277: and the your risk status => and your risk status
or 
and your risk status should then

src/data/faq.json
```diff
@@ -266,6 +266,19 @@
...
+                            "This error can occur if you reset the app between 20:00 and 0:00 and your device was constantly connected to Wifi for the whole day. The Corona-Warn-App has a daily limit of 6 periodic checks. In this case, resetting the app triggers an unscheduled risk check that can exceed the daily limit.The functionality of the app is not affected by the failed check.",
```
MikeMcC399, line 278: 0:00 or 00:00?
Wifi => WiFi (or officially Wi-Fi)

src/data/faq_de.json
```diff
@@ -1098,6 +1111,7 @@
                         "anchor": "API39508",
                         "active": true,
                         "textblock": [
+                            "<b>AKTUELL 16.01.2021</b><br/>Die Fehlermeldung lautet nun \"Limit bereits erreicht\". Bitte lesen Sie <a href='#delete_random_ids' target='_blank' rel='noopener'>den dazugehörigen FAQ-Eintrag</a> für weitere Informationen.",
```
MikeMcC399, line 1114: #delete_random_ids => #limit_reached

src/data/faq.json
```diff
@@ -1096,6 +1109,7 @@
                         "anchor": "API39508",
                         "active": true,
                         "textblock": [
+                            "<b>UPDATE 16.01.2021</b><br/>The error notification now reads \"Limit already reached\". Please see <a href='#delete_random_ids' target='_blank' rel='noopener'>the corresponding FAQ entry</a> for further information on this issue.",
```
MikeMcC399, line 1112: #delete_random_ids => #limit_reached

MikeMcC399 (changes requested):
Wrong link to `#limit_reached` otherwise just a few text suggestions for you.

heinezen:
@MikeMcC399 Thanks. All fixed :)

MikeMcC399 (approved):
LGTM 👍

MikeMcC399:
This PR addresses issue #848.

heinezen:
@dsarkar Please check if we should also change the FAQ redirect for API39508 in here: https://github.com/corona-warn-app/cwa-website/blob/master/src/data/faq_redirects.json

dsarkar:
@MikeMcC399 @heinezen ok, thanks. We need to be careful with changing the anchor. We might brake links and also, I think that this FAQ entry with the old anchor is linked from the CWA when this error should occur. 

Option 1: Don't change anchor
Option 2: Redirect https://github.com/corona-warn-app/cwa-website/blob/2342931cc1fa68c538eb3d1a25fb6f74620233d7/src/data/faq_redirects.json

heinezen:
Option 2 is better I think, since the app then links to the appropriate FAQ entry

MikeMcC399:
The app itself no longer links to any FAQ article. That is only is older versions.

The error message is now:

**Limit already reached**

"No more exposure checks possible today, as you have reached the maximum number of checks per day defined by your operating system. Please check your risk status again tomorrow."

I suggest to make a strategy decision about what versions of apps still need to have articles available to users. (I think that the current version and the previous version are all that is necessary, but you may think differently.)

The second point is about a policy decision whether to permanently provide a target destination for any URL that has ever been published. Because a published URL may have been written into a posting on GitHub, I believe it is always better to try not to leave a URL with a 404 file not found error. Again, you may think differently, but I think that there should continue to be a meaningful target for:
https://www.coronawarn.app/en/faq/#39508
https://www.coronawarn.app/en/faq/#API39508

If you want to continue to support old versions in the FAQ, then you should not change the link in the json file. Old apps will link to the old article with anchor #API39508 via anchor #39508.

If you decide not to support old versions in the FAQ, then you could remove all the text from the FAQ article #API39508, except for a link to the anchor #limit_reached.

dsarkar:
@MikeMcC399 @heinezen Latest info (1 minuet old) I have is: On the server the min. req. version iOS 1.5.3 and Android 1.0.4 **(!)**. There shouldn't be users with working CWA versions older than those mentioned, and all FAQ entries referring to < iOS 1.5.3 / Android 1.0.4 can be archived.

MikeMcC399 (approved):
@dsarkar 

Will you merge this PR now? I reviewed it again today and it looks good.

We have not had any new reports about this problem for some time now.

dsarkar:
@MikeMcC399 @heinezen Thanks!

PR #906, opened 2021-02-23
FAQ: In-app statistics (EN) correct translation discrepancies

MikeMcC399:
This PR corrects some translation discrepancies in 
https://www.coronawarn.app/en/faq/#further_details
**In-App statistics starting with release 1.11**

compared to
https://www.coronawarn.app/de/faq/#further_details
In-App Statistiken ab Release 1.11

This is the updated text:

In-app statistics EN

---

- The main problem was stating that the results in the app were deferred, which means a long delay, for example "The exhibition was deferred until next year due to the pandemic". I changed that to "slightly delayed" instead because deferred was not a correct use of the word in this context. (Reference: https://dictionary.cambridge.org/dictionary/english/defer.)
- The English text referred to "figures and graphs" where the German text only refers to "Zahlen". I removed the references to "graphs".
- I also wrote out the first reference to "Robert Koch Institute" in full then used the abbreviation (RKI).
- There are some other minor wording changes I made as well.

The text should now better correspond to the German text.


dsarkar (approved):

dsarkar:
@MikeMcC399 Looks good to me. Thanks.

dsarkar:
@GisoSchroederSAP Please, have a look and merge if you approve! Thanks!

heinezen (approved):

heinezen:
This change should be fine on its own. Will try to get approval again.

GisoSchroederSAP (approved):
Even if I don't see the necessity of replacing "defer" by "delay" in this context, I am okay with those changes. As long as the meaning becomes more clear to the reader, any enhancement is much appreciated!

dsarkar:
@MikeMcC399 Thanks for adjusting this article.

PR #909, opened 2021-02-23
Blog update

HannaHei:


MikeMcC399 (changes requested):
There is an extra character in the title

page-title: "RCorona-Warn-App’s risk calculation further adjusted after detailed tests"

It should be Corona not RCorona

Ein-Tim:
If this PR won't be merged it should be closed.

PR #955, opened 2021-03-07
FAQ - Add instructions to find ENF version through app 

heinezen:
Resolves https://github.com/corona-warn-app/cwa-website/issues/954

PR #958, opened 2021-03-08
Check Google ENS version from inside CWA app (closes #954)

### LINKED ISSUE #954: FAQ: Add instructions to find ENS version through app
## Where to find the issue

**[Google/Android]: Which version of the COVID-19 Exposure Notifications System is currently installed?**
https://www.coronawarn.app/en/faq/#ENF_version

**[Google/Android]: Welche Version der COVID-19-Benachrichtigungen habe ich?**
https://www.coronawarn.app/de/faq/#ENF_version

## Describe the issue

The instructions for determining the installed version of the Google Exposure Notifications System on Android do not include how to find this out directly in the Corona-Warn-App itself. 

There was a bug in 1.12.0 which prevented scrolling down to the ENS version number on small screens - see https://github.com/corona-warn-app/cwa-app-android/issues/2302. This has been fixed in the CWA 1.13.2 release, which means it would now be good to update the FAQ instructions accordingly.

Here is an example of where the version number is displayed:

App Information 1 13 2 Nexus 5

## Suggested change

- Describe how to access the ENS version number through CWA.
- Move the description about how to use the Google UI to a secondary position in the description.

I will submit a PR for this change.

MikeMcC399:
This PR resolves #954 by updating FAQs
https://www.coronawarn.app/en/faq/#ENF_version and
https://www.coronawarn.app/de/faq/#ENF_version
with instructions how to display the version number of the Google Exposure Notifications System from within the Corona-Warn-App using the App Information screen.

Information about the format of the version number used by Google prior to V1.5 has been removed because this is no longer relevant. Any mobile device using CWA for Exposure Logging must have an Internet connection in order to carry out risk evaluation. As soon as a device is connected to the Internet, Google Play services ensures that the Exposure Notifications System is updated through silent installation. The rollout to different devices can be staggered by some weeks due to different hardware dependent variants of Google Play services, but since V1.5 was already rolled out in Juli 2020 (more than 7 months ago), there can be no more active devices running a version older than V1.5.

---
**EN**
https://www.coronawarn.app/en/faq/#ENF_version 

ENF_version_EN

---
**DE**
https://www.coronawarn.app/de/faq/#ENF_version

ENF_version_DE


heinezen:
My main gripe with this is that the entry still suffers from:

*  "Too Much Information". Pretty much all users won't need to understand the version history or how the version number should be read. If we expect this FAQ entry to be used to aid in bug reports it should quickly get to the point (this is the reason why I removed so much in https://github.com/corona-warn-app/cwa-website/pull/955)
* Android-specificness. We want to avoid that

We could spplit the information about the ENS into a separate entry "What is the ENS?". (However, I'm not sure if this question is still a *frequently* asked question.) That approach would work out cleaner imo.

----

MikeMcC399:
@heinezen 
> Android-specificness. We want to avoid that

The current FAQ article https://www.coronawarn.app/en/faq/#ENF_version applies only to Google / Android. 

Do you have screenshots from iOS e.g. App Information to show how iOS displays which version of the Apple Exposure Notification framework is installed? As you say, it may be necessary to split the article in order to have one which applies to both Android and iOS.

This is from Google Android:

App Information EN ENF Version


I would tend to leave the details in the FAQ article for the moment and look to simplifying it later. This is a matter of opinion though.

Ein-Tim:
@MikeMcC399 

You can't find out the exact version under iOS, neither via the app nor via settings. 
You can only say which version the app is using and which iOS version you are running.

MikeMcC399:
@Ein-Tim 
> You can't find out the exact version under iOS, neither via the app nor via settings.
You can only say which version the app is using and which iOS version you are running.

Thanks for your input on iOS. In that case I think the FAQ article https://www.coronawarn.app/en/faq/#ENF_version should stay assigned to Google/Android only, since the App Information display only contains the ENF Version line in the Android version of CWA, not in iOS.

heinezen:
@Ein-Tim @MikeMcC399  That's unfortunate, but yes, we have to leave iOS out then. I would still recommend separating the instructions how to get the ENF number from the infos about the ENF.

MikeMcC399:
@heinezen 
I agree that there is a lot of information in this FAQ article and it could be split. I would suggest merging this change first and tackling the split and simplification separately.

dsarkar (approved):

dsarkar:
@MikeMcC399 ok, thanks for all the effort. The article improved, so lets merge it. Splitting is optional.

MikeMcC399:
@dsarkar 
> thanks for all the effort. The article improved, so lets merge it. Splitting is optional.

You're welcome! 

The original author of the article answered a lot more questions in the article than the title **"Which version of the COVID-19 Exposure Notifications System is currently installed?"** suggests.

If the article is split up, then consideration also needs to be given to the other articles which reference this article to make sure a change doesn't introduce inconsistencies into the information flow. That could be part of the review project described in https://github.com/corona-warn-app/cwa-website/issues/971.

PR #959, opened 2021-03-09
Screenshots page

svengabr:
added new screenshots page


MikeMcC399:
Hi @svengabr 
and welcome back!

Great enhancement to the website!

The screenshots don't say what version of the app they are from and it looks like the ones from the Contact Journal are from a future version, not the currently released 1.13.2.

PR #997, opened 2021-03-18
5703 lazyloading

mtb77:


mtb77:
@christianneu because of the use of IntersectionObserver we would throw out compatibility for IE: https://caniuse.com/intersectionobserver 
Option is to use https://github.com/w3c/IntersectionObserver/tree/main/polyfill Polyfill.js.

Issue #971, opened 2021-03-11
FAQ - Remove outdated information from FAQ and review existing entries 🔧 


heinezen:
## Motivation

Due to the now substantial version history of the CWA, the FAQ contains a lot of legacy information. Content describing issues/fixes for several releases is often mixed together, creating possibilities for confusion. Formatting also varies between the individual entries.

## Suggested change

Overhaul all entries by removing, improving or reordering them

- [x] Check existing entries for relevance
- [x] Check existing entries for outdated info
- [ ] Trim entries that are overly long or split them into separate issues
- [ ] Use consistent wording (for terminology: https://github.com/corona-warn-app/cwa-website/issues/524) and formatting, e.g. instructions should always be a enumerated list and the app name should be "Corona-Warn-App" with hyphens
- [ ] Move obsolete/resolved entries to *somewhere else* (either through https://github.com/corona-warn-app/cwa-website/pull/894 or other means)
 
----

FAQ entries that were updated:

- [x] who_can_help (https://github.com/corona-warn-app/cwa-website/pull/1083)
- [x] international_phone_numbers (https://github.com/corona-warn-app/cwa-website/pull/1078)
- [ ] further_details
- [ ] update_iOS
- [ ] cause2001  (https://github.com/corona-warn-app/cwa-website/pull/1081)
- [x] access-list  (https://github.com/corona-warn-app/cwa-website/pull/1069)
- [ ] battery_optimization_background  (https://github.com/corona-warn-app/cwa-website/pull/1085)
- [ ] no_risk_update_ios14 (https://github.com/corona-warn-app/cwa-website/pull/1075)
- [x] ios_black_screen  (https://github.com/corona-warn-app/cwa-website/pull/1077)
- [ ] encounter_but_green
- [ ] android_location
- [ ] qr_scan
- [x] qr_test  (https://github.com/corona-warn-app/cwa-website/pull/1073)
- [ ] red_card_how_to_test
- [x] battery (https://github.com/corona-warn-app/cwa-website/pull/1084)
- [ ] no_risk_update_ios
- [x] ios_power_save (https://github.com/corona-warn-app/cwa-website/pull/1076)
- [ ] no_risk_update
- [x] huawei_honor (https://github.com/corona-warn-app/cwa-website/pull/1079)
- [x] ENError13 (https://github.com/corona-warn-app/cwa-website/pull/1080)
- [ ] exposure_check
- [ ] cause9002_recovery
[...]

Issue #712, opened 2020-12-18
FAQ: Add equivalent distances to dB levels in FAQs


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/#encounter_count_19
- https://www.coronawarn.app/de/faq/#encounter_count_19
- https://www.coronawarn.app/en/faq/#encounter_19_calc
- https://www.coronawarn.app/de/faq/#encounter_19_calc

## Describe the issue

Some FAQ articles explain risk calculations referring to dB levels. There is no display of dB levels in the CWA so these figures have no meaning to the majority of app users. I imagine also that most readers will not know that dB is a logarithmic scale.

## Suggested change

Add explanatory distance levels to the dB levels referred to in the FAQs.

According to https://github.com/corona-warn-app/cwa-documentation/blob/master/cwa-risk-assessment.md#procedure
in section "Procedure", 3rd paragraph:
"each dB value is associated with a particular distance".

Attenuation | Distance
--- | ---
55 dB | 1.5m
63 dB | 3m
73 dB | 8m

Alternatively, use the definitions provided by @heinezen in https://github.com/corona-warn-app/cwa-documentation/issues/484#issuecomment-781382074 and add these to the FAQs:

* <55 dB is very close
* 55-63 dB is mid-range
* 63-73 dB is far away
* \>73dB is very far away

dsarkar:
OK, sounds good to me. Will be included.

vaggelisdouros:
2 questions from my side:

a) can you provide the formula that translates the bluetooth signal attenuation (measured in dB) into distance (measured in meters)? In multiple cases, it is mentioned that a 73 dB attenuation corresponds to a distance of 8 meters, whereas 55 dB are translated into 1.5 meters, but i have not found the formula used or at least a reference to a paper/textbook.

b) In the updated diagram for risk calculation with enf v2, it is unclear to me how the time where the signal attenuation is between 63 dB and 73 dB is taken into account. Unless there is a typo, It seems that this signal is considered harmless and does not take part in the calculation of the weighted exposure time (even with a 0.5 coefficient), therefore i do not really understand whether the threshold is set to 73 db or 63 dB https://github.com/corona-warn-app/cwa-documentation/blob/master/images/risk_calculation/risk_calculation_enf_v2_overview.pdf

MikeMcC399:
@vaggelisdouros 
[...]

PR #1050, opened 2021-04-05
[WIP] FAQ - update 'encounter_19_calc' entry for #971

heinezen:
Changelog:
- Move last paragraph's text sections to a better place
- "only" removed in english entry
- remove mention of 10 minutes signal attenuation time window and just keep 5 minutes


MikeMcC399 (changes requested):
Please change all references to 2.0 of the API to 2 (without a minor version ".0") otherwise the version might be confused with the major.minor version numbering of the Google Play services module, which is currently at 1.8.

---
Unfortunately Apple and Google have not aligned their nomenclature!

Apple uses the name "Exposure Notifications framework"
(see https://developer.apple.com/documentation/exposurenotification#overview)
and refers to version 2
(see https://developer.apple.com/documentation/bundleresources/information_property_list/enapiversion).

Google uses the name "Exposure Notifications system"
(see https://developers.google.com/android/exposure-notifications/exposure-notifications-api#architecture) 
and the version which follows "Legacy v1 mode" is **not** v2, it is called "ExposureWindow mode"
(see https://developers.google.com/android/exposure-notifications/exposure-notifications-api#modes).

In the Google world "ExposureWindow mode" was introduced with version 1.5 of Google Play services. The current version of Google Play services is version 1.8.

To avoid hitting CWA users with the confusing naming issues, I would suggest calling it "Exposure Notification Framework" in any place where Apple and Google are being referred to. 


You could change the introductory sentence to 
"With the introduction of Corona-Warn-App version 1.9, a new version of the Exposure Notification Framework (ENF) from Apple/Google is used, version 2 (also known as ExposureWindow mode by Google)."

After that only refer to version 2, not to ExposureWindow mode in the text.

MikeMcC399:
The description about the risk calculation regarding dB levels does not match the documentation on https://github.com/corona-warn-app/cwa-documentation/blob/master/images/risk_calculation/risk_calculation_enf_v2_overview.pdf.

MikeMcC399:
There is a lot of overlap between
http://localhost:8000/en/faq/#encounter_count_19 and
http://localhost:8000/en/faq/#encounter_19_calc

Please consider combining the two FAQ articles, so that the risk calculation is only described in one place.

MikeMcC399:
There is no explanation of what dB levels mean. See https://github.com/corona-warn-app/cwa-website/issues/712 and the explanation published to https://github.com/corona-warn-app/cwa-documentation/issues/484#issuecomment-781382074.

heinezen:
@MikeMcC399 I'll let this PR stay up until the PR for `encounter_count_19` is up too for direct comparison.

> To avoid hitting CWA users with the confusing naming issues, I would suggest calling it "Exposure Notification Framework" in any place where Apple and Google are being referred to.

Noted. I've added it to the general issues/remarks in #971

heinezen:
Closed in favour of https://github.com/corona-warn-app/cwa-website/pull/1222

PR #1080, opened 2021-04-13
FAQ - update 'ENerror13' entry

heinezen:
Changelog:
- less paragraphs
- slight improvement on wording


src/data/faq.json
```diff
@@ -297,11 +297,9 @@
...
-                            "If the Error doesn't disappear after 24h, please open a issue on <a href='https://github.com/corona-warn-app/cwa-app-ios/issues/' target='_blank' rel='noopener noreferrer'>GitHub</a>."
+                            "The Error 13 is thrown by Apple's Exposure Notification Framework, which is used by the Corona-Warn-App. It will appear if more than 6 exposure checks happened in the last 24 hours. Apple does not allow more than 6 exposure checks within 24 hours. The error will disappear after 24 hours at the latest. You don't have to do anything.",
+                            "Please do not remove and reinstall the Corona-Warn-App, because the error is caused by the underlying Apple Exposure Notification Framework which is independent from the app. Therefore, it will persist, even after reinstallation.",
+                            "If the Error doesn't disappear after 24h, please open an issue on <a href='https://github.com/corona-warn-app/cwa-app-ios/issues/' target='_blank' rel='noopener noreferrer'>GitHub</a>."
```
Ein-Tim, line 302: We currently have some users reporting that the Error 13 doesn't disappear after 24h (see https://github.com/corona-warn-app/cwa-app-ios/issues/2395).

You should internally decide wether you want a new issue for every case or if we focus on one of the issues in the above linked overview. 
If we focus on one, we should link to this one.

heinezen: there's still not enough information about this so that we can safely link to one issue I think. A new issue for every case is still preferred.

If we get new information this week, I'll update this issue and merge with high priority.

Ein-Tim: Okay thank you @heinezen!

dsarkar:
left new version is show.

image

image


LGTM

Thanks to  @heinezen @Ein-Tim .

PR #1085, opened 2021-04-13
FAQ - redirect 'battery_optimization_background' to 'no_risk_update'

heinezen:
Changelog:
- redirect to [#no_risk_update](https://www.coronawarn.app/en/faq/#no_risk_update)


MikeMcC399 (commented):
This article assumes that enabling the prioritized background activity directly disables battery optimization. This is however not true for current app versions. In fact the function just puts the user into the Android settings level of the operating system and leaves disabling to the user.

See https://www.coronawarn.app/en/faq/#no_risk_update for instructions about how to do it.

Also https://github.com/corona-warn-app/cwa-app-android/issues/1606 for an issue of the user interface usability.

heinezen:
@MikeMcC399 Ah yes, thank you for reminding me. This issue again... I don't think there is a quick solution for this PR right now because I do not want to link to https://www.coronawarn.app/en/faq/#no_risk_update before it has been reworked. Maybe we'll split the instructions from the https://www.coronawarn.app/en/faq/#no_risk_update entry into a new FAQ entry and then link it.

MikeMcC399:
@larswmh 

My comment (https://github.com/corona-warn-app/cwa-website/pull/1085#pullrequestreview-634734358) from one year ago is still relevant.

The Android app no longer changes the status itself. It puts the user into device settings where they make the change themselves.

Maybe it would be better just to remove this article and redirect it to [#no_risk_update](https://www.coronawarn.app/en/faq/#no_risk_update)?

larswmh:
@MikeMcC399 
 > Maybe it would be better just to remove this article and redirect it to [#no_risk_update](https://www.coronawarn.app/en/faq/#no_risk_update)?

Changes have been applied in https://github.com/corona-warn-app/cwa-website/pull/1085/commits/1ed9d007b8d0bab33971f09ba267c9bad236f7f9, thanks for the suggestion. Waiting for current build problems to be resolved and then merging this into master.

PR #1090, opened 2021-04-15
Download page

svengabr:
Added a new download page to be called directly

https://www.coronawarn.app/en/downlaod
https://www.coronawarn.app/de/downlaod

Also, the language redirector was extended with a "?target" functionality
https://www.coronawarn.app/?target=download

This was done for the new event registration process.


Ein-Tim:
@svengabr 

I think the deployment failed: https://github.com/corona-warn-app/cwa-website/actions/runs/752221705

MikeMcC399:
@svengabr 
There are typos in the links in this PR. They should be
https://www.coronawarn.app/en/download
https://www.coronawarn.app/de/download
not (downlaod).

The app naming on these pages (Corona Warn App) without hyphens is not consistent with the app naming in the stores where hyphens are used i.e. Corona-Warn-App.

PR #1091, opened 2021-04-15
Improve & correct download page

Ein-Tim:
This PR improves the wording of the [download page](https://www.coronawarn.app/?target=download).

## Done in this PR:

- [X] add hyphen to "Corona Warn App"
- [X] add "herunter" to write "lade die Corona-Warn-App herunter" instead of "Lade die Corona-Warn-App" 
- [X] correct typo "Dowload" 

## TODO (if wanted):
- [ ] change button order (#1094) - Ready to implement
- [ ] change HTML title (#1094) - Ready to implement
- [ ] change from "Du" to "Sie" (#1093) - Ready to implement

## Current status: 

Waiting on feedback from @svengabr, @heinezen or @dsarkar.

### DIFF
--- a/src/data/download.json
+++ b/src/data/download.json
@@ -1,7 +1,7 @@
 {
     "section-main": {
         "headline": {
-            "title": "Download the Corona Warn App to check in"
+            "title": "Download the Corona-Warn-App to check in"
         },
         "content": {
             "text": "Available for download on Apple App Store and Google Play."
--- a/src/data/download.json
+++ b/src/data/download.json
@@ -9,7 +9,7 @@
         "image": {
             "url": "/assets/img/app-mockup@2x.png",
             "url-webp": "/assets/img/app-mockup@2x.webp",
-            "title": "Corona Warn App",
+            "title": "Corona-Warn-App",
             "alt": "Corona-Warn-App Smartphone"
         }
     }
--- a/src/data/download_de.json
+++ b/src/data/download_de.json
@@ -9,7 +9,7 @@
         "image": {
             "url": "/assets/img/app-mockup@2x.png",
             "url-webp": "/assets/img/app-mockup@2x.webp",
-            "title": "Corona Warn App",
+            "title": "Corona-Warn-App",
             "alt": "Corona-Warn-App Smartphone"
         }
     }
--- a/src/data/download_de.json
+++ b/src/data/download_de.json
@@ -4,7 +4,7 @@
             "title": "Lade die Corona-Warn-App herunter um dich einzuchecken"
         },
         "content": {
-            "text": "Zum Dowload verfügbar im Apple App Store und auf Google Play."
+            "text": "Zum Download verfügbar im Apple App Store und auf Google Play."
         },
         "image": {
             "url": "/assets/img/app-mockup@2x.png",
--- a/src/data/download_de.json
+++ b/src/data/download_de.json
@@ -1,7 +1,7 @@
 {
     "section-main": {
         "headline": {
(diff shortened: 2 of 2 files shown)


MikeMcC399:
@Ein-Tim 
Are you involving @svengabr in this, since he was the one who merged the original files?

Ein-Tim:
@MikeMcC399 

Yes, I think it would be the best if @svengabr takes a look here. But my main point would be to get these changes merged before the release of version 2.0.

MikeMcC399:
@Ein-Tim

I tried going to https://www.coronawarn.app/?target=download on an Android phone with the Chrome browser and it showed the text for Apple and Google, but only showed the Google Play download button.

Did you try what happens on an iOS phone when you load that page? Does it only show the Apple download button?

If it only shows the button appropriate for the operating system where the browser is running, I think that is a nice feature. On Windows it shows both buttons.

Ein-Tim:
@MikeMcC399 

> Did you try what happens on an iOS phone when you load that page? Does it only show the Apple download button?

Yes! Thats a really nice feature! I tried on my iPhone and it only showed the App Store! 
In fact, my iPhone instantly redirected me to the Corona-Warn-App in the App Store. 
Do you see the same redirect on your Android phone?

MikeMcC399:
@Ein-Tim 
 >I tried on my iPhone and it only showed the App Store!
In fact, my iPhone instantly redirected me to the Corona-Warn-App in the App Store.

Interesting to hear!

>Do you see the same redirect on your Android phone?

On Google Chrome it just displayed the page with the button, which I think is fine. It is just one tap for the user to open the Google Play Store, assuming that the page is being opened on an Android app which has the Play Store installed, which is not the case for every Android mobile device.

Ein-Tim:
@MikeMcC399 

Opinion on adding the change to "Sie" form here? Or should this be a separate PR?

@dsarkar Maybe you can also take a look here, thanks! 🙂

MikeMcC399:
@Ein-Tim 
> Opinion on adding the change to "Sie" form here? Or should this be a separate PR?

I noticed some other issues, which I just described in https://github.com/corona-warn-app/cwa-website/issues/1094.

It seems like there are different points which could benefit from cleaning up. I would expect @svengabr to be given the opportunity to at least comment, if not make changes himself, since it is so soon after the original changes were merged in the fast-tracked PR https://github.com/corona-warn-app/cwa-website/pull/1090.

Ein-Tim:
@MikeMcC399 

Thank you. I'll wait for any comment of the Open-Source Team. 
Anyway I'll correct the typo you found in #1094.

Ein-Tim:
Marking this PR as WIP until there is a decision on #1094

svengabr (approved):

Ein-Tim:
@svengabr Thanks for merging. Could you come back to us in #1094?

svengabr:
Thanks for your contributions! For now, I am not sure how to continue with this. The download page was a work in progress and the "real" download page will be located at: https://e.coronawarn.app/

The purpose of that simple static site is that we need a slim page without any type of tracking. However, the current version that can be seen there is not the final one. It will be a copy of the download page discussed here.

I will make sure your changes will be applied to that little static site.
I would suggest not doing any additional HTML changes here for now.

Have a great day!

Ein-Tim:
@svengabr 

Thank you for letting us know!

> I would suggest not doing any additional HTML changes here for now

Okay agreed. Let's track the other issues separately and see how & when to make the changes (#1092, #1093. #1094)

PR #1130, opened 2021-04-22
Tracelocation qrcode generation

bastianjoel:
I've added a page to the main navigation where you can generate qr codes for the event registration. 

**TODO**
- [x] Add poster template
- [x] Start/end time for events
- [x] Validate form input via javascript (checkout time only numbers, max chars, required inputs)

**EDIT by Open Source Community Support Member**
Rolling display support to show QR code in stores on tablets or digital signage displays:
- [ ] Add refresh interval as a form field to allow automatic rotation of QR codes
- [ ] Put qr code on a separate view after generation to use web app as a rolling display as already suggested, see https://github.com/corona-warn-app/cwa-wishlist/issues/413
- [ ] Read out GET parameters to automatically fill out form 

Support larger stores with printouts:
- [ ] Generate Batch of qr codes for chain of stores  

---
Internal Tracking ID: [EXPOSUREAPP-5880](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5880)


Ein-Tim:
That's amazing @bastianjoel! Thank you so much! I really hope this gets merged!

This would implement https://github.com/corona-warn-app/cwa-wishlist/issues/399

dsarkar:
@bastianjoel Thank you very much, we will have a look at this.

Ein-Tim:
One small point: Testing this locally I can't get it to work on Safari (14.0.3) on my MacBook Air M1 2020 with Big Sur 11.2.3.
This would be a bit annoying for Safari users.

My Safari log shows:
Bildschirmfoto 2021-04-22 um 17 09 23
Bildschirmfoto 2021-04-22 um 17 09 33

bastianjoel:
@Ein-Tim Thank you for that hint. I will have a look at it.

thomasaugsten:
We have to think about this because. We have to give the safety that we do not save the generated QR codes. 
A frontend only calculation would be better.
It is also needed to use the current poster template

Ein-Tim:
@bastianjoel I can confirm that it's working now. 

Two more comments:

1. NPM informs me during the build that the event registration section is very big (294 Kib). The limit seems to be 244 KiB.
2. You can enter anything into the "Checkout after minutes" field, not only numbers but also letters. This should be changed.

Ein-Tim:
@thomasaugsten Does this implementation save the QR-Codes somewhere?

bastianjoel:
@thomasaugsten What do you mean with frontend only calculation? 

Adding the poster shouldn't be a problem. We could just add it to the canvas.

thomasaugsten:
This was my misinterpretation of the form. 

You can find the template here:
[pt-poster-1.0.0.pdf](https://github.com/corona-warn-app/cwa-website/files/6359566/pt-poster-1.0.0.pdf)

Ein-Tim:
The browser now no longer forces me to enter a description, an address or the Checkout after minutes. 
empty

mtb77:
Really great job @bastianjoel ! I added some ideas to the top

bastianjoel:
@Ein-Tim I will add an additional validation layer.

@thomasaugsten Thank you for the information. 

Would it be sufficient to export the poster as PNG? For PDF export I would need to add another library.

fynngodau:
Another enhancement request would be to modify the "Checkout after minutes" field in such a way that it only allows picking values divisible by 15 minutes, and that it displays its value in the format hh:mm.

thomasaugsten:
I would recommend pdf because it is very likely that the poster is print out on A3 and high quality png can be very large

fynngodau:
Sometimes, the site generates a QR code that can be scanned, but not decoded by CWA. I reproduced this by repeatedly generating new QR codes, some of which would work and some of which would cause a "Payload decoding failed" message (using the Android client).

This is one of the QR codes that does not work:

code

Maybe somebody could try to find out what the issue is.

Ein-Tim:
@fynngodau 

I do not experience any problems on my iOS device with this QR-Code.
It says "private Feiere", "Release", "Wishlist repo"

thomasaugsten:
The URL Payload needs to encoded in base64URL not only base64

Ein-Tim:
@thomasaugsten 

Is there a difference between the implementation under iOS and under Android?

@fynngodau gave me some more "broken" QR-Codes and they are not working on his Android device, but on my iOS device I have no problems with these codes.

thomasaugsten:
Yes iOS is filtering not so strict like Android

bastianjoel:
I'm actually using base64URL as of https://tools.ietf.org/html/rfc4648#section-5

Implemented here:
https://github.com/corona-warn-app/cwa-website/pull/1130/files#diff-8de9dfb8d87337a70a7adde57c8eade645e27292e05bc10c31182a2cfe39a5d4R162

thomasaugsten:
You are sure this is replacing all characters not only the first occurence?

Ein-Tim:
Everything looks so good and works as expected now! 
One thing that would be quite cool ist to have a picker for "Checkout after minutes" but mainly for the "Start time" and the "End time". On a Desktop/Laptop it's okay to enter it via the keyboard, but on a tablet or on a phone it's a bit hard. A picker would help here. 
But I don't know how hard this is, so if this would be too much work, it's also great how it is now!
Thank you!

Ein-Tim:
@dsarkar Would you like to approve the workflow so that it can run. 
Than we could see if the CI finds anything broken.

svengabr:
Hello @bastianjoel and everyone,

thank you very much for your contribution here! We took a look at the changes and they are looking awesome. I took over the review of your changes and I will go forward checking the licensing of the new modules.

I will do the following further changes:
- Changing the Desktop design to `col-6` (left form, right side for the QR code)
- Some minor CSS changes in typography
- Maybe some JS changes if the form gets changed AFTER generating the QR code

I am looking forward to deploying the changes to live by end of the day.

src/data/global.json
```diff
@@ -59,6 +59,11 @@
                 }
             },
             "nav": [
```
svengabr, line 61: Navigation looks a little bit broken now on small desktop screens - right before we change to the burger menu. But no blocker for now

package.json
```diff
@@ -23,6 +23,7 @@
     "autoprefixer": "^9.8.6",
     "babel-loader": "^8.2.2",
     "browser-sync": "^2.26.14",
+    "google-protobuf": "^3.15.8",
```
svengabr, line 26: Please check for licensing, MIT is good

package.json
```diff
@@ -41,8 +42,10 @@
     "js-yaml": "^3.14.1",
     "lodash.throttle": "^4.0.8",
     "panini": "^1.7.1",
+    "qrcode": "^1.4.4",
```
svengabr, line 45: Please check for licensing, MIT is good

package.json
```diff
@@ -41,8 +42,10 @@
...
     "rimraf": "^3.0.2",
     "slick-carousel": "^1.8.1",
+    "uint8-to-base64": "^0.2.0",
```
svengabr, line 48: Please check for licensing, MIT is good

package.json
bastianjoel: BSD-3-Clause

Source:
https://www.npmjs.com/package/google-protobuf

package.json
bastianjoel: MIT

Source:
https://www.npmjs.com/package/qrcode

package.json
bastianjoel: ISC

Source:
https://www.npmjs.com/package/uint8-to-base64

src/data/global.json
bastianjoel: Maybe we could change the naming to something shorter for now?

Ein-Tim:
Please remove the `do-not-merge` label here since this PR is already merged.

tik-boomer:
@fynngodau In general steps of 15 minutes should be fine. Please do not strictly limit the duration entry to this. Events generated for Education / Classroom are not generally compatible with this. 

I have suggested a different approach for school / classroom application, https://github.com/corona-warn-app/cwa-wishlist/issues/477#issue-865816494.

svengabr:
Please continue the discussion here:
https://github.com/corona-warn-app/cwa-website/pull/1132

dsarkar:
FYI https://www.coronawarn.app/en/blog/2021-04-23-qr-code-generator/

PR #1148, opened 2021-04-25
Make labels in QR code generator consistent with app

fynngodau:
### Where to find the issue

### Describe the issue
The Place field is called Address. This is inconsistent with the mobile app, where the field is called Place.

In German, the same applies. Additionally, the label is "Beschreibung", not "Bezeichnung" like in the app.

### Additional context

A user was [confused](https://social.tchncs.de/@erAck/106126160685373319) why an Address was required on the website.

### DIFF
--- a/src/data/eventregistration.json
+++ b/src/data/eventregistration.json
@@ -9,7 +9,7 @@
             "fields": {
                 "locationtype": "Place/Event",
                 "description": "Description",
-                "address": "Address",
+                "address": "Place",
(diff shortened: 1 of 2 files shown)


heinezen (approved):
True, let's fix it

fynngodau:
Looking at it again, it now looks a little duplicate, since there is one "Place/Event" field and one "Place" field…

Ein-Tim:
@fynngodau 

I don't think that's a problem since you can't enter anything manually in the "Place/Event" field.

Issue #1153, opened 2021-04-25
Align Start/Ende to android app, fix date pattern to TT.MM.JJJJ


DerVogel2020:
## Where to find the issue
https://www.coronawarn.app/de/eventregistration/
Switch Ort/Event e.g. to Vereinsaktivität.

## Describe the issue

The German "Endzeit" means something different: See https://www.duden.de/rechtschreibung/Endzeit . Maybe "Enduhrzeit" could be used.
While comparing with the android app, I think "Beginn" and "Ende" would be better.

Furthermore, the pattern for the date should be TT.MM.JJJJ.

## Suggested change
While comparing with the android app, I think "Beginn" and "Ende" would be better.

Furthermore, the pattern for the date should be TT.MM.JJJJ.

svengabr:
Should the English version remain unchanged then?

```
"starttime": "Start time",
"endtime": "End time",
```

DerVogel2020:
> Should the English version remain unchanged then?
> 
> ```
> "starttime": "Start time",
> "endtime": "End time",
> ```

Good point. See #1159.

svengabr:
The pull request was merged, thank you very much for cour contribution @DerVogel2020

PR #1154, opened 2021-04-25
Align Start/Ende to android app, fix date pattern to TT.MM.JJJJ

DerVogel2020:
See https://github.com/corona-warn-app/cwa-website/issues/1153


src/data/eventregistration_de.json
```diff
@@ -11,8 +11,8 @@
                 "description": "Bezeichnung",
                 "address": "Ort",
                 "defaultcheckinlength": "Typische Aufenthaltsdauer (Minuten)",
-                "starttime": "Startzeit",
-                "endtime": "Endzeit",
+                "starttime": "Beginn",
```
svengabr, line 14: How about the English version? It still has the old labels
```
"starttime": "Start time",
"endtime": "End time",
```

DerVogel2020: Good point.

Ein-Tim:
Let's continue in #1159 👍

PR #1160, opened 2021-04-26
added location type placeholder and error message

svengabr:
* Location type is no longer pre-filled
* Added error message for empty location type


svengabr:
Internal tracking id `EXPOSUREAPP-6713`

PR #1169, opened 2021-04-28
Remove "other" from QR-Code generator

Ein-Tim:
This PR removes the option to choose "Other" in the QR-Code generator since there is no option for this in the app and thus the app shows an error/unlocalized string.

Related: https://github.com/corona-warn-app/cwa-app-ios/issues/2564

### DIFF
--- a/src/data/eventregistration.json
+++ b/src/data/eventregistration.json
@@ -49,8 +49,7 @@
                 { "id": 10, "name": "Club activities" },
                 { "id": 11, "name": "Private party" },
                 { "id": 12, "name": "Religious service" },
-                { "id": 2, "name": "Other event" },
-                { "id": 0, "name": "Other" }
+                { "id": 2, "name": "Other event" }
             ]
         }
     }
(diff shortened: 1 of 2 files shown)


dsarkar:
@Ein-Tim Thanks

anderes Event ==> Anderes Event

Anyway, I am wondering why the word Event is used instead of Veranstaltung

dsarkar:
@Ein-Tim I confirm, Other must be removed, leads to error on CWA

DerVogel2020:
> anderes Event ==> Anderes Event

This should be fixed for both apps and the website and all occurrances where like "private Feier" or "öffentliches Gebäude" see #2981.

> Anyway, I am wondering why the word Event is used instead of Veranstaltung

See also https://github.com/corona-warn-app/cwa-app-android/issues/2888#issuecomment-824115062

dsarkar (approved):

dsarkar:
@DerVogel2020 @Ein-Tim Merged, since urgent. I see, wording is aligned with CWA. Will do for the moment. Thanks.

dsarkar:
@DerVogel2020 Right, I forgot you mentioned all that already.

DerVogel2020:
@dsarkar just opened https://github.com/corona-warn-app/cwa-wishlist/issues/491 because this regards both apps and the website

PR #1186, opened 2021-05-02
FAQ and screenshots fix

dsarkar:


Ein-Tim:
@dsarkar FYI there are some outdated screenshots live, I'll follow up with an issue.

PR #1196, opened 2021-05-03
Improve eventregistration download button behavior 

### LINKED ISSUE #1136: Pressing 'Download' button in QR code generator before 'Create' yields empty image
### Problem

Downloading the QR code fetches the last created QR code for the resulting image. If no QR code has been generated yet, the resulting image is empty (it returns a 300x150 image with only alpha values).

### Steps to reproduce

1. Go to https://www.coronawarn.app/en/eventregistration/
2. Fill out forms
3. Do **not** select 'Create'
3. Select 'Download'
4. Receive empty image

### Possible fix

Ensure that a QR code is always generated before download.

### LINKED ISSUE #1144: QR-Code should not change when "Download" is pressed
## Where to find the issue
https://www.coronawarn.app/de/eventregistration/

## Describe the issue
The QR-Code changes when the download button is pressed.
This could result in a situation where the person who created the QR-Code checks in to the QR-Code shown on the website after he/she downloaded the QR-Code and thus he/she checks in to another event than the people who check in with the QR-Code which was downloaded.
The event will have the same name and the user can't tell a difference, but it's another event. 
Thus, in case of an infection, the person who checked in to the QR-Code shown on the website won't be warned.

## Suggested change
Deactivate [`b7d04d8` (#1132)](https://github.com/corona-warn-app/cwa-website/pull/1132/commits/b7d04d820370a58e69eb9f060438646f0caf911f) for the "Download" button.

## Related issue 
- #1137

bastianjoel:
This PR resolves #1144 and also resolves #1136. 

With this PR the download button is initially disabled when accessing the page. The disabled attribute is removed when the "create" button is clicked. 

Also I removed the logic which creates a new qr code when clicking the download button. To make it more clear that the download button only downloads the current banner I've repositioned the button under the banner.

### DIFF
--- a/src/assets/js/eventregistration.js
+++ b/src/assets/js/eventregistration.js
@@ -23,6 +23,7 @@ UpdateQRForm();
 document.getElementById('qrform').addEventListener('change', function (e) {
   document.getElementById('eventplaceholder').classList.remove('d-none');
   document.getElementById('eventqrcode').classList.add('d-none');
+  document.getElementById('downloadCode').disabled = true;
 });
 
 document.getElementById('generateQR').addEventListener('click', function (e) {
--- a/src/assets/js/eventregistration.js
+++ b/src/assets/js/eventregistration.js
@@ -31,6 +31,8 @@ document.getElementById('generateQR').addEventListener('click', function (e) {
   if (ValidateQRForm()) {
     GenerateQRCode();
 
+    // Active download button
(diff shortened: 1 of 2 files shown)


Ein-Tim (approved):
Tested locally - Works as expected! LGTM! Thank you!

svengabr (approved):

PR #1197, opened 2021-05-03
add print button to eventregistration

### LINKED ISSUE #1145: Add option to print directly from the website
## Feature description

The [QR-Code generator](https://www.coronawarn.app/en/eventregistration/) should have a button "Print", which opens the print dialog of the browser when it's pressed.

## Problem and motivation
Currently the user has to download the QR-Code before it can be printed. The Corona-Warn-App Android and iOS apps offer a print button directly from inside of the app to make it as easy as possible to print out the code.

bastianjoel:
This PR resolves #1145.

Disabling logic and button position for the new print button are already implemented as in #1196.

### DIFF
--- a/src/assets/js/eventregistration.js
+++ b/src/assets/js/eventregistration.js
@@ -23,6 +23,7 @@ UpdateQRForm();
 document.getElementById('qrform').addEventListener('change', function (e) {
   document.getElementById('eventplaceholder').classList.remove('d-none');
   document.getElementById('eventqrcode').classList.add('d-none');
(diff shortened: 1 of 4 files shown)


Ein-Tim (approved):
Also tested this - LGTM too! 

Just that I get it right, when both #1196 & #1197 are merged, the buttons "Print" and "Download" will be moved below the poster, right? 

Have a nice evening!

bastianjoel:
@Ein-Tim Correct

svengabr (approved):

PR #1200, opened 2021-05-03
Event date time pickers and select fields for defaultcheckinlength

### LINKED ISSUE #1143: Add picker to QR-Code generator
## Feature description

Add a picker to the https://www.coronawarn.app/en/eventregistration/ page.
The fields: 
- "Typical length of stay", - **IMPLEMENTED**
- "Start time" &
- "End time"

could be enhanced by adding a picker to them, which makes it easier to enter the information.

## Problem and motivation
The fields only accept specific formatted inputs (see https://github.com/corona-warn-app/cwa-website/pull/1132#issuecomment-825760191). This can be confusing for the user. With a picker it is ensured that the right format is entered. Also the Corona-Warn-App Android and iOS app use pickers, so the user already knows them.

### LINKED ISSUE #1155: Up/down buttons for length of issue don’t work
## Where to find the issue

https://www.coronawarn.app/de/eventregistration/

## Describe the issue

Go to https://www.coronawarn.app/en/eventregistration/
Click the up or down button in the “Typical length of stay (minutes)” field:

grafik

Even if you enter a value like 0 or 15 I cannot press the buttons, respectively they do nothing.

Furthermore, I can enter a negative number even though the `min=0` value is set in HTML:
grafik

## Suggested change

Fix the issue.

Firefox 88

### LINKED ISSUE #1192: QR-Code Generation Typical length of stay is not strongly enforced
## Where to find the issue

## Describe the issue
Typical length of stay is not strongly enforced you can override manually the max value of 3600

## Suggested change
Check in NodeJs if the value over the limit and return an error

bastianjoel:
This PR adds cleave.js to the date and time fields which forces users to use the given date pattern and adds separators automatically which resolves #1143. 

Also the input field defaultcheckinlength got replaced by two select fields for hours and minutes. This resolves #1192 because with the select fields it is not possible anymore to select more than 3600 minutes. In addition this resolves #1155 because the number input field got replaced by select fields.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -60,6 +60,11 @@ Licensor: Andrea Giammarchi and other contributors
 Website: https://www.npmjs.com/package/uint8-to-base64
 License: ISC
 
+Component: cleave.js
+Licensor: Max Huang
+Website: https://nosir.github.io/cleave.js/
+License: Apache-2.0 License
+
 --------------------------------------------------------------------------------
 Apache License 2.0 (Roboto Font)
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -3631,6 +3631,11 @@
         }
       }
     },
+    "cleave.js": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/cleave.js/-/cleave.js-1.6.0.tgz",
+      "integrity": "sha512-ivqesy3j5hQVG3gywPfwKPbi/7ZSftY/UNp5uphnqjr25yI2CP8FS2ODQPzuLXXnNLi29e2+PgPkkiKUXLs/Nw=="
+    },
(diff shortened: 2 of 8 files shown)


svengabr:
@bastianjoel Thanks for your contribution!

We were discussing internally if we should move forward with a datetime picker instead of user input. Do you think this would make more sense?

I took a look at the mobile layout and there is still a small issue:

2021-05-04 10_17_57-Window

Once this issue is resolved, I am fine to approve the changes for a golive.

However, its most likely the functionality will get replaced by the mentioned datepicker in the near future.

svengabr (changes requested):
Please take a look at the mobile layout for Event type

svengabr (approved):

svengabr:
Looking good @bastianjoel , thanks!

PR #1211, opened 2021-05-05
Add incompatibility_warning FAQ

fynngodau:
The Android version of CWA, from version 2.2 on, [contains a link](https://github.com/corona-warn-app/cwa-app-android/blob/21c2473dab091482a162f208134fef38fd0e8f6a/Corona-Warn-App/src/main/res/values/strings.xml#L2022) to an `incompatibility_warning` FAQ entry (PR: https://github.com/corona-warn-app/cwa-app-android/pull/2873).

This entry does not exist yet. Here's my suggestion for what it could be.

There's some more details in our [CCTG original](https://codeberg.org/corona-contact-tracing-germany/cwa-android#there-s-a-card-on-my-home-screen-telling-me-that-my-device-is-only-partially-compatible-what-does-this-mean) which I left out to make it more suitable for the CWA audience (specifically, no advice to install custom ROMs on Fairphone 2 ;).

German | English
---|---
image | image

---
Internal Tracking ID: [EXPOSUREAPP-7118](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7118)

PR #1216, opened 2021-05-06
Replace Lokation with Ort

fynngodau:
As @Ein-Tim noted in https://github.com/corona-warn-app/cwa-website/pull/1211#discussion_r626730347, "Lokation" sounds odd and is not a good translation for the term "location". Since the app is using "Ort", I'd suggest the FAQ to consistently use this translation too.


dsarkar (approved):

dsarkar:
@fynngodau @Ein-Tim Thanks!

PR #1225, opened 2021-05-08
Remove "International phone numbers" from the title of "international_phone_numbers"

Ein-Tim:
This PR removes "International phone numbers" from the title of the FAQ entry "internation_phone_numbers", since we also store the national phone numbers there.
Also, a "." was moved out of a link in "who_can_help". Instead, the "." is now normal text (;


dsarkar:
@Ein-Tim Thanks. What do you think about "national and International phone numbers" in the title? Just in case somebody uses the search function?

Ein-Tim:
@dsarkar Yeah, we could do this but I think if somebody wants to find out the phone numbers he/she will simply search for "phone numbers". 
But as you like, I can change it if you want!

dsarkar (approved):

dsarkar:
@Ein-Tim ok, agreed. We adopt your suggestion.

Ein-Tim:
Thanks @dsarkar 

Have a nice sunny Sunday!

PR #1256, opened 2021-05-12
Screenshots - fix iOS EN fig 1

dsarkar:


MikeMcC399:
@dsarkar 
It should be possible to delete the branch release-screenshots.

PR #1275, opened 2021-05-17
Add introduction version info for BLE compatibility check (closes #1274)

### LINKED ISSUE #1274: FAQ: Bluetooth incompatibility does not mention introduction version
## Where to find the issue

**[Google/Android] A card on my Start Screen is notifying me that my device is not or only partially compatible. What does this mean?**
https://www.coronawarn.app/en/faq/#incompatibility_warning

**[Google/Android] Auf meinem Hauptbildschirm werde ich gewarnt, dass mein Gerät nicht oder nicht vollständig kompatibel ist. Was bedeutet das?**
https://www.coronawarn.app/de/faq/#incompatibility_warning

## Describe the issue

The introduction of the functionality to check Bluetooth Low Energy compatibility has not been announced as tied to a particular release anywhere. This leaves users updating to CWA Android 2.2.x (the version where it was introduced) who are seeing an incompatibility message for the first time wondering if there is a bug in the app.

- The blog announcement https://www.coronawarn.app/en/blog/2021-05-12-corona-warn-app-version-2-2/ has no mention of the BLE compatibility check.
- The Android [release_info_strings.xml](https://github.com/corona-warn-app/cwa-app-android/blob/release/2.2.x/Corona-Warn-App/src/main/res/values/release_info_strings.xml) mentions only:
o "Record Troubleshooting Logs"
o "Create Rapid Test Profiles"
So also here in the App Information > New Features section of the app there is no mention of the new checks.

## Suggested change

Explain in the FAQ that this is a new check introduced with version 2.2 of the Android app. You could say that the Google Play Store checked previously for the presence of Bluetooth before allowing for installation, however the new check does a more detailed compatibility check compared to previous versions.

The suggestion was already made in https://github.com/corona-warn-app/cwa-website/pull/1211#pullrequestreview-652528871 by @Ein-Tim and the subject was raised again in https://github.com/corona-warn-app/cwa-app-android/issues/3170#issuecomment-841830041.

Maybe @fynngodau as the original author would like to pick this up as a PR? Or perhaps @Ein-Tim, who raised the flag previously? Please let me know if you need assistance.

MikeMcC399:
This PR adds the information to the FAQ entries
https://www.coronawarn.app/en/faq/#incompatibility_warning and
https://www.coronawarn.app/de/faq/#incompatibility_warning 

that the Bluetooth Low Energy compatibility checks were introduced with version 2.2 of CWA. The PR addresses issue https://github.com/corona-warn-app/cwa-website/issues/1274.

Without this information users may think that there is a bug in the app if they had previous versions of the app installed, have updated and are seeing a warning message for the first time.

Incidentally also one typo in the EN version of the FAQ entry is corrected at the same time.

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -493,10 +493,11 @@
                         "active": true,
                         "textblock": [
                             "If you see such a notice, your device is not able to send any random IDs via Bluetooth. This can be caused by a restriction due to your combination of hardware and software on your device. Because of it, the Corona-Warn-App cannot use Bluetooth Low Energy in its so-called “peripheral mode”.",
-                            "In case you are infected and you decide to warn others, only users who were checked in to the same event or the same location, at the same time as you were, can be warned. Your own risk assesment is <b>not affected</b>. All other features of the app are also not affected.",
+                            "In case you are infected and you decide to warn others, only users who were checked in to the same event or the same location, at the same time as you were, can be warned. Your own risk assessment is <b>not affected</b>. All other features of the app are also not affected.",
                             "Please ensure that you are using the latest version of the operating system available for your smartphone. If the problem persists, please contact your device manufacturer.",
                             "If the card shows that your device is completely incompatible, the cause could be that your device does not contain a Bluetooth chip. Only in this case, you cannot exchange random IDs with others and risk assessment can only base its calculation on locations and events you checked in to.",
-                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction in any case</b> and only proximity-based ID recording is affected."
+                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction in any case</b> and only proximity-based ID recording is affected.",
+                            "Bluetooth compatibility checking and warnings were introduced with Android Corona-Warn-App version 2.2."
                         ]
                     }
                 ]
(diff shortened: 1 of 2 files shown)


src/data/faq.json
```diff
@@ -493,10 +493,11 @@
...
                             "Please ensure that you are using the latest version of the operating system available for your smartphone. If the problem persists, please contact your device manufacturer.",
                             "If the card shows that your device is completely incompatible, the cause could be that your device does not contain a Bluetooth chip. Only in this case, you cannot exchange random IDs with others and risk assessment can only base its calculation on locations and events you checked in to.",
-                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction in any case</b> and only proximity-based ID recording is affected."
+                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction in any case</b> and only proximity-based ID recording is affected.",
+                            "Bluetooth compatibility checking and warnings were introduced with Android Corona-Warn-App version 2.2."
```
fynngodau, line 500: This sounds a bit like Bluetooth warnings were added with this version (if you read "compatibility checking" as one unit)

MikeMcC399: I started out with a more detailed description and then I simplified it, so I hope that the simplification doesn't cause any confusion! I was a bit surprised that the PR got merged so quickly, but now that it is merged, I would leave it like it is, unless it is going to cause problems.

PR #1278, opened 2021-05-17
 update dsfa - Bericht zur Datenschutz-Folgenabschätzung

dsarkar:


dsarkar:
Fixes https://github.com/corona-warn-app/cwa-website/issues/1066

dsarkar:
The relevant subsection is now 5.6.12

MikeMcC399:
@dsarkar 
> Fixes #1066

I can see that changes were made, but I don't think that the issue is fixed. 😞

Issue #1227, opened 2021-05-09
Typo: Check-In


DerVogel2020:
## Where to find the issue
https://www.coronawarn.app/de/eventregistration/

## Describe the issue
While grepping for https://github.com/corona-warn-app/cwa-app-android/issues/3120 I found 2 typos: Check-In is used instead of Check-in.

## Suggested change
Change the text to Check-in.

DerVogel2020:
Fixed in #1228 .

dsarkar:
@DerVogel2020 Thanks.


---
Corona-Warn-App Open Source Team

PR #1283, opened 2021-05-17
Fix check in capitalisation

DerVogel2020:
Found two more "Check-In" in the german faq and the android-screenshot-section.
Still to #1227.


svengabr (approved):

src/data/screenshots_de.json
```diff
@@ -55,7 +55,7 @@
                 ]
             },
             {
-                "title": "Check-In",
+                "title": "Check-in",
                 "anchor": "android_check_in",
                 "images": [
                     { "filename": "de/android/check_in_host_view_my_qr_codes", "alt": "Check In image 1 of 8" },
```
svengabr, line 61: Might consider in a future update to also update the alternative texts of the images. Same applies for iOS

Issue #1288, opened 2021-05-18
FAQ - Revise articles about incompatibility warning card


heinezen:
## Where to find the issue
https://www.coronawarn.app/de/faq/#incompatibility_warning
https://www.coronawarn.app/en/faq/#incompatibility_warning

## Describe the issue

We agreed in https://github.com/corona-warn-app/cwa-website/pull/1211#discussion_r627438916 to split the article in two new articles to make it more user-friendly. The problem with the current article are:

* Partial and full incompatibility are mixed which can lead to the wrong assumption that the app doesn't work at all
* A lot of users wrongly assumed that the incompatibility was **caused** by release 2.2, while in fact it was only made visible by the new checks (see https://github.com/corona-warn-app/cwa-app-android/issues/3170)
* The wording/terminology used in the article turned out to be too technical in some places as we know from app store reviews (e.g. peripheral mode, BLE, bluetooth chip).

## Suggested change

Make two new FAQ articles for the changes introduced in 2.3: https://github.com/corona-warn-app/cwa-app-android/pull/3101

- `part_incompat` should explain partial incompatability
- `full_incompat` should explain full incompatibility

----

The version number was added by @MikeMcC399 in https://github.com/corona-warn-app/cwa-website/pull/1275

However, the sentence could be improved (https://github.com/corona-warn-app/cwa-website/pull/1275#discussion_r633271249). Maybe we should start the first sentence with this information:

"Since CWA release 2.2, the app will check if your device is fully capable to exchange random IDs..." 

----
We may need to drop most technical terms and focus more more on the outcome of incompatibility, e.g. instead of

> Because of it, the Corona-Warn-App cannot use Bluetooth Low Energy in its so-called “peripheral mode”.

we should use

> This means that your device is unable to **send** random IDs to other devices. However, you are still able to **receive** random IDs. Your own risk assesment is therefore not affected.

----
The term "incompatibility" can also be a bit problematic IMO, so we maybe have to choose a more "tame" term in the app and the FAQ. 

----
**Store reviews for context**

[...]

PR #1291, opened 2021-05-23
Revise incompatibility warning FAQs

fynngodau:
I implemented the ideas from #1288 and split up the `incompatibility_warning` article into `part_incompat` and `full_incompat`.

To avoid duplicate fixes in response to your feedback, I am only providing an English version until all your concerns with this version are addressed.

image

image

I added a redirect from `incompatibility_warning` to `part_incompat`.


src/data/faq.json
```diff
@@ -491,16 +491,30 @@
...
+                            "Since version 2.2, the Corona-Warn-App for Android contains a check which determines whether your device is fully capable of all Bluetooth-related features which the app needs to operate normally. If you are seeing this message, the operating system of your device is lacking a functionality that Corona-Warn-App normally uses.",
+                            "Due to this, your device is not able to <i>send</i> any random IDs via Bluetooth. However, you are still able to <i>receive</i> random IDs. Therefore, <b>your own risk assesment is not affected</b>.",
+                            "The only limitation which this incompatibility causes is that, in case you are infected and you decide to warn others, only users who were checked in to the same event or the same location, at the same time as you were, can be warned. All other features of the app are not affected.",
+                            "To resolve the problem, please ensure that you are using the latest version of the operating system available for your smartphone. If this doesn't help, please contact your device manufacturer.",
```
Ein-Tim, line 501: Maybe add a link to https://support.google.com/android/answer/3094742?hl=en&ref_topic=7313011 to "contact your device manufacturer"?

src/data/faq.json
```diff
@@ -491,16 +491,30 @@
...
+                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction</b> and only proximity-based ID recording is affected.",
```
Ein-Tim, line 502: I'd say "are fully functional", since we're talking about more than one feature, or?

src/data/faq.json
```diff
@@ -491,16 +491,30 @@
...
+                            "In version 2.2, a check regarding Bluetooth capabilities of your device was added to the Corona-Warn-App for Android. If a warning message that links to this FAQ entry is shown, your device is not capable of Bluetooth Low Energy and can therefore neither send nor receive random IDs.",
+                            "Your risk assesment will only be based on events and locations you were checked in to.",
+                            "Possible reasons for this message are:<ul><li>Your device does not contain the hardware that is required to use Bluetooth Low Energy.</li><li>Your device does not contain the software required to use the Bluetooth Low Energy capabilities of your hardware.</li><li>You are running the device in the Android Emulator.</li></ul>",
+                            "When in doubt, please ensure that you are using the latest version of the operating system available for your smartphone. For further assistance, please contact your device manufacturer.",
```
Ein-Tim, line 516: Reg. "contact your device manufacturer", see my other comment.

src/data/faq.json
```diff
@@ -491,16 +491,30 @@
...
+                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction</b> and only proximity-based ID recording is affected."
```
Ein-Tim, line 517: Same here reg. "is" -> "are"

Ein-Tim (commented):
One grammatically point and one suggestion. Entry looks good to me! 
Nice that you added the redirect!

src/data/faq.json
fynngodau: It seems like a helpful resource to me, but I'm not sure whether we can link it and would like to have an official confirmation before implementing this.

src/data/faq.json
fynngodau: From my feeling, since I'm separating the first batch and "retrieving […] test results" through an "as well as", the verb would grammatically refer only to the test result feature, but honestly I have no idea how to find out which version is correct :sweat_smile:

Ein-Tim: Maybe @MikeMcC399 would like to help out here? 🙂

src/data/faq.json
Ein-Tim: Okay, let's wait for confirmation then. This is already linked to from https://www.coronawarn.app/en/faq/#various_device_problems, so shouldn't be a problem, I assume.

heinezen:
I have a few general remarks for improvement
----
The error message ("incompatibility warning") should be part of the entry heading. This way, search engines will hopefully lead  users directly to the FAQ entry if they search for the warning online.

----

Using a more active language and shorter sentences is encouraged

Example:

> Since version 2.2, the Corona-Warn-App for Android contains a check which determines whether your device is fully capable of all Bluetooth-related features which the app needs to operate normally.

can be improved to

> Since version 2.2, the Corona-Warn-App **checks** if your device **supports** all Bluetooth-related features normally used by the app.

----
If possible, we should get straight to the point

Example:

> If you are seeing this message, the operating system of your device is lacking a functionality that Corona-Warn-App normally uses.

we can merge that sentence with the next paragraph

> If you are seeing this message, the operating system of your device **lacks the ability to send random IDs** [to devices in your proximity]. [...]

The reason is that we want to get to the ">your own risk assesment is not affected" part quickly.

----
"Schachtelsatz" construction should be avoided, especially in english, but it's also a good idea for the german entry.

> The only limitation which this incompatibility causes is that, in case you are infected and you decide to warn others, only users who were checked in to the same event or the same location, at the same time as you were, can be warned.

The way we could split this is by separating the information communicated here into separate sentences:

1. What is the limitation/What is limited? (warning people who were near you)
2. What does still work? (warning through Check-In functionality)

src/data/faq.json
heinezen: It's a good idea to include the link IMO

src/data/faq.json
```diff
@@ -491,16 +491,30 @@
...
-                            "Bluetooth compatibility checking and warnings were introduced with Android Corona-Warn-App version 2.2."
+                            "Since version 2.2, the Corona-Warn-App for Android contains a check which determines whether your device is fully capable of all Bluetooth-related features which the app needs to operate normally. If you are seeing this message, the operating system of your device is lacking a functionality that Corona-Warn-App normally uses.",
+                            "Due to this, your device is not able to <i>send</i> any random IDs via Bluetooth. However, you are still able to <i>receive</i> random IDs. Therefore, <b>your own risk assesment is not affected</b>.",
+                            "The only limitation which this incompatibility causes is that, in case you are infected and you decide to warn others, only users who were checked in to the same event or the same location, at the same time as you were, can be warned. All other features of the app are not affected.",
```
MikeMcC399, line 500: Suggest:
"All other features of the app are not affected." =>
"None of the other app features are affected."

src/data/faq.json
MikeMcC399: This sentence seems redundant, because the message has already been given in L500. 

Grammatically it is a little awkward because the main sentence says:
"Both Check-in and Contact Journal functionality is fully functional without restriction and only proximity-based ID recording is affected."
functionality / functional appears twice and "fully" and "without restriction" is a repeat.

I would tend to just drop this sentence.

src/data/faq.json
```diff
@@ -491,16 +491,30 @@
...
+                        ]
+                    },
+                    {
+                        "title": "[Google/Android] A card on my Start Screen is notifying me that my device is incompatible. What does this mean?",
+                        "anchor": "full_incompat",
+                        "active": true,
+                        "textblock": [
+                            "This does not usually happen on unmodified end-user devices. Therefore, this section doesn't apply to the vast majority of devices. Usually, your device would only be <a href='#part_incompat' target='_blank' rel='noopener'>partially incompatible</a>.",
+                            "In version 2.2, a check regarding Bluetooth capabilities of your device was added to the Corona-Warn-App for Android. If a warning message that links to this FAQ entry is shown, your device is not capable of Bluetooth Low Energy and can therefore neither send nor receive random IDs.",
+                            "Your risk assesment will only be based on events and locations you were checked in to.",
```
MikeMcC399, line 514: Add after L514
"The risk assessment cannot take account of other close-by smartphones because it cannot detect them."

src/data/faq.json
MikeMcC399: Suggest also here leaving out this problematic sentence. The message should be covered by L514.

MikeMcC399 (changes requested):
I made some suggestions for change.

src/data/faq.json
fynngodau: @heinezen suggested to add this information twice in https://github.com/corona-warn-app/cwa-website/pull/1211#discussion_r627500614.

@MikeMcC399, I think you are right with considering the sentence without the additional sentence part starting with "as well as", and then it needs to be "Both […] are".

Lu9Bro:
The revised text looks good to me, thanks.

In one section, the text reads "To resolve the problem, please ensure that you are using the latest version of your smartphone's operating system. If this doesn't help, please contact your device manufacturer."

Do you think it would help to add two sentences here like 
"(Reverting to an older version of CWA does not help, because the technical incompatibility existed in previous versions, too. Just it was not visible to the user.)"    ?

This is meant for users like in #1288 
> Seit dem letzten Update v. 10.05. ist mein Bluetooth Chip plötzlich inkompatibel. Das kann ja wohl nicht sein. Mit der Vorgängerversion wieder alles okay. Geht das bei der Corona App jetzt genauso wie bei den meisten Banking Apps? Jedes Update eine Verschlimmbesserung?

But I’m uncertain if this suggestion would be too much.

fynngodau:
@Lu9Bro Since there is no official downgrade procedure in both Google Play as well as in the stock Android system, I think the information that the check was added in 2.2 is sufficient. However some users might try this anyway… what do the others think?

src/data/faq.json
```diff
@@ -491,16 +491,29 @@
...
+                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, are fully functional without restriction</b> and only proximity-based ID recording is affected.",
+                            "<hr/>",
+                            "Technical explanation: Your hardware is capable of Bluetooth Low Energy, but your Bluetooth driver is missing a capability called “peripheral mode”. This can only be fixed by your device manufacturer through an update of your operating system. Because the capability is missing, the Corona-Warn-App cannot cause your device to periodically broadcast its own random IDs via Bluetooth."
+                        ]
+                    },
+                    {
+                        "title": "[Google/Android] An “Incompatibility warning” card on my Start Screen is indicating that my device is not compatible with Exposure Notifications att all. What does this mean?",
```
MikeMcC399, line 507: "att" => "at"

src/data/faq.json
```diff
@@ -491,16 +491,29 @@
...
+                        "anchor": "full_incompat",
+                        "active": true,
+                        "textblock": [
+                            "This does not usually happen on unmodified end-user devices. Therefore, this section doesn't apply to the vast majority of devices. Most devices are, at most, <a href='#part_incompat' target='_blank' rel='noopener'>partially incompatible</a>.",
+                            "Since version 2.2, the Corona-Warn-App for Android checks the Bluetooth capabilities of your device. If a warning message that links to this FAQ entry is shown, your device is not capable of Bluetooth Low Energy and can therefore neither send nor receive random IDs.",
+                            "Your risk assesment will only be based on events and locations you were checked in to.",
```
MikeMcC399, line 513: "assesment" => "assessment"

src/data/faq.json
```diff
@@ -491,16 +491,29 @@
...
-                            "If the card shows that your device is completely incompatible, the cause could be that your device does not contain a Bluetooth chip. Only in this case, you cannot exchange random IDs with others and risk assessment can only base its calculation on locations and events you checked in to.",
-                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, is fully functional without restriction in any case</b> and only proximity-based ID recording is affected.",
-                            "Bluetooth compatibility checking and warnings were introduced with Android Corona-Warn-App version 2.2."
+                            "Since version 2.2, the Corona-Warn-App for Android checks whether your device is fully capable of all needed Bluetooth-related features. If you are seeing this message, the operating system of your device lacks the ability to <i>send</i> random IDs via Bluetooth. However, it is still able to <i>receive</i> random IDs. Therefore, <b>your own risk assesment is not affected</b>.",
```
MikeMcC399, line 498: "assesment" => "assessment"

MikeMcC399 (changes requested):
I marked up some typos.

fynngodau:
I have added a German translation now.

src/data/faq_de.json
```diff
@@ -493,16 +493,29 @@
                         ]
                     },
                     {
-                        "title": "[Google/Android] Auf meinem Hauptbildschirm werde ich gewarnt, dass mein Gerät nicht oder nicht vollständig kompatibel ist. Was bedeutet das?",
-                        "anchor": "incompatibility_warning",
+                        "title": "[Google/Android] Eine Karte mit einer „Inkompatibilitätswarnung” auf meinem Hauptbildschirm benachrichtigt mich, dass mein Gerät nur teilweise kompatibel ist. Was bedeutet das?",
```
MikeMcC399, line 496: Maybe use the term "Startseite" displayed under the house icon by the app instead of "Hauptbildschirm"?

src/data/faq_de.json
```diff
@@ -493,16 +493,29 @@
...
+                            "<b>Die Funktionalitäten zum Einchecken, zum Abrufen von sowohl PCR- als auch Schnelltestergebnissen sowie das Kontakttagebuch funktionieren trotz der Warnmeldung ohne Einschränkungen</b> und nur die Funktion zur nähebasierten Aufzeichnung von Zufalls-IDs ist betroffen.",
+                            "<hr/>",
+                            "Technische Erklärung: Ihre Hardware unterstützt Bluetooth Low Energy, aber Ihrem Bluetooth-Treiber fehlt eine Funktion namens „peripheral mode”. Dies kann nur von Ihrem Gerätehersteller durch ein Systemupdate behoben werden. Weil diese Funktion nicht verfügbar ist, kann die Corona-Warn-App Ihr Gerät nicht periodisch eigene Zufalls-IDs aussenden lassen."
+                        ]
+                    },
+                    {
+                        "title": "[Google/Android] Eine Karte mit einer „Inkompatibilitätswarnung” auf meinem Hauptbildschirm zeigt an, dass mein Gerät nicht mit Exposure Notifications kompatibel ist. Was bedeutet das?",
```
MikeMcC399, line 509: Same comment about "Hauptbildschirm" here too.

MikeMcC399 (changes requested):
Suggestion about using the term "Startseite".

src/data/faq_de.json
fynngodau: Both terms are used throughout the FAQ, but I agree that Startseite, as a term, is a bit better.

src/data/faq.json
```diff
@@ -491,16 +491,29 @@
...
+                        ]
+                    },
+                    {
+                        "title": "[Google/Android] An “Incompatibility warning” card on my Start Screen is indicating that my device is not compatible with Exposure Notifications at all. What does this mean?",
+                        "anchor": "full_incompat",
+                        "active": true,
+                        "textblock": [
+                            "This does not usually happen on unmodified end-user devices. Therefore, this section doesn't apply to the vast majority of devices. Most devices are, at most, <a href='#part_incompat' target='_blank' rel='noopener'>partially incompatible</a>.",
+                            "Since version 2.2, the Corona-Warn-App for Android checks the Bluetooth capabilities of your device. If a warning message that links to this FAQ entry is shown, your device is not capable of Bluetooth Low Energy and can therefore neither send nor receive random IDs.",
+                            "Your risk assessment will only be based on events and locations you were checked in to.",
```
fynngodau, line 513: Having thought about this again, since risk assessment is toggled together with Exposure Logging, you can't actually enable it (on devices without Bluetooth), can you?

MikeMcC399: @fynngodau 
I can confirm on Pixel 3a API 29 emulator (which has no Bluetooth support) that it is **not** possible to activate exposure logging. Attempting to activate produces a Cause: 3 error. Clicking on the DETAILS button in the emulator produces no detailed error output.

This probably needs to be raised as an issue in https://github.com/corona-warn-app/cwa-app-android, because the text output "You can send and receive warnings that result from check-ins." by the app is inaccurate. I will leave that to you unless you say you want me to do something. Hopefully there will be very few users affected by this.

I don't have any physical Android device without Bluetooth to test on. The emulators cannot always be relied on to show the exact same behavior that a physical device shows, so please take the results of my emulator test with caution.

MikeMcC399: I am seeing some entries in the log file such as
`2021-05-28T05:38:30.598Z  V/TracingRepository: TracingProgress: TracingProgress.IsCalculating, isExposureDetecting=false, isEWDownloading=false, isEWCalculatingRisk=true, isPTDownloading=false, isPTCalculatingRisk=false`
EW = Exposure Window
PT = Presence Tracing
I was not checked into any events in my test though.
Maybe it does work without activating exposure logging?

heinezen:
@fynngodau I would take over now and make remaining in preparation for the next release. So if you still have local changes to push, the best time to do that is now :)

fynngodau:
> So if you still have local changes to push, the best time to do that is now :)

This PR is all I have at the moment ;)

src/data/faq.json
```diff
@@ -491,16 +491,16 @@
...
-                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, are fully functional without restriction</b> and only proximity-based ID recording is affected.",
+                            "Since version 2.2, the Corona-Warn-App for Android checks whether your device supports all necessary Bluetooth features. If you are seeing this message, the operating system of your device lacks the ability to <i>send</i> random IDs via Bluetooth. However, it is still able to <i>receive</i> random IDs. <b>Your own risk assessment is therefore not affected</b>.",
+                            "The only limitation caused by this is that you cannot warn contacts discovered via Bluetooth when sharing a positive test result. However, users who were checked in to the same event or the same location, at the same time as you were, will still be warned. The Check-In functionality, the retrieval of PCR and rapid test results as well as the contact journal featue are not affected and can be used normally.",
```
MikeMcC399, line 499: "featue" => "feature"

src/data/faq_de.json
```diff
@@ -493,16 +493,16 @@
...
-                            "<b>Die Funktionalitäten zum Einchecken, zum Abrufen von sowohl PCR- als auch Schnelltestergebnissen sowie das Kontakttagebuch funktionieren trotz der Warnmeldung ohne Einschränkungen</b> und nur die Funktion zur nähebasierten Aufzeichnung von Zufalls-IDs ist betroffen.",
+                            "Seit Version 2.2 prüft die Corona-Warn-App für Android, ob Ihr Gerät alle benötigten Bluetooth-Funktionen unterstützt. Wenn Sie diese Nachricht sehen, kann das Betriebssystem Ihres Geräts keine Zufalls-IDs per Bluetooth <i>versenden</i>, allerdings jedoch <i>empfangen</i>. <b>Ihre eigene Risiko-Ermittlung ist daher nicht betroffen</b>.",
+                            "Die einzige dadurch verursachte Einschränkung ist, dass Sie Ihre Bluetooth-Kontakte beim Teilen Ihres positiven Testergebnisses nicht warnen können. Personen, die gleichzeitig mit Ihnen an einem Ort oder einem Event eingecheckt waren, werden jedoch weiterhin gewarnt. <b>Die Check-In-Funktion, das Abrufen von PCR- und Schnelltestergebnissen sowie die Nutzung des Kontakttagebuch sind nicht von der Inkompatibilität betroffen und können normal genutzt werden</b>.",
```
MikeMcC399, line 501: "des Kontakttagebuch" => "des Kontakttagebuchs"

MikeMcC399 (changes requested):
I marked up a couple of typos.

heinezen:
I made quite a few changes to the existing text:

* More precise definition of the problem
* Merge the last paragraph (before the technical explanation) into the second paragraph
* Align text and wording between the two FAQ entries
* Remove first paragraph of `full_incompat` because it could be confusing
* Add technica explanation to `full_incompat`

src/data/faq.json
```diff
@@ -504,16 +504,17 @@
...
                             "Possible reasons for this message are:<ul><li>Your device does not contain the hardware that is required to use Bluetooth Low Energy.</li><li>Your device does not contain the software required to use the Bluetooth Low Energy capabilities of your hardware.</li><li>You are running the app in the Android Emulator.</li></ul>",
-                            "When in doubt, please ensure that you are using the latest version of your smartphone's operating system. For further assistance, please <a href='https://support.google.com/android/answer/3094742?hl=en'>contact your device manufacturer</a>.",
-                            "<b>Both Check-in and Contact Journal functionality, as well as retrieving PCR and rapid test results, are fully functional without restriction</b> and only proximity-based ID recording is affected."
+                            "To resolve the problem, please update to the latest version of your smartphone's operating system. If this doesn't help, please <a href='https://support.google.com/android/answer/3094742?hl=en'>contact your device manufacturer</a>.",
```
fynngodau, line 514: This does not apply to 2 out of 3 possible problems that are mentioned above. This was the reason that I chose a different phrasing here.

But I think that in the third case, it is already clear to users what they are doing ;)

src/data/faq.json
```diff
@@ -504,16 +504,17 @@
                         ]
                     },
                     {
-                        "title": "[Google/Android] An “Incompatibility warning” card on my Start Screen is indicating that my device is not compatible with Exposure Notifications at all. What does this mean?",
+                        "title": "[Google/Android] A message “Incompatibility warning” on my Start Screen indicates that my device is not compatible with Exposure Notifications at all. What does this mean?",
                         "anchor": "full_incompat",
                         "active": true,
                         "textblock": [
-                            "This does not usually happen on unmodified end-user devices. Therefore, this section doesn't apply to the vast majority of devices. Most devices are, at most, <a href='#part_incompat' target='_blank' rel='noopener'>partially incompatible</a>.",
```
fynngodau: I understand that this is not a great introduction to the entry. However, I was worried that users could scroll down to the next entry by accident (also specifically since it now looks very similar), and falsely think that their device cannot bluetooth at all.

fynngodau (commented):
I like your changes @heinezen. Here's some thoughts on two of them.

src/data/faq.json
MikeMcC399: Perhaps instead of saying
"To resolve the problem"
write
"To resolve any software incompatibility" ?

heinezen: I figured that the people who use Android Emulator could probably figure out that it's not a problem for them :D

src/data/faq.json
heinezen: My problem with the introduction sentence is that for someone who reads this without knowing about the nature of the incompatibility, it would be even more confusing. Because then it is not really clear which variant (partial or full) actually applies to the user.

I agree that this could be a problem if a user scrolls down by accident or ends up reading the wrong entry when they use the FAQ search, but that's not something that can be easily fixed with more text and links. Making the warning messages distinct would be one way to solve it I guess.

Ein-Tim:
I don't know what's the plan under Android but iOS version 2.3 has been released today.

PR #1316, opened 2021-05-30
add whitespaces between z. and B. as it ist in many other occurences in german faq

DerVogel2020:


MikeMcC399:
@DerVogel2020 
You might need to put in `&nbsp;` if you do that. I'm not sure about the convention for this, but maybe you have some references? Is it OK for z. and B. to be split at the end of a line?

DerVogel2020:
> You might need to put in `&nbsp;` if you do that. I'm not sure about the convention for this, but maybe you have some references? Is it OK for z. and B. to be split at the end of a line?

Sure, you're right. Reference is https://www.duden.de/node/212250/revision/212286 and https://www.duden.de/node/208327/revision/243442 , there always with whitespace.

svengabr (approved):

svengabr:
Thanks, @DerVogel2020 , I have merged your changes.

Cheers!

DerVogel2020:
Thanks.

PR #1336, opened 2021-06-06
Replace "on" with "," (Fixes #1201)

### LINKED ISSUE #1201: Translate subtitles to blogentry
## Where to find the issue
https://www.coronawarn.app/de/blog/
Hanna Heine on 2. Mai 2021 and all other blog entrys

## Describe the issue
Hanna Heine on 2. Mai 2021 has to be Hanna Heine am 2. Mai 2021

## Suggested change
I think this should be fixed in https://github.com/corona-warn-app/cwa-website/blob/master/src/services/blog-processor.js
There in https://github.com/corona-warn-app/cwa-website/blob/5774e095cf51cf8986a0cce9e396f49cee22dc18/src/services/blog-processor.js#L53 is the hardcoded "on".
In https://github.com/corona-warn-app/cwa-website/blob/5774e095cf51cf8986a0cce9e396f49cee22dc18/src/services/blog-processor.js#L128 it's also hardcoded in the english date-string.

_Maybe_ it has to be removed in L53 and added in https://github.com/corona-warn-app/cwa-website/blob/5774e095cf51cf8986a0cce9e396f49cee22dc18/src/services/blog-processor.js#L129 with the word "am".
_Or_ otherwise generate the div-Element in L53 dependent on the language.

Ein-Tim:
This PR fixes what has been described in #1201.

I tried to do it like this: `{{#if lang_de}}am{{else}}on{{/if}}` but this only worked on a individual blog post but not on https://www.coronawarn.app/en/blog/ for some reason, so the simplest solution I saw was to replace "on" with a ",", which works in both languages.

### DIFF
--- a/src/services/blog-processor.js
+++ b/src/services/blog-processor.js
@@ -50,7 +50,7 @@ const generateBlogEntry = (blog, content, lang, showButton = false) => {
 
   <div class="blog-entry">
     ${headline}
-    <div class="sub-title"><span class="text">${getAuthors(blog.author)} on ${blog.dateFormatted}</span></div>
+    <div class="sub-title"><span class="text">${getAuthors(blog.author)}, ${blog.dateFormatted}</span></div>
     ${content}
     ${button}
   </div>


dsarkar:
@Ein-Tim ok, thanks. Let's have a look another day, how to solve this properly.

dsarkar (approved):

PR #1363, opened 2021-06-14
fix typos

thomasaugsten:


MikeMcC399:
@thomasaugsten 
Unfortunately, again, no chance to comment before you merged.

"digitial"

is misspelled. It should be "digital".

PR #1400, opened 2021-06-18
FAQs on data

dsarkar:


dsarkar:
EN will follow soon

MikeMcC399 (changes requested):
Google writes "Deinstallieren" without a "-" 
The same goes for Duden too https://www.duden.de/rechtschreibung/deinstallieren.

so "De-installation" becomes "Deinstallation" and
"de-installieren" becomes "deinstallieren".

dsarkar:
@MikeMcC399 Thanks, will fix.

Internal Tracking ID: [EXPOSUREAPP-8001](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8001)

Issue #1439, opened 2021-06-29
Selection for 2.3 screenshots does not work


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/screenshots/

## Describe the issue

When selecting 2.3 in the Versions drop-down on https://www.coronawarn.app/en/screenshots/ a 404 error is displayed. The screenshots for 2.3 are not displayed.

View code shows the following:

````
                  <div class="d-flex justify-content-sm-end position-relative mb-3 mb-sm-0">
                      <select name="archived-screenshots" class="form-select form-select-sm w-auto position-sm-absolute">
                              <option value="">Versions</option>
                                  <option value="/en/screenshots-archive/2-2.html">2.2</option>
                                  <option value="/en/screenshots-archive/2-3">2.3</option>
                                  <option value="/en/screenshots">2.4</option>
                      </select>
                  </div>
````

Follow on from open issue https://github.com/corona-warn-app/cwa-website/issues/1402.

## Workaround

To view 2.3 screenshots open the page explicitly by hand
- https://www.coronawarn.app/en/screenshots-archive/2-3.html (EN)
- https://www.coronawarn.app/de/screenshots-archive/2-3.html (DE)

## Additional information

This problem is not reproducible locally using `npm start`.
Perhaps the Open Source Team has a test server which mirrors the configuration on https://www.coronawarn.app/ to investigate and fix the problem?

MikeMcC399:
PR #1440 successfully resolves this issue.

Issue #1402 regarding 404 errors when switching languages is still open.

dsarkar:
@MikeMcC399 Thanks for PR

PR #1440, opened 2021-06-29
Add .html extension to 2.3 screenshots link

MikeMcC399:
This PR attempts to resolve the issue reported in https://github.com/corona-warn-app/cwa-website/issues/1439 where screenshots for version 2.3 cannot be selected via the drop-down version menu.

The selection of version 2.2 works correctly, so I added the .html extension to the 2.3 link in the hope that this would make 2.3 selectable too.

In a local installation tested with `npm start` the selection of version 2.3 screenshots works before and after the change.

I cannot test if this will work properly on the production webserver, because the issue was **not** reproducible on a local webserver installation anyway. If you have a test server instance, you could try it out there, otherwise I suggest to merge it into production and check it there. If it breaks anything, then the change would have to be reverted. (I think it is unlikely to have any harmful side-effects though.)

```
        "archive-dropdown": [
            {
                "version": "2.2",
                "link": "/screenshots-archive/2-2.html"
            },
            {
                "version": "2.3",
                "link": "/screenshots-archive/2-3.html"
            },
            {
                "version": "2.4",
                "link": "/screenshots"
            }
        ]
```


dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks.

MikeMcC399:
Verified in production. It worrks!

PR #1454, opened 2021-07-02
Feature/glossary

svengabr:
added new glossary to website


src/data/faq_de.json
```diff
@@ -1996,5 +1996,141 @@
                 ]
             }
         ]
-    }
+    },
+    "glossary": {
+        "glossaryHeadline": "Glossariung",
```
MikeMcC399, line 2001: Or maybe just "Glossar" ?

Ein-Tim: Yes, please! 😄

svengabr: Sorry the content is still in redaction, this is just dummy content ;)

MikeMcC399: Thanks for changing to "Draft" status. That makes it clearer now.

PR #1473, opened 2021-07-08
Fix partner names in English blog cwa-virus-variants

MikeMcC399:
This PR removes the quotes surrounding the name of The Fraunhofer Institute for Integrated Circuits IIS in 
https://www.coronawarn.app/en/blog/2021-07-08-cwa-virus-variants/ to make it consistent with how the other partner names are written. 

Also added "the" before the names of the Robert Koch Institute and the Deutsche Telekom to be consistent with the Fraunhofer Institute for Integrated Circuits IIS (but not before SAP, where the word "the" does not fit grammatically).


heinezen:
Thanks!

PR #1474, opened 2021-07-09
Remove "Recommendation" from FAQ text

Ein-Tim:
This PR removes the word "Recommendation" (DE: "Handlungsempfehlung") from the FAQ entry, "vac_cert_red_warning".

It should be clear that this is a Handlungsempfehlung and we don't use the word "Recommendation"/"Handlungsempfehlung" [here](https://www.coronawarn.app/en/faq/#rat_red_card_which_test), which is also a "Recommendation".


MikeMcC399:
@Ein-Tim 
To me, "Recommendation" in https://www.coronawarn.app/en/faq/#vac_cert_red_warning says this is not a hard rule, just a guideline. I would leave it in, even if https://www.coronawarn.app/en/faq/#rat_red_card_which_test is not worded this way.

svengabr (approved):

Ein-Tim:
@MikeMcC399 

Thank you for your feedback, but I think it just sounds nicer without the "Handlungsempfehlung". I'll leave the decision to the Open-Source-Team here.

MikeMcC399:
@Ein-Tim 
Certainly it's a matter of preference, so there is no right or wrong answer.

I've captured the existing text to assist other reviewers.

image

Your PR removes "Recommendation:" (and similarly for the German entry).

MikeMcC399:
BTW: This is a hot topic in UK!
https://www.independent.co.uk/news/health/covid-self-isolation-nhs-app-b1881315.html

dsarkar:
@MikeMcC399 @Ein-Tim Thanks. I think we should keep the word "Recommendation". @GisoSchroederSAP ?

dsarkar:
@Ein-Tim @MikeMcC399 We discussed this internally: This communication has been agreed internally with stakeholders. We might soften the English wording to "Guidance", however, I suggest closing this issue for now. I am pretty sure that we might adapt this text later on during the pandemic.

Ein-Tim:
@dsarkar @MikeMcC399 

Agreed, you won! (-;

PR #1478, opened 2021-07-09
Fix typos, etc. in glossary (Closes #1477)

Ein-Tim:
This PR fixes some issues with the text of the glossary.
You can find more informations in #1477.

---

**Closes #1477**

PR #1483, opened 2021-07-11
Improve Glossary

Ein-Tim:
This PR addresses the false statements mentioned in https://github.com/corona-warn-app/cwa-website/issues/1477#issuecomment-877214152.

This is a follow up to PR #1478.


dsarkar (approved):

Ein-Tim:
Please don't merge this yet, I have one more thing to address.

MikeMcC399:
@Ein-Tim 
> Please don't merge this yet, I have one more thing to address.

You could set the PR to Draft if you don't want it reviewed or merged yet.

Ein-Tim:
I've now addressed one more inaccuracy and this is PR is ready to merge @dsarkar!

src/data/faq_de.json
```diff
@@ -2030,11 +2030,11 @@
             "G": [
                 {
                     "term": "Genesenen-Zertifikat",
-                    "description": "Das Genesenen-Zertifikat ist ein => digitales Zertifikat, das einer Person ausgestellt wird, die nachweislich eine COVID-19-Erkrankung überstanden und die Immunität durch einen (PCR-) Test nachgewiesen hat."
+                    "description": "Das Genesenen-Zertifikat ist ein => digitales Zertifikat, das einer Person ausgestellt wird, die nachweislich eine COVID-19-Erkrankung überstanden und die Immunität durch einen PCR-Test nachgewiesen hat."
                 },
                 {
                     "term": "Gültigkeit eines Zertifikats",
```
MikeMcC399, line 2036:
```suggestion
                    "term": "Gültigkeit eines Impfzertifikats",
```
The description is only exact for a vaccination certificate, not for a test certificate, nor for a recovery certificate.

Sorry that I didn't notice that before!

If somebody with more knowledge could provide more information, then maybe the entry could be extended to cover the other two types of certificate.

MikeMcC399 (changes requested):
Change title regarding validity (Gültigkeit).

src/data/faq_de.json
Ein-Tim: Thank you & no problem that you didn't notice it before! Done in https://github.com/corona-warn-app/cwa-website/pull/1483/commits/909d4a69b632cd585b30be3b9677732d324d79ff.

GisoSchroederSAP (commented):
We cannot strictly specify "Gültigkeit eines Impfzertifikats", sorry. The validity really depends on the type of the certificate (vaccination certificate, test certificate, and recovery certificate). The article uses the vaccination certificate as an example only. But we need to enhance this article by the different validity periods for the different types. This is already in preparation.

Ein-Tim:
Okay, thanks for the feedback @GisoSchroederSAP! I'll revert the change for now, but please don't forget to adjust this, especially if every certificate has a different validity.

Ein-Tim:
So, reverted. I can adjust the wording tomorrow (if needed). But tbh, the entry really is all about vaccination certificates, so why can't we leave "Impfzertifikat" there for now and only change it ones the entry was made more general? Please think about it again @GisoSchroederSAP.

Thank you for your work, have a goos night!

GisoSchroederSAP:
Basically, the intention of the whole article is to explain the proper separation of the "technical" validity from the "effective" validity.
And yes, at the time of writing this, I only had a sample of an vaccination certificate at hand. 
In fact, the whole explanation could have been done as well with a travel passport and trans "translated" to certificates in general - but my concern is, most of the users are just focussed on the vaccination certificate.
As mentioned earlier: I am totally fine to name the article "Gültigkeit eines COVID-Zertifikats" 'cause _COVID-Zertifikat_ covers all three types of certificates.

Ein-Tim:
> As mentioned earlier: I am totally fine to name the article "Gültigkeit eines COVID-Zertifikats" 'cause COVID-Zertifikat covers all three types of certificates.

Yes, but then the title covers all three types but not the entry, so this is (in my opinion) not really helpful for the user. 

But I'm fine with that, I'll change the wording to "Gültigkeit eines COVID-Zertifikats" and I'm sure you'll come up with a more general explanation then.

Thanks for your feedback.

Ein-Tim:
Title changed, ready for merge (from my POV)!

MikeMcC399:
1) I suggest to change the title to
"Gültigkeit eines EU digitalen COVID-Zertifikats"
to match the official name on the document.

2) Before the example about the Impfzertifikat starts (before the text "In dem Zertifikat"), then add a sub-header in bold which says:
**Impfzertifikat** 
otherwise the reference "in diesem Falle" has no meaning.

3) I am confused about the paragraph starting "Die Check App". Why is that here in a glossary about the Corona-Warn-App? Shouldn't the reference in this and the following paragraph be to the Corona-Warn-App and not to the Check App? In any case there is no app with this exact name. If the CovPassCheck-App is meant, then the proper name of this app should be used (see [Google Play Store description](https://play.google.com/store/apps/details?id=de.rki.covpass.checkapp)).

4) The logic in the last sentence "Unerheblich davon..." is not so good and the sentence could be made simpler and more understandable.

Ein-Tim:
> 1. I suggest to change the title to
"Gültigkeit eines EU digitalen COVID-Zertifikats"
to match the official name on the certificate documents.

Done in https://github.com/corona-warn-app/cwa-website/pull/1483/commits/a8374a566f8f312b778607c546367aef81863dbd.

> 2. Before the example about the Impfzertifikat starts (before the text "In dem Zertifikat"), then add a sub-header in bold which says:
Impfzertifikat
otherwise the reference "in diesem Falle" has no meaning.

Done in https://github.com/corona-warn-app/cwa-website/pull/1483/commits/a8374a566f8f312b778607c546367aef81863dbd too, but now I see the problem that the second bullet point is a bit out of context? 

Bildschirmfoto 2021-07-13 um 00 03 36

> 5. Concerning 3) and 4) I don't think that a Glossary should refer to an app at all. I suggest replacing the final two paragraphs with something like:
"Ein Zertifikat behält seine gesamte Gültigkeit bis entweder die technische Gültigkeit oder die fachliche Gültigkeit abläuft."

Also done in https://github.com/corona-warn-app/cwa-website/pull/1483/commits/a8374a566f8f312b778607c546367aef81863dbd.

Ein-Tim:
@GisoSchroederSAP Please take another look if these changes are fine with you. Thanks!

MikeMcC399:
@Ein-Tim 
Thanks for making the changes! I think the article is now much more understandable than it was before.

If @GisoSchroederSAP can soon add the information about the other types of certificate that would be perfect.

Ein-Tim:
Since Giso (not mentioning him here intentionally) [said](https://github.com/corona-warn-app/cwa-documentation/issues/645#issuecomment-880119469) he will take a (well deserved) summer break, I guess we won't here feedback from him in the next few weeks. 

So maybe @dsarkar, could you take this over and get an approval on the changes? If not, please apply the `on hold` label here and we'll wait for Giso to return. 

Thanks everyone enjoy the evening and sleep well! 😴

src/data/faq_de.json
```diff
@@ -2047,11 +2047,11 @@
...
+                    "term": "Gültigkeit eines EU digitalen COVID-Zertifikats",
+                    "description": "Wir unterscheiden grundsätzlich<ul><li>die Gültigkeit des digitalen Zertifikats (die technische Gültigkeit) und</li><li>die Gültigkeit des Impfschutzes (die fachliche Gültigkeit).</li></ul>Das digitale Zertifikat hat ein Ausstellungsdatum und ein technisches Ablaufdatum (so wie auch Ihr Reisepass und PA). Die Gültigkeit des Zertifikats endet 1 Jahr nach Ausstellung. <br/><br/> <b> Impfzertifikat </b> <br/><br/> In dem Zertifikat werden Daten transportiert, in diesem Falle Informationen zur Impfung. Diese Impfung fand an einem Datum X statt und gilt ab dann (nach aktuellem Stand der wissenschaftlichen Erkenntnisse) z.&nbsp;B. in Deutschland 1 Jahr. Danach ist der Impfschutz erloschen. Die Dauer des Impfschutzes bezeichnen wir als 'fachliche' Gültigkeit. Die fachliche Gültigkeit Impfung zählt also ab dem letzten Impfdatum (Datum X). <br/><br/> Ein Zertifikat behält seine gesamte Gültigkeit bis entweder die technische Gültigkeit oder die fachliche Gültigkeit abläuft."
```
GisoSchroederSAP, line 2054: "Ein Zertifikat behält seine Gültigkeit, bis..." - das ginge auch kürzer: "Ein Zertifikat ist solange valide, bis..."

GisoSchroederSAP (approved):
From my perspective, all changes are okay to be accepted; the first commented proposal could be even more polished, but I don't insist here. 
Thanks for having an eye on our Glossary.

src/data/faq_de.json
Ein-Tim: Thanks for your feedback! Done in https://github.com/corona-warn-app/cwa-website/pull/1483/commits/5ed5439c84b3c596f1f708f88e3b793609b5036a.

dsarkar (approved):

dsarkar:
@Ein-Tim @MikeMcC399 @GisoSchroederSAP  Thanks! Merged.

Issue #1462, opened 2021-07-07
FAQ: Article missing for 2.5 app error HC_PREFIX_INVALID


MikeMcC399:
## Technical details

- Device name: Samsung Galaxy A5 (2017) SM-A520F
- Android version: 8.0.0
- App version: 2.5.1-RC1 (pre-release version)

Also affects iOS.

## Describe the bug

When a scan error occurs scanning an EU digital COVID certificate, the error message HC_PREFIX_INVALID includes an option "FAQ FOR CERTIFICATES" which links to an FAQ article https://www.coronawarn.app/en/faq/#hc_prefix_invalid.

The articles

- https://www.coronawarn.app/en/faq/#hc_prefix_invalid and
- https://www.coronawarn.app/de/faq/#hc_prefix_invalid

do not exist.

## Steps to reproduce the issue

1. Tap "Certificates" in the bottom menu bar
2. Tap "ADD CERTIFICATE"
3. Scan any QR code which is not a valid EU digital COVID certificate such as the following QR code which contains the text "DE12345678"

![QR Code DE12345678](https://user-images.githubusercontent.com/66998419/105467552-25905d80-5c96-11eb-8408-b4835ee6d61f.png)

![image](https://user-images.githubusercontent.com/66998419/124722041-f1d68680-df09-11eb-8d65-01beeffd468e.png)

## Expected behavior

Tapping on "FAQ FOR CERTIFICATES" should display a relevant FAQ article.

## Possible Fix

Write and publish articles which display under the following URLs:

- https://www.coronawarn.app/en/faq/#hc_prefix_invalid and
- https://www.coronawarn.app/de/faq/#hc_prefix_invalid

---

This is a follow-on from https://github.com/corona-warn-app/cwa-documentation/issues/653.

Internal Tracking ID: [EXPOSUREAPP-8216](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8216) iOS
Internal Tracking ID: [EXPOSUREAPP-8226](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8226) Android

MikeMcC399:
@dsarkar 
CWA 2.5.1 was released today (https://www.coronawarn.app/en/blog/2021-07-12-cwa-version-2-5/) and the FAQ article for the app error is still missing.

Ein-Tim:
PR #1494 added the entry.

MikeMcC399:
PR #1495 submitted.

[...]

PR #1495, opened 2021-07-12
Make hc_prefix_invalid bookmarks work

MikeMcC399:
This PR fixes two issues in
https://www.coronawarn.app/en/faq/#hc_prefix_invalid and
https://www.coronawarn.app/de/faq/#hc_prefix_invalid
(see https://github.com/corona-warn-app/cwa-website/issues/1462#issuecomment-878303906).

1) In the text of https://www.coronawarn.app/en/faq/#hc_prefix_invalid it says:
"Why do I get the error message 'This QR code is not a valid certificate. (HC_PREFIX_INVALID)' when I try to scan my vaccination QR code?"

The word "vaccination" is removed, since the error can refer to vaccination, test or recovery certificates.

2) The URL https://www.coronawarn.app/en/faq/#hc_prefix_invalid does not work correctly, although it does manage to show the article by falling back to search.
The URL https://www.coronawarn.app/de/faq/#hc_prefix_invalid does not work correctly. Instead it shows the Glossar. Due to the search fallback, if the user scrolls down from the Glossar the article can be viewed. 

The trailing space in the anchor is removed for both en and de sources.

`"anchor": "hc_prefix_invalid ",` is changed to
`"anchor": "hc_prefix_invalid",`

which allows the article to be directly displayed via bookmark selection instead of the imprecise way through search.

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -126,8 +126,8 @@
                 "id": "eu_dcc",
                 "accordion": [
                     {
-                        "title": "Why do I get the error message 'This QR code is not a valid  certificate. (HC_PREFIX_INVALID)' when I try to scan my vaccination QR code?",
-                        "anchor": "hc_prefix_invalid ",
+                        "title": "Why do I get the error message 'This QR code is not a valid  certificate. (HC_PREFIX_INVALID)' when I try to scan my QR code?",
+                        "anchor": "hc_prefix_invalid",
                         "active": true,
                         "textblock": [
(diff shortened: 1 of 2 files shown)


dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks a lot !!!!

MikeMcC399:
Verified in production that
https://www.coronawarn.app/en/faq/#hc_prefix_invalid and
https://www.coronawarn.app/de/faq/#hc_prefix_invalid
now open correctly!

PR #1514, opened 2021-07-19
FAQ: Make hc_prefix_invalid title impersonal

MikeMcC399:
This PR changes the title of

https://www.coronawarn.app/en/faq/#hc_prefix_invalid and
https://www.coronawarn.app/de/faq/#hc_prefix_invalid

to refer to "a QR code" instead of "my QR code" / "einen QR-Code" instead of "meinen QR-Code", since the QR code can be from any person. The app is not restricted to only scanning QR codes from one person any more.

In the English text I also tidied up some whitespace placement.

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -127,11 +127,11 @@
                 "id": "eu_dcc",
                 "accordion": [
                     {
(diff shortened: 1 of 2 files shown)


dsarkar (approved):

dsarkar:
@MikeMcC399 Thank you!

PR #1536, opened 2021-07-26
Correct typo in FAQ eu_dcc_validity

MikeMcC399:
This PR corrects a typo in https://www.coronawarn.app/en/faq/#eu_dcc_validity

`holders's` was incorrect.

Singular possessive is `holder's`
(Plural possessive would be` holders'`, however this does not fit the context.)


heinezen:
Thanks!

The correct terminology would actually be *subject* instead of *holder* but I think we should discuss that in a separate issue. For now, i'll just merge your corrections.

----

MikeMcC399:
@heinezen 
In German, the phrase is:
'Datum des ersten positiven PCR-Test-Ergebnisses der Person' 

which could be written in German as:
'date of the person's first positive PCR test result'

I couldn't find "holder" or "subject" in any document of the EU Health Network.

I don't have any strong feelings on which term to use ("holder" / "subject" / "person"). I was just trying to correct the grammar. 🙂 If you think it should be changed, then please feel free to go ahead.

Issue #1550, opened 2021-07-28
Issues with the screenshots page


Ein-Tim:
- [ ] Please finally remove the contact journal onboarding screens under iOS (align it to the Android section).
- [x] "Check-In - Scan QR-Code" section: 
	- [x] Title is not translated into German (should be "Check-in - QR-Code scannen"
- [x] "Check-in - QR-Code erstellen" / "Check In - Create Event section:
	- [x] Titles should be aligned (German to English or vice versa)
	- [x] Picture 3 is still broken
- [x] "Aktuelle Zahlen" / "Statistics" section:
	- [x] First picture shows a screenshot of the "Länderübergreifende Risiko-Ermittlung" -> should be replaced
- [x] "Test registrieren & andere warnen" section:
	- [x] Picture 2 is stil broken
- [x] "Testergebnis abrufen" / "Test result home"
	- [x] English title should be aligned to the German one
- [x] "Zertifikate" / "Certificates" section:
	- [x] All pictures have the same name ("Test certificate Image")
- [x] "Testzertifikat" / "Test certificate" section: 
	- [x] All pictures have the same name ("Test certificate Image")
	- [x] Second picture is broken
- [x] "Impfzertifikat" / "Vaccination Certificate" section:
	- [x] All pictures have the same name ("Vaccination certificate Image 11 of 11")
- [x] "Genesenenzertifikat" section:
	- [x] All pictures have the same name ("Genesenenzertifikat")
	
from https://github.com/corona-warn-app/cwa-website/pull/1522#pullrequestreview-712689255

MikeMcC399:
Also https://github.com/corona-warn-app/cwa-website/issues/1549.

Ein-Tim:
Everything mentioned here has been fixed through #1552. 

The contact journal onboarding is tracked in #1307. 

Closing as fixed.

PR #1552, opened 2021-07-28
Fix screenshots with date picker and keyboard 

dsarkar:
Fix iOS screenshot issues (see #1550)


src/data/screenshots.json
```diff
@@ -268,7 +268,7 @@
                 "anchor": "ios_registertest",
                 "images": [
                     { "filename": "/assets/screenshots/2-6/en/ios/iPhone 11 Pro-tan_submissionflow_qr_0002", "alt": "Test registration & Key sharing image 1 of 11" },
-                    { "filename": "/assets/screenshots/2-6/en/ios/iPhone 11 Pro-tan_submissionflow_tan_0003", "alt": "Test registration & Key sharing image 2 of 11" },
+                    { "filename": "/assets/screenshots/2-5/en/ios/iPhone 11 Pro-tan_submissionflow_tan_0003", "alt": "Test registration & Key sharing image 2 of 11" },
```
Marcono1234, line 271: What is the reason for this change? I don't see any obvious difference between the screenshots:
|2.5|2.6|2.7|
|-|-|-|
||||

Though for the German screenshots (referenced below) the keyboard is indeed broken:
|2.5|2.6|2.7|
|-|-|-|
||||

However, wouldn't it make sense then to delete the unused screenshots to avoid confusion?

PR #1574, opened 2021-08-02
fix: check if screenshot-tabs exists before updating from hash

mojoaxel:
### Problem 
In the current version of the app (87571d529b21e7d578901abd143de2d1874f13e3) there is an error on the console:

image

This error is caused by the `checkHashAndChangeTab` function that expects the screenshot tabs to be present.

This results in a lower lighthouse score:

image

### Solution

Check if the screenshot tabs exist on the page before executing the `checkHashAndChangeTab` function.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -275,6 +275,10 @@ $(document).ready(function(){
             scrollTo(hash)
         }
     }
-    checkHashAndChangeTab()
+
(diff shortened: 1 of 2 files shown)


dsarkar:
@mojoaxel Thanks. We will review this.

dsarkar:
@mojoaxel Thanks again for contributing. This has been fixed in the meantime. Closing this issue

Issue #1168, opened 2021-04-27
eventregistration.js triggers build warning regarding size limit


MikeMcC399:
## Where to find the issue

https://github.com/corona-warn-app/cwa-website/blob/master/src/assets/js/eventregistration.js

## Describe the issue

When the website is built through https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml a warning is shown regarding `eventregistration.js`.

```
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  eventregistration.js (658 KiB)

```

For example the pull request https://github.com/corona-warn-app/cwa-website/pull/1167 triggered the action which is shown in the logs under https://github.com/corona-warn-app/cwa-website/actions/runs/788829128. Here is the relevant part of the log:

```
Built at: 04/27/2021 11:00:24 AM
               Asset     Size  Chunks                    Chunk Names
              app.js  139 KiB       0  [emitted]         app
eventregistration.js  658 KiB       1  [emitted]  [big]  eventregistration
Entrypoint app = app.js
Entrypoint eventregistration [big] = eventregistration.js

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  eventregistration.js (658 KiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  eventregistration (658 KiB)
      eventregistration.js


WARNING in webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

## Suggested change

Please consider if something should be done to avoid this warning appearing.

---
Internal Tracking ID: [EXPOSUREAPP-13404](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13404)

MikeMcC399:
It seems that #1575 may partially address this issue.

MikeMcC399:
PR #1772 causes the warning to no longer appear

```
[17:42:45] Version: webpack 4.46.0
Built at: 29/09/2021 17:42:45
               Asset      Size  Chunks                    Chunk Names
              app.js   140 KiB       0  [emitted]         app
eventregistration.js   679 KiB       1  [emitted]  [big]  eventregistration
[...]

PR #1575, opened 2021-08-02
refactor: remove moment.js from eventregistration bundle

mojoaxel:
The PR removes `moment` from the `eventregistration` bundle (It is still used in the blog-preprocessing).

### Why?

* moment is BIG. Removing it reduces the distribution bundle size from 679 kB to 318 kB!
* moment is [deprecated](https://momentjs.com/docs/#/-project-status/).
* moment is not needed:
  * testing if a string is a valid date can easily be done in javascript.
  * getting a unix timestamp from a date is also basic javascript.
* moment was not mentioned in the THIRD-PARTY-NOTICES.
* moment was only installed as devDependency and should not used in dist bundles.


MikeMcC399:
@mojoaxel 
It looks like this PR partially addresses https://github.com/corona-warn-app/cwa-website/issues/1168.
Thank you!

dsarkar:
@mojoaxel Thanks. We will review this.

mojoaxel:
If that helps here are some quick "test" I did to ensure my functions did what they are suppose to do.

```
> dateTimeToUnixTimestamp("2021-01-01", "12:00")
1609498800
> moment("2021-01-01 12:00").format("X")
"1609498800"
```

```
> isValidDate("2021-01-01")
true
> moment("2021-01-01").isValid()
true
```

```
> isValidDate("2021-01-40")
false
> moment("2021-01-40").isValid()
false
```

svengabr:
> 
> 
> @mojoaxel
> It looks like this PR partially addresses #1168.
> Thank you!

Partially addresses yes, but unfortunately it does not fix it.

I did a test locally and I cant find any bugs in this pull request so I will approve this pull request.

Thanks for your contribution @mojoaxel

svengabr (approved):

Ein-Tim:
FYI: #1706 reverts this PR as it caused a quite significant error: #1703

PR #1615, opened 2021-08-12
Add FAQ article hc_signature_invalid 

larswmh:
As a response to [this issue](https://github.com/corona-warn-app/cwa-website/issues/1594), here's the pull request to add the discussed FAQ article.

German:
chrome_6kTJLme6SU

English:
chrome_so9OPcEDZt

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -173,6 +173,18 @@
                             "Using this step-by-step guide, you can remove certificates from the Corona-Warn-App. Please note that the first three steps apply to both iOS and Android, but only one of the last two steps needs to be followed, depending on which operating system matches yours. Vaccination certificates, test certificates and recovery certificates are mentioned as 'Certificates', since the process of removing them is identical.",
                             "<ul><li>Tap on the 'Certificates' tab</li><li>Tap on the certificate you would like to remove</li><li>Scroll down to and tap on the tile that matches the certificate you want to delete.</li><li>iOS: Tap on 'Delete Certificate' at the bottom of the screen</li><li>Android: Tap on the three-dot symbol on the top right of the screen, and then tap on 'Remove'</li></ul>"
                         ]
+                    },
+                    {
+                        "title": "Why do I get the error message that my certificate is invalid?",
+                        "anchor": "hc_signature_invalid",
+                        "active": true,
+                        "textblock": [
+                            "The signature of your certificate is invalid.",
+                            "This invalidates the entire certificate as a digital document. Therefore, it cannot be used as a proof and also cannot be imported into the app. If this error message occurs with a certificate that has already been imported, it is now marked as invalid.",
+                            "Please check if the validity of your certificate has expired:",
+                            "<ul><li>A vaccination certificate issued in Germany is usually valid for one year from the date of issue</li><li>A recovery certificate on the other hand may not be valid for longer than 180 days from the date of the last positive PCR test, as defined by the EU.</li></ul>",
+                            "If your certificate is technically still valid, it was obviously issued by an unauthorized body or incorrectly. In this case, we ask you to contact the <a href='#international_phone_numbers' target='_blank' rel='noopener noreferrer'>technical hotline</a> and tell us when and by which body this certificate was issued. Then please have a new certificate with a valid signature issued by an authorized body."
+                        ]
                     }
                 ]
             },
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -180,7 +180,7 @@
                         "active": true,
                         "textblock": [
                             "The signature of your certificate is invalid.",
-                            "This invalidates the entire certificate as a digital document. Therefore, it cannot be used as a proof and also cannot be imported into the app. If this error message occurs with a certificate that has already been imported, it is now marked as invalid.",
+                            "This invalidates the entire certificate as a digital document. Therefore, it cannot be used as proof of health status and also cannot be imported into the app. If this error message occurs with a certificate that has already been imported, it is now marked as invalid.",
                             "Please check if the validity of your certificate has expired:",
                             "<ul><li>A vaccination certificate issued in Germany is usually valid for one year from the date of issue</li><li>A recovery certificate on the other hand may not be valid for longer than 180 days from the date of the last positive PCR test, as defined by the EU.</li></ul>",
                             "If your certificate is technically still valid, it appears it was issued by an unauthorized body or incorrectly. In this case, we ask you to contact the <a href='#international_phone_numbers' target='_blank' rel='noopener noreferrer'>technical hotline</a> and tell us when and by which body this certificate was issued. Then please have a new certificate with a valid signature issued by an authorized body."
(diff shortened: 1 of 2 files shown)


dsarkar (approved):

Ein-Tim:
I'd suggest to add the `do not merge` label here until version 2.7 has been released to avoid an accidental merge 😅

MikeMcC399:
@Ein-Tim 
> I'd suggest to add the do not merge label here until version 2.7 has been released to avoid an accidental merge 😅

I don't see a problem with publishing the article before version 2.7 is released, because there is nothing which would be wrong for earlier versions. The error message would just not be displayed in earlier versions.

Ein-Tim:
Agreed @MikeMcC399 😅

src/data/faq_de.json
```diff
@@ -173,6 +173,18 @@
...
+                    },
+                    {
+                        "title": "Warum erhalte ich die Fehlermeldung, dass mein Zertifikat ungültig ist?",
+                        "anchor": "hc_signature_invalid",
+                        "active": true,
+                        "textblock": [
+                            "Die Signatur des Zertifikats ist ungültig.",
+                            "Damit ist das ganze Zertifikat als digitales Dokument ungültig. Es kann daher nicht als Nachweis verwendet und auch nicht in die App importiert werden. Tritt diese Fehlermeldung bei einem bereits importierten Zertifikat auf, wird es nun als ungültig markiert.",
+                            "Prüfen Sie bitte, ob die Gültigkeit Ihres Zertifikats abgelaufen ist:",
+                            "<ul><li>In der Regel ist ein in Deutschland ausgestelltes Impfzertifikat ein Jahr ab Ausstellungdatum gültig.</li><li>Ein Genesenen-Zertifikat darf laut Festlegung der EU dagegen nicht länger als 180 Tage ab dem Datum des letzten positiven PCR-Tests gültig sein.</li></ul>",
```
MikeMcC399, line 185:
```suggestion
                            "<ul><li>In der Regel ist ein in Deutschland ausgestelltes Impfzertifikat ein Jahr ab Ausstellungdatum gültig.</li><li>Ein Genesenenzertifikat darf laut Festlegung der EU dagegen nicht länger als 180 Tage ab dem Datum des letzten positiven PCR-Tests gültig sein.</li></ul>",
```
Spelling of "Genesenenzertifikat" as used in the app and elsewhere in the FAQs.

larswmh: Changed it, thanks!

src/data/faq.json
```diff
@@ -173,6 +173,18 @@
...
+                            "This invalidates the entire certificate as a digital document. Therefore, it cannot be used as a proof and also cannot be imported into the app. If this error message occurs with a certificate that has already been imported, it is now marked as invalid.",
+                            "Please check if the validity of your certificate has expired:",
+                            "<ul><li>A vaccination certificate issued in Germany is usually valid for one year from the date of issue</li><li>A recovery certificate on the other hand may not be valid for longer than 180 days from the date of the last positive PCR test, as defined by the EU.</li></ul>",
+                            "If your certificate is technically still valid, it was obviously issued by an unauthorized body or incorrectly. In this case, we ask you to contact the <a href='#international_phone_numbers' target='_blank' rel='noopener noreferrer'>technical hotline</a> and tell us when and by which body this certificate was issued. Then please have a new certificate with a valid signature issued by an authorized body."
```
MikeMcC399, line 186:
```suggestion
                            "If your certificate is technically still valid, it appears it was issued by an unauthorized body or incorrectly. In this case, we ask you to contact the <a href='#international_phone_numbers' target='_blank' rel='noopener noreferrer'>technical hotline</a> and tell us when and by which body this certificate was issued. Then please have a new certificate with a valid signature issued by an authorized body."
```
If you say "obviously", it implies that the user must be an idiot not to have noticed it themselves previously.
It is kinder to the user to say "it appears it was issued ..."

larswmh: Changed it, thanks!

src/data/faq.json
```diff
@@ -173,6 +173,18 @@
...
                             "<ul><li>Tap on the 'Certificates' tab</li><li>Tap on the certificate you would like to remove</li><li>Scroll down to and tap on the tile that matches the certificate you want to delete.</li><li>iOS: Tap on 'Delete Certificate' at the bottom of the screen</li><li>Android: Tap on the three-dot symbol on the top right of the screen, and then tap on 'Remove'</li></ul>"
                         ]
+                    },
+                    {
+                        "title": "Why do I get the error message that my certificate is invalid?",
+                        "anchor": "hc_signature_invalid",
+                        "active": true,
+                        "textblock": [
+                            "The signature of your certificate is invalid.",
+                            "This invalidates the entire certificate as a digital document. Therefore, it cannot be used as a proof and also cannot be imported into the app. If this error message occurs with a certificate that has already been imported, it is now marked as invalid.",
```
MikeMcC399, line 183:
```suggestion
                            "This invalidates the entire certificate as a digital document. Therefore, it cannot be used as proof of health status and also cannot be imported into the app. If this error message occurs with a certificate that has already been imported, it is now marked as invalid.",
```

In this context "proof" is an uncountable noun, so "a proof" is not correct - see https://www.macmillandictionary.com/dictionary/british/proof_1.  I suggest instead "proof of health status".

The EU page https://ec.europa.eu/info/live-work-travel-eu/coronavirus-response/safe-covid-19-vaccines-europeans/eu-digital-covid-certificate_en disagrees on this point where it writes "An EU Digital COVID Certificate is a digital proof that a person has either been vaccinated against COVID-19, received a negative test result or recovered from COVID-19.", I would still favor my suggestion above though.

MikeMcC399 (changes requested):
I made some word change suggestions.

MikeMcC399:
I thought it would be good also to include the error codes which lead to the FAQ article, however it seems in Android at least that there are many different codes:
```
        ErrorCode.HC_COSE_NO_SIGN1,
        ErrorCode.HC_COSE_PH_INVALID,
        ErrorCode.HC_COSE_NO_ALG,
        ErrorCode.HC_COSE_UNKNOWN_ALG,
        ErrorCode.HC_DSC_NO_MATCH,
        ErrorCode.HC_DSC_OID_MISMATCH_TC,
        ErrorCode.HC_DSC_OID_MISMATCH_VC,
        ErrorCode.HC_DSC_OID_MISMATCH_RC,
        ErrorCode.HC_DSC_NOT_YET_VALID,
        ErrorCode.HC_DSC_EXPIRED,
```

It would blow up the size of the article if each of these codes were to be explained. Hopefully the codes will not be triggered for most users, although many of the published test codes for developers are unsigned, which would lead to `HC_DSC_NO_MATCH`.

So this is just a remark without asking for any change in the article in this respect.

GisoSchroederSAP:
Good point @MikeMcC399., thanks for mentioning. If there are that many error codes that lead to the same or similar reason, we may have to create a more "abstract" description/explanation. 
On the other hand, as of today, our FAQ points to very specific but frequently occurring error codes.  Maybe, we give it some time and observe, if there are patterns that we comment later on. I just want to avoid "overloading" the FAQ reading consumers 
with technical codes that maybe never or just rarely appear.

src/data/faq.json
larswmh: Changed it, thanks!

dsarkar:
@larswmh @MikeMcC399 @GisoSchroederSAP @Ein-Tim Good morning! Thanks to everybody for contributing! @MikeMcC399  This https://github.com/corona-warn-app/cwa-website/pull/1615#issuecomment-897679017  is actually an interesting idea: We will see, how often this error occurs. Hopefully almost never, but if several of these errors occur more frequently it might be helpful to have the error code list in the article. Affected users will remember the error code and with this information  the dev team might find the root cause of a problem more easily.

Issue #1540, opened 2021-07-27
Glossary placement and naming


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/de/faq/ Glossar

Assuming that PR https://github.com/corona-warn-app/cwa-website/pull/1534 is merged, then also
https://www.coronawarn.app/en/faq/ Glossary

## Describe the issue

The Glossary takes a prominent place at the top of all FAQ articles.

The name and the placement implies that it is a glossary covering all aspects of CWA, however the currently published German-language terms only deal with EU Digital COVID Certificates (DCC) and nothing more. The terms in the English-language PR https://github.com/corona-warn-app/cwa-website/pull/1534 are currently a subset of the terms appearing the the German-language Glossar.

The fact that the Glossary appears at the top of the FAQ hides the fact that there is an FAQ search field available on the page. The user must then search for the search field by scrolling down to it.

## Suggested change

- Move the Glossar / Glossary to a less prominent place on the web site. For instance make it a section of the FAQs or separate it from the FAQs.
- Change the name to Zertifikatsglossar / Certificate Glossary
- Link to the [app Glossary](https://github.com/corona-warn-app/cwa-documentation/blob/master/glossary.md) in the [documentation](https://github.com/corona-warn-app/cwa-documenation) repository.

---
Internal Tracking ID: [EXPOSUREAPP-8934](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8934)

larswmh:
I'm planning to work on this. I think that putting the glossary below the entire FAQ would solve the problem, but I would still leave some kind of hyperlink on top so people can quickly get to it if that's what they are specifically looking for.

dsarkar:
@MikeMcC399 Thanks for proposal. @larswmh implemented a solution for now to mitigate the "worst". Thanks.

Ein-Tim:
@MikeMcC399 Since the glossary was moved to the bottom (PR #1617), the main part of this issue has been fixed. Would you maybe want to open a new one regarding the naming? If you want to leave this issue open that's also fine for me.

MikeMcC399:
@Ein-Tim 
The naming of the "Glossar" is an open issue, so I would leave this issue open.

Provide Glossary in English is still unaddressed https://github.com/corona-warn-app/cwa-website/issues/1476.

https://github.com/corona-warn-app/cwa-website/issues/1566 is solved, so I will make a note there.

[...]

PR #1617, opened 2021-08-13
Repositioning of the glossary

larswmh:
As a response to [this issue](https://github.com/corona-warn-app/cwa-website/issues/1540), here's the pull request to reposition the glossary. The glossary will be displayed at the bottom of the page instead of on top, but there's still a hyperlink below the search bar that will lead to the glossary, so you don't have to scroll all the way down. 

Comparison of the top of the new FAQ page and the old FAQ page:
vergleich

### DIFF
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -2094,6 +2094,7 @@
     "glossary": {
         "glossaryHeadline": "Glossar",
         "optionalText": "",
+        "linkToGlossary": "Link zum Glossar",
         "glossaryWords": {
             "B": [
                 {
--- a/src/partials/page-faq.html
+++ b/src/partials/page-faq.html
@@ -1,8 +1,63 @@
 {{#if page-contents}}
+<div class="container container_flex">
+  <aside class="side-menu js-menu" id="side-menu">
+    <div class="menu-close"><button class="btn btn-close js-toggle" data-target="#side-menu"></button></div>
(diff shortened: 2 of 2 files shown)


dsarkar (approved):

dsarkar:
@larswmh Thanks.

Ein-Tim:
@larswmh 

Thank you for this PR! Do you see a way to get some more space between the "Geklärt" text and the headline of the Glossary?
Bildschirmfoto 2021-08-13 um 11 10 37

Right now it looks a bit tight, I'd say 😅

larswmh:
@Ein-Tim thanks for the information! I'll try to find a solution for this as soon as possible. The spacing on my fullscreen window was perfectly fine when I tested it before my pull request, but now if it's a smaller size window the spacing gets tighter and eventually the sidebar is inside of the glossary

MikeMcC399:
It's actually unreadable on a mobile device in landscape mode.

larswmh:
Thanks for both of your comments. The next commit will fix the sidebar going into the glossary, reassuring readability on smaller monitors or landscape mode on mobile devices for example. As for the other problems that this brought to my attention, like the Glossary title being unaligned with FAQ section titles, or the sidebar disappearing in the footer (which seems to have been an issue before too), I can't come up with a solution right now.

Ein-Tim:
Thanks for taking care of it @larswmh!

PR #1620, opened 2021-08-13
Restore readability to recent glossary reposition

larswmh:
As a follow-up to this [pull request](https://github.com/corona-warn-app/cwa-website/pull/1617), here's the pull request to restore it's readability on smaller screens.


MikeMcC399:
@larswmh 
The problem which this PR seeks to resolve is also reproducible with Google Chrome on a desktop with an aspect ratio and size similar to a mobile device in landscape mode.

This is a display of https://www.coronawarn.app/de/faq/#glossary:

image

This PR puts the Glossary back up to the top of the page https://www.coronawarn.app/de/faq/ again where it displays correctly.

dsarkar (approved):

larswmh:
@MikeMcC399 I've tested this branch locally and it does not bring the glossary back up to the top of the page. Might not look like it in the code, but putting the glossary code into main prevented it from being interacted with by the sidebar. Would've included a gif to show it in many different aspect ratios/sizes but it's too large for github upload

MikeMcC399:
@larswmh 
It seems to work correctly now in production. 👍🏻 
As you say, with the "Glossar" down at the bottom.

PR #1628, opened 2021-08-15
R 2.7 screenshots

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-8599](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8599)


Marcono1234:
Could it be that by accident some outdated screenshots were added? For example [`iPhone 11 Pro-statistics_at_least_one_vaccination.png`](https://github.com/corona-warn-app/cwa-website/blob/master/src/assets/screenshots/2-7/en/ios/iPhone%2011%20Pro-statistics_at_least_one_vaccination.png) says "Up to Nov 8, 20**20**".
The same also applies for the corresponding 2.6 screenshot.

Though, interestingly this issue is not directly visible on the website because (for some reason) the JSON files for 2.6 and 2.7 include the screenshot of version 2.5.

dsarkar:
@Marcono1234 In this case the specific reason to use 2.5 screenshots is that the news apparently are made with a date where no statistics data was provided (at least not at the source the app takes its values) and the calculated numbers and shown in the screenshots do not make much sense.

Ein-Tim:
@Marcono1234 The data shown in the screenshots is 100% fake, it's not relevant what dates they show.

Marcono1234:
Thanks for the response. Is there any point in having the screenshot files then given that they are neither used by 2.6 nor by 2.7, or can they be deleted (to reduce some confusion)?

Is that also the reason why the content of `iPhone 11 Pro-screenshot_recovery_certificate_overview_part1` differs between versions:
|2.5|2.6|2.7|
|-|-|-|
||||

The screenshot JSONs seem to exploit this by referencing the 2.6 version as "valid certificate" and the 2.7 version as "invalid certificate".
I am slightly afraid that such things will become a maintenance nightmare in the future.

Ein-Tim:
@Marcono1234 The screenshots page is definitely something that could cause maintenance issues in the future, that's why I opened #1660 to discuss about this. Feel free to add your opinion there.

To answer your question: Yes, as you can see in the screenshots from version 2.5 the name was "ERIKA<DOERTE", the standardized name, which was not correct. This was fixed in the 2.6 screenshots and with version 2.7 this screenshot was reused for something different...
I agree with you that this is really strange.

dsarkar:
@Marcono1234  @Ein-Tim The screenshots are generated automatically, however, not always when a new feature was implemented the automatic screenshot feature is updated and/or adapted. 

For R 2.8 there are screenshots, that differentiate between "valid" and "invalid" certificates, see assets for R2.8  here: https://github.com/corona-warn-app/cwa-app-ios/releases

Issue #1653, opened 2021-08-23
Screenshots page: Android uses anchor `ios_rat`


Marcono1234:
## Where to find the issue
https://github.com/corona-warn-app/cwa-website/blob/ff40e69e20050ef9f0110c987aa0442b34e1d85d/src/data/screenshots-archive-2-3.json#L112

## Describe the issue
A section for Android screenshots uses an anchor called `ios_rat`.

:warning: This might affect more than just the version 2.3 page.

## Suggested change
- Fix the incorrect anchor.
- If possible: Add a GitHub workflow (or enhance the current workflow) verifying that all Android anchors start with `android_` and all iOS anchors start with `ios_`.

dsarkar:
@Marcono1234 Thanks. Do you want to offer a fix PR?

Marcono1234:
@dsarkar, created #1656

dsarkar:
@Marcono1234 Your Pr was merged. Thanks.

dsarkar:
fixed

PR #1656, opened 2021-08-23
Fix `ios_rat` anchors being used for Android screenshots

Marcono1234:
Resolves partially #1653

Ideally as mentioned in #1653 the GitHub workflow would be adjusted to detect such incorrect anchors.


dsarkar (approved):

dsarkar:
@Marcono1234 Thanks for the fix.

Issue #1655, opened 2021-08-23
Screenshots page: Multiple screenshots have alt text "Image/Bild"


Marcono1234:
## Where to find the issue
https://github.com/corona-warn-app/cwa-website/blob/d8ab82dd48c6142786b62f3debe3d4109e63990b/src/data/screenshots.json#L136

## Describe the issue
Multiple screenshots just have the generic `alt` text "Image/Bild". This might not be very good for accessibility.

:warning: This affects more than the linked `screenshots.json`.

## Suggested change
Not quite sure what the best way to solve this would be. This might not be that problematic because the whole page is dedicated to screenshots and therefore most likely not visited by someone with reduced vision (?).

So some possible solutions might be:
- Use a proper `alt` text
- Simply omit the pointless "Image/Bild" `alt` text since it adds little value
- Omit the pointless "Image/Bild" and for screenshots without `alt` use "Screenshot" as default; this could possibly be specified [here](https://github.com/corona-warn-app/cwa-website/blob/0b5f325b232ccd883a69e0f6ed54cb5377f7e63d/src/partials/page-screenshots.html#L75-L76)?

---

Also for what it's worth, the numeration of some of the screenshots is incorrect, e.g. (all have "1 von 3"):
https://github.com/corona-warn-app/cwa-website/blob/d8ab82dd48c6142786b62f3debe3d4109e63990b/src/data/screenshots-archive-2-6_de.json#L341-L343

Ideally this text ("n von N" / "n of N") would be automatically generated (maybe [here](https://github.com/corona-warn-app/cwa-website/blob/0b5f325b232ccd883a69e0f6ed54cb5377f7e63d/src/partials/page-screenshots.html#L75-L76)?) since this is often incorrect or outdated.

And some of the English screenshots use the German `alt` texts, e.g.:
https://github.com/corona-warn-app/cwa-website/blob/ff40e69e20050ef9f0110c987aa0442b34e1d85d/src/data/screenshots.json#L369

larswmh:
Fixed with PR #1663

Issue #1549, opened 2021-07-28
Android Screenshots: "Check certificates for validity before a trip" wrong OS


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/screenshots/#android_check_validity_certificate "Check certificates for validity before a trip"
https://www.coronawarn.app/de/screenshots/#android_check_validity_certificate "Zertifikate vor einer Reise auf Gültigkeit prüfen"

First reported for CWA version: 2.8
Reviewed as not yet fixed for CWA version: 2.23 screenshots

## Describe the issue

The Android screenshots for "Check certificates for validity before a trip" are unlike the screens actually presented on Android.

They are screenshots from iOS.

### Steps to reproduce

1. View https://www.coronawarn.app/en/screenshots/#android_check_validity_certificate
2. View https://www.coronawarn.app/de/screenshots/#android_check_validity_certificate
3. Note that the English and German language screenshots for Android are from iOS. This is for instance recognizable because "Check Validity" is shown in mixed case. On a real Android device the button label appears in upper case as "CHECK VALIDITY".

## Suggested change

Replace the screen shots by ones from Android.

---
Internal Tracking ID: [EXPOSUREAPP-8503](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8503)

MikeMcC399:
The Android section is using iOS screenshots.

https://github.com/corona-warn-app/cwa-website/blob/b9237487273a8f0d84df47007d3e0a9212706d02/src/data/screenshots.json#L171-L178

There don't appear to be appropriate screen shots available in https://github.com/corona-warn-app/cwa-website/tree/master/src/assets/screenshots/2-6/en/android.

MikeMcC399:
A correct screenshot taken from CWA Android 2.6 is

![VaccinationDetailsFragment_2 6 1](https://user-images.githubusercontent.com/66998419/129439402-9a17fdf2-223b-4b6b-9949-af32b412ba8b.png)

Compare to the posted screenshot (which is from iOS, not from Android):

![iPhone 11 Pro-screenshot_first_vaccination_certificate_details_part1](https://user-images.githubusercontent.com/66998419/129439635-297f3143-685d-4e3e-8d5f-6c904ab89061.png)

dsarkar:
@MikeMcC399 Internal Tracking ID: EXPOSUREAPP-8503

Marcono1234:
[...]

PR #1663, opened 2021-08-24
Release 2.8 Screenshots + alt changes

larswmh:
In addition to the new screenshot section for 2.8, I also replaced the old undefined/falsely defined alts with correct alts. ([this issue](https://github.com/corona-warn-app/cwa-website/issues/1655)). Furthermore, I've removed every unused image from the 2.8 screenshots folder.


dsarkar (approved):

Ein-Tim:
Bildschirmfoto 2021-08-24 um 22 43 20

iOS screenshots in Android section?

MikeMcC399:
@Ein-Tim 
> iOS screenshots in Android section? This is also live currently, maybe we can change this with this PR?

This is the same issue as already reported for the 2.6 screenshots in https://github.com/corona-warn-app/cwa-website/issues/1549

Ein-Tim:
@MikeMcC399 Thanks for mentioning this issue. I knew that we have an issue reg. this but I couldn't find it. 
Anyways, I think this PR would be a good opportunity to fix this.

MikeMcC399:
@Ein-Tim 
>  I think this PR would be a good opportunity to fix this.

The Android app needs to automatically generate the required screenshots for the sections:

1. "Check certificates for validity before a trip"
2. "Invalid certificates"

and as far as I can see, the functionality to capture these screenshots has not yet been programmed into the Android app, so that would need a separate PR in https://github.com/corona-warn-app/cwa-app-android/pulls. The website PR can only deal with available screenshots.

@larswmh 
Should we submit a separate issue against the Android app to get these screenshots generated or is this already in hand internally?

dsarkar:
@MikeMcC399 @Ein-Tim The mentioned Android screenshots are not yet produced automatically (therefore not in the assets). We will correct this at a later stage. We have that internally tracked, stand by for ticket numbers. 

Thanks!

MikeMcC399:
@dsarkar 

Unfortunately there is a typo:

"In addition to the name in country-specific spelling, the name is now also displayed in standardized atin spelling in each certificate"

That should be "Latin" not "atin".

PR #1658, opened 2021-08-23
Fix bad indentation of JSON files

Marcono1234:
Fixes bad indentation of some JSON files, that means these are only whitespace changes.
I have fixed the indentation manually to not re-format all the files, so I might have missed a few cases.

PR #1684, opened 2021-08-29
Specify JSON formatter config

Marcono1234:
Proof of concept for specifying a JSON formatter config using [Prettier](https://prettier.io/).
This pull request is based on the changes of #1658.
See also https://github.com/corona-warn-app/cwa-website/pull/1658#issuecomment-907844052.

Changes:
- Change VSCode settings to use Prettier as formatter for JSON files
- Specify Prettier config using 4 spaces as JSON indentation
- Add npm scripts `format-json` for manually fixing formatting; `check-format-json` for checking JSON formatting.
Both are currently only run for JSON files under `src/data`.
- Change GitHub workflow to invoke `check-format-json`

Feedback is appreciated!


MikeMcC399:
@Marcono1234 
This seems like quite a major change. It might be better to open an enhancement request issue first so that the Open Source Team has the opportunity to provide feedback.

What happens to a contributor who is not using VSCode as editor? I would not want to see this editor made effectively mandatory. I personally use Atom on a regular basis,  for instance.

Ein-Tim:
I am also worried about this major change since such things are very likely to produce merge conflicts in the other open PRs.

Marcono1234:
> What happens to a contributor who is not using VSCode as editor? I would not want to see this editor made effectively mandatory. I personally use Atom on a regular basis, for instance.

The underlying formatter it uses is Prettier. The VSCode settings are there to help those whose are using VSCode.
At least based on the [Prettier website](https://prettier.io/), it seems to be widely supported by editors (including a package for Atom) so this should hopefully not be an issue.

> I am also worried about this major change since such things are very likely to produce merge conflicts in the other open PRs.

There is no hurry to merge (or even accept) this pull request. I just wanted to provide a proof of concept to address the suggestion @MikeMcC399 made.

MikeMcC399:
@Marcono1234 
You have some good ideas. Perhaps we can implement them with some smaller incremental steps which make improvements without causing disruptions to contributors' work or other unwanted side-effects?

Marcono1234:
For this suggestion I am not sure if it is possible to break this into smaller steps. Though note that the formatting changes of this pull request to the JSON files were produced by Prettier, so the maintainers can hopefully easily reproduce them and apply to them to their locally changed files by running `npm run format-json` (respectively using the VSCode formatter option).

I will wait for the comment of the maintainers on this pull request.

svengabr:
Currently, there are already 15 merge conflicts that involve some pretty huge files. We have multiple editors with multiple branches currently waiting for rollout. I personally can't approve this pull request because it would cause much trouble with little to no benefits.

Marcono1234:
@svengabr, would it be useful for you if I dropped all JSON file changes (therefore no conflicts) and the GitHub workflow change from this pull request?
Then you as well as external contributors could profit from unified JSON formatting (matching the desired formatting described in [`docs/CHANGE_WEBSITE_CONTENT.md`](https://github.com/corona-warn-app/cwa-website/blob/master/docs/CHANGE_WEBSITE_CONTENT.md#keep-in-mind)) by either using the VSCode "Format Document" command, the `format-json` npm script, or for other editors their respective Prettier support. But no one would be forced to adhere to this JSON formatting style yet. Especially for newly added JSON files I can imagine this being useful; there this consistent formatting could then be applied right from the start.

MikeMcC399:
@Marcono1234 
In https://github.com/corona-warn-app/cwa-website/pull/1726#issuecomment-917500211 you wrote that you were hoping to get feedback from your [previous comment here](https://github.com/corona-warn-app/cwa-website/pull/1684#issuecomment-910837570) , which was addressed to @svengabr.

If you are interested in pursuing your suggestion, then I would open a new enhancement request before investing any time in a new pull request. I'm not sure that there is the need for prettifier functionality, but it would be the Open Source Team to decide. 

json formatting hasn't been much of an issue in the past, although I am happy that we have at least got the faq.json / faq_de.json files formatted according to a standard. These are the two files which get a lot of editing done on them by various people. The other json files tend not to be edited so often, or the edits are done by one or two people only.

Thank you in any case for your ideas!

Marcono1234:
> I'm not sure that there is the need for prettifier functionality

The main point was to use an external tool (Prettier) to not be restricted to the built-in formatting support of VSCode, which also addressed your concern of using different editors such as Atom.
But either way, there is now a formatting configuration added by your pull request, so unless there is some active interest from the maintainers to use Prettier I won't create a new enhancement request or pull request.

MikeMcC399:
@Marcono1234 
When we started this discussion I didn't know what other editors were in use by regular contributors to this repository. @dsarkar and @Ein-Tim (two very active users here) both said they are using VSCode. I use Atom. I don't know if there are any other editors which are playing a role here. 
Although there are formatting differences between the json files in this repository, I don't think this is causing an issue. If the maintainers do think at any stage that there is an issue, we could revive this conversation and your suggestions. In the meantime I think you are right not to push for an enhancement request.
Thanks once again!

PR #1700, opened 2021-09-03
Fix nav tabs bug and change a word on eventregistration.json

brifemu:


src/assets/js/eventregistration.js
```diff
@@ -3,8 +3,13 @@ import { proto } from './lib/trace_location_pb';
 import { encode } from 'uint8-to-base64';
 import Cleave from 'cleave.js';
 import Papa from 'papaparse';
+import $ from 'jquery';
 import 'regenerator-runtime/runtime'
```
svengabr, line 8: Please use the existing tab javascript:
https://github.com/corona-warn-app/cwa-website/blob/master/src/assets/js/app.js#L233-L241

svengabr (changes requested):
Waiting for javascript to be resolved

src/assets/js/eventregistration.js
brifemu: Ok, done as you suggested

brifemu:
HTML tabs refactored taking by reference the tabs coded on the site.

PR #1725, opened 2021-09-10
Harmonize OS specific FAQ titles and correct typos

MikeMcC399:
This PR harmonizes all FAQ titles which are OS specific so they include a colon ":". This affects
- [Google/Android]
- [Apple/iOS]

titles.

For instance, the title of https://www.coronawarn.app/en/faq/#full_incompat, which is currently
`[Google/Android] A message “Incompatibility warning” on my Start Screen indicates that my device is not compatible with Exposure Notifications at all. What does this mean?`
is changed to
`[Google/Android]: A message “Incompatibility warning” on my Start Screen indicates that my device is not compatible with Exposure Notifications at all. What does this mean?`
like most of the other OS specific titles in the FAQs.

A few typos are corrected in both English and German FAQs.  A few words which used English (United Kingdom) spelling are changed to English (United States) spelling to conform to the spelling convention used in the Corona-Warn-App texts.


dsarkar (approved):

dsarkar:
@MikeMcC399 Thank you!

PR #1728, opened 2021-09-11
Update FAQ #keys_matches (closes #1028)

### LINKED ISSUE #1028: FAQ: "What do the exposure check logs show?" no longer matches Android UI
## Where to find the issue

**What do the exposure check logs show?**
https://www.coronawarn.app/en/faq/#keys_matches

**Was bedeutet die Überprüfung auf mögliche Begegnungen?**
https://www.coronawarn.app/de/faq/#keys_matches

## Describe the issue

Google changed the UI with v1.8 of the Exposure Notifications System. See [Exposure Notification release notes - v1.8, January 2021](https://developers.google.com/android/exposure-notifications/release-notes#v18_january_2021).

In Android there is no count "Exposure checks" any more.

The Hash value changes for every risk evaluation.

There are other inaccuracies in the English version noted in the next section.

## Suggested change

For Android:

The following section needs to be removed:

```
Exposure checks

This value does not reflect the number of encounters. 
Please tap one level further: This is where the checks of up to 15 days are logged (today plus the last 14 days). 
Selecting an entry leads to a detailed view with the corresponding time stamps and the number of matched keys.
```

The second sentence below does not make sense. It needs to be re-translated from the correct German.
```
Timestamp

Date and time at which the check happened. 
This is independent of the timestamp of potential exposures since every key stored in the downloaded files, a separate validity timestamp is stored.
```

The Android label is "App" not "Data Source" (English and German)
```
Data Source

Identifies the app that initiated the check. 
If you, for example, also installed Immuni or SwissCovid, their app identifiers could be noted there instead of 'Corona-Warn'
```

The Hash value changes every time. The second sentence should be deleted.
```
Hash

Checksum of the file containing the downloaded diagnosis keys. 
As long as the file content (diagnosis keys) does not change, this checksum will not change throughout the relevant time frame of 2 weeks.
```

MikeMcC399:
This PR resolves the issue #1028 "FAQ: "What do the exposure check logs show?" no longer matches Android UI".

It updates 

- https://www.coronawarn.app/en/faq/#keys_matches "What do the exposure check logs show?"

- https://www.coronawarn.app/de/faq/#keys_matches "Was bedeutet die Überprüfung auf mögliche Begegnungen?"
## Changes made

1. The Google UI no longer displays a count of the number of exposure checks which have been carried out. The text referring to this number has been removed.
2. A sentence has been added to clarify that multiple checks can be made per day.
3. The steps to access the list of exposure checks has changed. To avoid making the article too complicated, by giving exact navigation steps for Android and iOS, reference to tapping levels has been removed.
4. The label for terms for Apple/iOS has been changed from "Apple" to "iOS" to align to the Google/Android use of the term "Android" only.
5. The text about the hash sum not changing has been removed, since this is not true.
6. Other minor wording changes have been made.


Ein-Tim (approved):
LGTM! Only checked if what's written there is actually the implementation under iOS.

MikeMcC399:
@dsarkar 
Could you please review this PR which improves the accuracy of the "FAQ: "What do the exposure check logs show?" ?

dsarkar (approved):

dsarkar:
@MikeMcC399 @Ein-Tim Thanks!

Issue #1781, opened 2021-09-30
Add link checking to https://www.coronawarn.app


MikeMcC399:
## Feature description

For website https://www.coronawarn.app add link checking to site tools.

This follows on from the question https://github.com/corona-warn-app/cwa-website/issues/889.

Consider integrating link checking into the checks done in workflows:

- https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml and
- https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml

Also, consider regularly monitoring https://www.coronawarn.app in order to catch changing and dead external links.

## Problem and motivation

1. There is no check done when changes are made to the website https://www.coronawarn.app which might ensure that existing (internal) links are not broken. For example see issue https://github.com/corona-warn-app/cwa-website/issues/1778.
2. If external links change, there is no check done to ensure that the links on the website https://www.coronawarn.app/en/ continue to work. For instance there are currently at least two dead external links (see  https://github.com/corona-warn-app/cwa-website/issues/1779 and https://github.com/corona-warn-app/cwa-website/issues/1780) on the website.

## Is this something you're interested in working on

Unfortunately I don't have the experience to implement this.

---
Internal Tracking ID: [EXPOSUREAPP-9734](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9734)

dsarkar:
@MikeMcC399 Thanks. We will look into this. Internal Tracking ID: [EXPOSUREAPP-9734](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9734)

MikeMcC399:
https://www.deadlinkchecker.com gives some false positives for e-mail obfuscation and links with umlauts in them, but it does also find some dead links on https://www.coronawarn.app/en/ which in some cases have been there for quite a while.

For instance https://www.coronawarn.app/de/blog/2021-05-12-corona-warn-app-version-2-2/, which is four months old, includes a link 

`<a href="+493049875401">+49 30 498 75401</a>` which resolves to the bad URL

`https://www.coronawarn.app/de/blog/2021-05-12-corona-warn-app-version-2-2/+493049875401`

that is probably intended to be 

`<a href="tel:+493049875401">+49 30 498 75401</a>`

as used in the English language version of the page on https://www.coronawarn.app/en/blog/2021-05-12-corona-warn-app-version-2-2/

[...]

PR #1786, opened 2021-10-01
Cypress - check for broken links

dsarkar:
- Mitigates issue https://github.com/corona-warn-app/cwa-website/issues/1781
- checks coronawarn.app pages for broken links

ToDo:
- filter out email and tel
- filter out duplicates, e.g. #top
- add all pages

Usage on local built:
- `npx Cypress open`
- click on check_links.js
image
- execution
image
- see result, e.g.
- 
image

---
Internal Tracking ID: [EXPOSUREAPP-9734](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9734)


MikeMcC399:
@dsarkar 
This is a great addition to the cypress tests!

It looks like you are not testing the whole site though, just selected sections. `/en/blog/` & `/de/blog` for instance is missing. What about the top level `/en` and `/de`? The original analysis PR broke some links at the top level, so it would be good to check there as well.

`npm run test:open`  then selecting `check_links.js` runs the test as expected.

I tried also running `npm run test` where all available tests are chained. 

The goal of course should be to get `check_links.js` and the website itself in a state where the tests all pass!

mtb77 (approved):

PR #1880, opened 2021-10-15
Fix fig. 15 in science blog 3 and refactor science.js

brifemu:
In this PR I have fixed the fig. n 15. Now is interactive. Also, I removed science.js file and dive the content on independent scripts loaded on science.html header

https://user-images.githubusercontent.com/44208107/137455009-58557da0-c7a8-43b8-9971-027103fb8478.mp4


brifemu:
Closed to split science.js changes and fig fix

PR #1894, opened 2021-10-16
Remove double sentences from science blog 4

Ein-Tim:
This PR removes some sentences which were in the science blog 4two times.


dsarkar:
Good morning @Ein-Tim ! Alrighty, thanks again! Enjoy your Sunday!

PR #1915, opened 2021-10-19
Add faq article root_detection_android

larswmh:
This pull request solves [New FAQ R2.12 #root_detection_android #1914
](https://github.com/corona-warn-app/cwa-website/issues/1914)
EN:
chrome_muaiB5VpZf

DE:
chrome_yswdtx0gYl


MikeMcC399:
Using `rel='noopener noreferrer'` on internal links is unnecessary and that is another topic I'd like to bring up as a general topic soon. It's only necessary for external links and only as protection for vulnerabilities that existed with older browser versions.

Issue #1792, opened 2021-10-01
Different analysis pages and their links


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/analyse/
- https://www.coronawarn.app/de/analyse/
- https://www.coronawarn.app/en/analysis/
- https://www.coronawarn.app/de/analysis/

## Describe the issue

There are now 4 analysis pages published to production.

The two pages from a previous publication include "Analyse" in the header menu. (See https://github.com/corona-warn-app/cwa-website/issues/1776)
- https://www.coronawarn.app/en/analyse/
- https://www.coronawarn.app/de/analyse/

The PR for the two new pages
- https://www.coronawarn.app/en/analysis/
- https://www.coronawarn.app/de/analysis/

didn't include any link to "Corona-Warn-App (CWA): Kennzahlen" from 
- https://www.coronawarn.app/en/ or
- https://www.coronawarn.app/de/

## Suggested change

I'm not sure of the current intention, but probably 
- https://www.coronawarn.app/en/analyse/
- https://www.coronawarn.app/de/analyse/

need to be manually removed from the production server.

If the idea is to keep the pages
- https://www.coronawarn.app/en/analysis/
- https://www.coronawarn.app/de/analysis/

hidden from view until they have been tested in production, then fine, otherwise a link to the pages needs to be added to 
- https://www.coronawarn.app/en/ or
- https://www.coronawarn.app/de/

MikeMcC399:
This issue is partially addressed by PR https://github.com/corona-warn-app/cwa-website/pull/1793.

dsarkar:
@FabioFuschi-SNK FYI

MikeMcC399:
@Ein-Tim 
> The menu now includes a link to https://www.coronawarn.app/de/analysis/ or https://www.coronawarn.app/en/analysis/, so I guess this is the page which will be used.

That is what I meant when I wrote:

> This issue is partially addressed by PR #1793.

The first part of the issue reported is not yet resolved.

Ein-Tim:
@MikeMcC399 

Ah, sorry, I didn't understand this. (:

dsarkar:
@MikeMcC399 Thanks, yes, we will have a look!

dsarkar:
@MikeMcC399 We will check your proposal at some point. We from the community team do not have access.

PR #1919, opened 2021-10-19
Fix analyse redirect

brifemu:
In this PR I had fixed issue #1792 

Due to we don't have access to FTP server, I inserted again 'analyse' page which redirects to Analysis page. Without access to the FTP server, I think this is the only one solution.


MikeMcC399:
@Ein-Tim 
For me the prod workflow shows as finished now. 👍🏻 

It was retriggered was successful the second time.

image

Issue #1986, opened 2021-10-29
Cypress check_anchor_links.js fails with syntax error


MikeMcC399:
## Where to find the issue

https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_anchor_links.js

## Describe the issue

The Cypress test check_anchor_links.js fails with "Syntax error, unrecognized expression: #" in /de/blog entries and /en/blog entries

Execute:
```
npm run test:prepare
npm run test:open
```
click on `check_anchor_links.js` in Cypress console to run test

![image](https://user-images.githubusercontent.com/66998419/139408694-befd7ac1-e3a7-4b47-940a-0b8a2aa96520.png)

MikeMcC399:
@dsarkar 
Have you executed
`npm run test:prepare`
on your system as instructed in https://github.com/corona-warn-app/cwa-website/blob/master/README.md#testing?

As far as I can see, [check_anchor_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_anchor_links.js) fails checking
`en/blog/2099-01-01-visual-test-page/index.html`
Line 166
`<li>diam nonumy eirmod <a href="#">link to Google</a></li>`

which is generated from
https://github.com/corona-warn-app/cwa-website/blob/master/cypress/fixtures/blog/2099-01-01-visual-testing/index.md
Line 21
`- diam nonumy eirmod [link to Google](#)`

It looks like [check_anchor_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_anchor_links.js) doesn't deal with the case of a simple anchor reference consisting of just `#`. This is a valid link to the top of the page.

If I delete the directory
`cwa-website/blog/2099-01-01-visual-testing`
then execute
`npm run build`
`npm run test:open`
then select `check_anchor_links.js` to run in the Cypress console, the test shows no error.

(`npm run test:open` sometimes times out on the first run, which is why I wrote `npm run build` as the first step above. The timeout is a different issue though.)

MikeMcC399:
I have opened a new issue regarding the timeout when running a Cypress test from npm - see https://github.com/corona-warn-app/cwa-website/issues/1993.

MikeMcC399:
This issue is resolved by PR #1996. Thank you to @brifemu!

PR #1996, opened 2021-11-02
Fix anchor link Cypress test

brifemu:
Due to issue #1986 I had added two more conditions to search for a container ID. Now the anchor must be different from 'top' and also apart from have a '#', the part behind '#' have to contains content. Let's see some examples:

Our test will check for:

- page.com/url#tag

Our test will avoid:

- page.com/url#top
- page.com/url#

### Why I included #top like a condition?
Because it is very common to use this tag to go to the top of the page without the necessity of exist a container with a 'top' ID

### Advantages
- The test will works for everyone without errors

### Disadvantages
- If an editor forgot to include content behind '#' the test will not detect it or can be confused by use '#' to go to the top of the page

### DIFF
--- a/cypress/integration/check_anchor_links.js
+++ b/cypress/integration/check_anchor_links.js
@@ -76,7 +76,7 @@ context("Check for broken anchor links on entries", () => {
         if (url.prop('href') && url.prop('href').includes("localhost") && url.prop('href').split("#")[1] !== "top" && url.prop('href').includes(sub) && !pagesToAvoid.includes(url.prop('href').replace('http://localhost:8000', '')))  {  
           cy.visit({log: false, url: url.prop('href')} )
           cy.get("a[href*='#']").each(link => {
-            if(link.prop('href').includes(url.prop('href'))) {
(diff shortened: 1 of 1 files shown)


MikeMcC399:
After executing:
```
npm run test:prepare
npm run test:open
```
then selecting `check_anchor_links.js` in Cypress console, the test runs with 31 passes and 0 fails:

image

From my point of view this should be merged. The current test in the `master `branch is failing and this should be replaced as soon as possible.

----
Running the full test
`npm test`
on Windows often produces errors like

```
Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:209:20)
 {
  errno: -4077,
  code: 'ECONNRESET',
  syscall: 'read'
}
Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:209:20)
````    
or
`
 CypressError: `cy.screenshot()` timed out waiting `50000ms` to complete.
`

or the warning:

```
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
```

This is a more general problem about the reliability of this test suite though. It should not stop this PR being merged.

dsarkar (approved):

dsarkar:
@brifemu @MikeMcC399 Thanks!

Issue #1998, opened 2021-11-02
FAQ menu is not useable on iPad Air (4th generation) in horizontal mode


Ein-Tim:
## Where to find the issue
- https://www.coronawarn.app/en/faq/
- https://www.coronawarn.app/de/faq/

## Technical details
- Device name: iPad Air 4. Gen
- iPadOS version: 15.1

**This might also affect other tablets or in general other devices with a similar screen size.**

## Describe the issue
The FAQ menu is not visible/useable on the 4. Gen iPad Air if it's in horizontal mode.
If the device is in landscape mode, everything works.
Here's a screenshot:
<img src="https://user-images.githubusercontent.com/67682506/139956954-49e73f88-0111-4e13-b6eb-8ca357cb2198.png" width="300">

## Suggested change
Fix the issue.

## Additional information
This bug seems to be reproducible using Google Chrome dev tools.

MikeMcC399:
This problem is also reproducible on a desktop version of Google Chrome or Mozilla Firefox. Start with a browser window showing https://www.coronawarn.app/en/faq/ using the full height and width of the screen. Reduce the width of the browser window gradually. The sequence is:

1. Left menu fully visible. All menu items visible.
2. Menu column width reduced. Menu items start to wrap.
3. Menu column width further reduced. Menu items start to wrapped further. Vertical scroll bar appears.
4. Menu column width further reduced. Menu items wrapped and truncated. Horizontal scroll bar appears. (Need to scroll down main window to see horizontal scroll bar of menu items.)
5. Menu column completely collapsed. No horizontal scroll bar visible. No arrow available.
6. Menu column completely collapsed. No horizontal scroll bar visible. Vertical scroll bar for menu items disappears. No arrow available.
7. Arrow appears which, when clicked, overlays the menu items on top of the main window.

The part of the sequence where the menu items are not readable or have disappeared and are not accessible is a problem. On a desktop the problem is not so big because the browser window width can be changed. Setting the window either wider or narrower works around the issue. The user of a mobile device does not have quite this flexibility of changing the width of display.

Ein-Tim:
@MikeMcC399 

Thank you for your feedback! 

Yes, I also experienced these 3 problems when playing around with this in the Google Chrom dev tool.
Either the arrow to access the menu is shown, then everything is fine, or the menu is cut off, which is not good, or the menu is just not shown at all and one has no chance to access it.

Ein-Tim:
[...]

PR #1999, opened 2021-11-03
Fix FAQ menu for iPad Air 4th gen

brifemu:
Due to issue #1998 I had changed the breakpoint size for $screen-md from 862px to 918px, is when the menu starts to break.

Is very important to test the menu and all the pages on a real iPad Air 4th gen before merge this, so please @Ein-Tim give me back your feedback after test it.


Ein-Tim:
@brifemu 

Thank you for the quick fix! 
I will test this change as soon as I can!

Ein-Tim (commented):
I verified this change on my iPad Air 4. Gen, the FAQ page now looks like this:

IMG_A4BA4A8A87D5-1

This is definitely better than before. But one thing which should be changed is the arrow to expand the menu, it is currently cut off when opening the FAQ page. Only if one scrolls down a bit it becomes fully visible.

This change,however, also changed how the menu on the screenshot page is accessible. 
While it was visible without any problems before this change, it's now only visible after clicking on the arrow to expand it. But this does not seem like a big issue to me.

| Screenshots page before this change | Screenshots page after this change |
|---|---|
| IMG_0026 | IMG_0025 |

Also, this change changed the behaviour of the menu (of the whole coronawarn.app site) to not be shown like this anymore:


It's now shown like this:


This is a welcome improvement IHMO.

brifemu:
Thank you so much @Ein-Tim for your feedback.

That is why it is so important to test the real device, simulating it in Chrome's DevTools, the menu button error did not happen.

I had increased the bottom space for the arrow container. Is weird that the arrow position only changes on FAQ page, to prevent, I had increased it 25px in the main style.

I will wait for your feedback to continue working on it.

Ein-Tim:
@brifemu 

> Thank you so much @Ein-Tim for your feedback.

For sure! Thank you for fixing this issue! 

> I had increased the bottom space for the arrow container. Is weird that the arrow position only changes on FAQ page, to prevent, I had increased it 25px in the main style.
I will wait for your feedback to continue working on it.

I retested based on the latest commit [`01d67da` (#1999)](https://github.com/corona-warn-app/cwa-website/pull/1999/commits/01d67dae61cb2da6765d3c81308bb8da327af149), but sadly, this bug still persists.

Here's a screenshot. I added a red "circle" so that it's visible what the problem is:

brifemu:
@Ein-Tim I gave it again more bottom space. I discovered that the issue was caused to have more than one tab open, the browser creates the bar and that affected the arrow position.

Ein-Tim:
Thanks @brifemu 

I will test again.

Ein-Tim (approved):
Works as intended now!

Only one tab opened:


More than one tab opened (the arrow is visible but the box which is around it is still cut off a bit):

brifemu:
@Ein-Tim can you try it again with a smaller system font size, please? I guess the problem is the system font size. Someone with default font size would see the arrow higher.

Ein-Tim:
@brifemu 

I can do this, but I did not change the font size (=I should be using the default one).

Ein-Tim:
So, sorry for the delay, was a busy day.

When setting the font size to the smallest one possible, the box around the arrow is still cut off, but **this is really no big problem as it does not affect the functionality of the button in any way.**

**Screenshots**:

iPad Air 4. Gen only one tab opened:

| Scrolling position | Currently | After this change |
|---|---|---|
| Not scrolled down |   |  |
| Scrolled down |  |  |

iPad AIr 4. Gen two tabs opened:

| Scrolling position | Currently | After this change |
|---|---|---|
| Not scrolled down |  |  |
| Scrolled down |  |  |

This is good to go for me! Also tested on iPhone XR. iPhone 6s & Huawei P Smart 2019. No issues.

dsarkar (approved):

dsarkar:
@Ein-Tim Thanks for report and all the feedback!

Ein-Tim:
@dsarkar 

For sure! Thanks for fixing this so quick! Have a nice weekend!

PR #2008, opened 2021-11-03
Update FAQ: cert_recycle_bin

larswmh:
There was a mistake in the description about deleting certificates from the recycle bin on iOS. Unlike mentioned on the website right now, you can't swipe to the left to delete a certificate if you tapped 'Edit' first.

DE:
image

EN: 
image


MikeMcC399:
@larswmh 
Will you also change the term "trash" to "recycle bin"?

MikeMcC399:
You don't need to refer to the name of the Start Screen in previous versions, because they did not have the function of the "Recycle Bin".

MikeMcC399:
There is also a bug in the 2.13.1 Android app. It offers "Delete All Permanently" / "Alle endgültig löschen" and then only deletes one certificate. That doesn't make sense to me.

I will check the 2.14 branch to see it if is already fixed.

MikeMcC399:
https://github.com/corona-warn-app/cwa-app-android/issues/4323 logged, since the misleading text "Delete All Permanently" is in both 2.13.x and 2.14.x.

dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks for reviewing! Very much appreciated!

Issue #2006, opened 2021-11-03
Screenshots: Trash now named Recycle Bin (EN)


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/screenshots/#android_recycle_bin

## Describe the issue

In the Android [release/2.13.1-RC1](https://github.com/corona-warn-app/cwa-app-android/tree/release/2.13.x) the place where removed certificates go is called the "Recycle Bin". During the development of 2.13.x it was earlier called "Trash".

The screenshots in https://www.coronawarn.app/en/screenshots/#android_recycle_bin use both names. Only "Recycle Bin" is correct. "Trash" is no longer correct.

Number | Screenshot | Name used
--- | --- | ---
1 | [RecyclerBin_from_menu.png](https://www.coronawarn.app/assets/screenshots/2-13/en/android/RecyclerBin_from_menu.png) | Recycle Bin
2 | [RecyclerBinOverviewFragment_empty.png](https://www.coronawarn.app/assets/screenshots/2-13/en/android/RecyclerBinOverviewFragment_empty.png) | Trash
3 | [RecyclerBinOverviewFragment_full.png](https://www.coronawarn.app/assets/screenshots/2-13/en/android/RecyclerBinOverviewFragment_full.png) | Trash
4 | [RecyclerBinOverviewFragment_full_2.png](https://www.coronawarn.app/assets/screenshots/2-13/en/android/RecyclerBinOverviewFragment_full_2.png) | Trash

## Suggested change

- Android English Screenshots from https://www.coronawarn.app/en/screenshots/#android_recycle_bin for the Recycle Bin numbered 2, 3 and 4 should be replaced with ones using the name "Recycle Bin"

MikeMcC399:
The FAQ https://www.coronawarn.app/en/faq/#cert_recycle_bin refers to "trash" for iOS in one place. I don't have an iOS device to check, but perhaps this also needs to be corrected?

Ein-Tim:
iOS calls it "Recycle Bin":

<img src="githubusercontent.com/67682506/140082557-bc2c5bb1-82b7-4a5c-9abe-2010f11e320a.png width="300">

MikeMcC399:
Thanks @Ein-Tim!
I didn't want to rely on the screenshots on cwa-website because for Android they need correcting.

MikeMcC399:
[...]

PR #2010, opened 2021-11-04
Android Screenshots: Replace 'Trash' with 'Recycle Bin'

larswmh:
This PR solves #2006.
image


MikeMcC399:
@larswmh 
This is better than before, but still not right.

In src/assets/screenshots/2-13/en/android/RecyclerBinOverviewFragment_empty.png it shows the text:
"The Trash is Empty"

The text in the app which I see on my test mobile is however:
"The Recycle Bin is Empty"
I am looking at CWA 2.13.1-RC1 as I do not have the released version delivered to me yet via Google Play Store.

MikeMcC399 (approved):
Now it looks good!
I think this can be merged now.

dsarkar (approved):

PR #2025, opened 2021-11-06
Change anchor "vac_cert_invalid_211111" to "vac_cert_invalid_211115" & add FAQ redirect

Ein-Tim:
This PR changes the anchor "vac_cert_invalid_211111" to "vac_cert_invalid_211115", as the date was changed from 11.11.20212 to 15.11.2021.
I also added a FAQ redirect and confirmed that this works in German and English locally.

### DIFF
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -335,7 +335,7 @@
                     },
                     {
                         "title": "Warum erhalte ich die Meldung, dass mein Zertifikat neu ausgestellt werden muss?",
-                        "anchor": "vac_cert_invalid_211111",
+                        "anchor": "vac_cert_invalid_21_1",
                         "active": true,
                         "textblock": [
                             "Diese Meldung informiert Sie darüber, dass Ihr Impfzertifikat ab dem 15.11.2021 nicht mehr gültig ist und Sie sich rechtzeitig um ein neues digitales Zertifikat bemühen sollten.",
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -329,7 +329,7 @@
                         "active": "true",
                         "textblock": [
                             "<ul>",
-                            "<li><a href='#vac_cert_invalid_211111' target='_self'>Warum erhalte ich die Meldung, dass mein Zertifikat neu ausgestellt werden muss?</a></li><li><a href='#vac_cert_booster_info' target='_self'>Wie erfahre ich von Auffrischungsimpfungen im Kontext von Corona?</a></li><li><a href='#vac_cert_booster_cwa' target='_self'>Wie verwalte ich Auffrischungsimpfungen in der CWA?</a></li><li><a href='#vac_booster_notification' target='_self'>Warum erhalte ich eine Benachrichtigung über eine Impfempfehlung der Ständigen Impfkomission (STIKO)?</a></li>",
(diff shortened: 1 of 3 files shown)


MikeMcC399:
@Ein-Tim / @dsarkar 
I wonder if it would be better to code the anchor with a sequence number instead of the exact date, in case there is another need to revoke certificates and there is uncertainty about the date.

So instead of using `vac_cert_invalid_211115`, make it `vac_cert_invalid_21_1` as (year) + (sequence number). The next one would be `vac_cert_invalid_21_2`, ... 21_3, ... 21_4 etc. and next year  `vac_cert_invalid_22_1`, ... 22_2, ... 22_3 etc.

Ein-Tim:
@MikeMcC399

Yes, I would also prefer a solution where we don't put any dates into the anchor. I would be happy to change that in this PR, however, I'll first wait for feedback from the Open-Source-Team.

dsarkar:
@Ein-Tim @MikeMcC399 Thanks for the proposals. Let's go then for vac_cert_invalid_21_1. @Ein-Tim you change the PR?

Ein-Tim:
@dsarkar 

Yes, I will change the PR.

Ein-Tim:
Changed @dsarkar! Also tested locally and confirmed that the redirect works.

dsarkar (approved):

src/data/faq_de.json
```diff
@@ -329,13 +329,13 @@
...
                             "<ul>",
-                            "<li><a href='#vac_cert_invalid_211111' target='_self'>Warum erhalte ich die Meldung, dass mein Zertifikat neu ausgestellt werden muss?</a></li><li><a href='#vac_cert_booster_info' target='_self'>Wie erfahre ich von Auffrischungsimpfungen im Kontext von Corona?</a></li><li><a href='#vac_cert_booster_cwa' target='_self'>Wie verwalte ich Auffrischungsimpfungen in der CWA?</a></li><li><a href='#vac_booster_notification' target='_self'>Warum erhalte ich eine Benachrichtigung über eine Impfempfehlung der Ständigen Impfkomission (STIKO)?</a></li>",
+                            "<li><a href='#vac_cert_invalid_21_1' target='_self'>Warum erhalte ich die Meldung, dass mein Zertifikat neu ausgestellt werden muss?</a></li><li><a href='#vac_cert_booster_info' target='_self'>Wie erfahre ich von Auffrischungsimpfungen im Kontext von Corona?</a></li><li><a href='#vac_cert_booster_cwa' target='_self'>Wie verwalte ich Auffrischungsimpfungen in der CWA?</a></li><li><a href='#vac_booster_notification' target='_self'>Warum erhalte ich eine Benachrichtigung über eine Impfempfehlung der Ständigen Impfkomission (STIKO)?</a></li>",
```
Ein-Tim, line 332: @dsarkar Does this need to be changed in English too? Should I quickly provide a PR or will you do so?

MikeMcC399: @Ein-Tim 
It needs changes because otherwise it is broken.

MikeMcC399:
@Ein-Tim 
The link in https://www.coronawarn.app/en/faq/#vac_cert_booster is now broken because it points to the now non-existent
https://www.coronawarn.app/en/faq/#vac_cert_invalid_211111

src/data/faq_de.json
Ein-Tim: @MikeMcC399 

For me it still works (maybe because I've set my browser to always open links in new tabs?). But I'll provide a PR fixing this in a few minutes.

Ein-Tim:
@MikeMcC399 

PR is coming in a few minutes.

Ein-Tim:
PR #2031.

PR #2046, opened 2021-11-09
blog - fix urls

dsarkar:


MikeMcC399:
Actually using the `../` construction for links in blogs is a bit tricky. That's why /en and /de are more reliable.

If you use `../` it depends if you put a link in the overview of a blog or in the detail text. I the overview you need `../` and in the detail you need `../../` which is easy to get wrong.

dsarkar:
so we keep it as it is.

dsarkar:
Thanks @MikeMcC399

PR #2047, opened 2021-11-09
Mention luca in check_in_qr_code

Ein-Tim:
This PR updates https://www.coronawarn.app/de/faq/#check_in_qr_code to reflect that one can now scan luca QR codes.

PR #2050, opened 2021-11-09
FAQ Update check_in_index, check_in_luca and check_in_qr_code

larswmh:
This PR includes a change to FAQ `check_in_qr_code`, which I have been already working on when it was suggested in #2047. Still many thanks to @Ein-Tim for quickly noticing. Apart from that, the terms `luca app (EN)` and `luca-App (DE)` use the same terminology anywhere now. `check_in_luca` questions have been added to `check_in_index` so users can quickly see a summary of all answered questions.

EN:
image

DE:
image


larswmh:
Unfortunately an error occured in the german file. I'll open another PR for the german version

PR #2051, opened 2021-11-09
Update FAQ entries about interoperability of event QR codes with other contact tracing apps

floscher:
This updates this section of the FAQ to now say that CWA-QR-codes can not be scanned by other apps, but the CWA can scan both CWA-codes as well as luca-codes.


floscher:
I just saw, that #2050 and #2047 also propose similar changes.

dsarkar (approved):

larswmh (approved):

dsarkar:
Text changes including suggestions from @Ein-Tim

larswmh:
@floscher thanks for your pull request! As you have already noticed, we've made similar pull requests at the same time, we ended up adjusting yours a little bit and merged it

Ein-Tim:
Okay, I see, we 3 now worked on the same thing...
In the future I'll create an issue for such things first to better coordinate and don't have unnecessary efforts (-;

floscher:
@Ein-Tim No worries, I wouldn't have seen an open issue on the topic either. Because it was such a small change, I only looked up, if the `master` branch was updated, before making the change. And then only realized that there already were PRs for that after creating mine.
@larswmh thanks for adding our changes!

PR #2058, opened 2021-11-10
Social media icons on footer

brifemu:
In this PR I have included the icons of CWA social media on the footer.

### Before

image

### After

image

---
Internal Tracking ID: [EXPOSUREAPP-9865](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9865)


Ein-Tim:
@brifemu 

Don't forget to also change https://github.com/corona-warn-app/cwa-event-landingpage.

MikeMcC399:
The Instagram link https://www.instagram.com/rki_fuer_euch/ doesn't seem to be specifically for the CWA app, so the title is a bit misleading.

Also there has not been any announcement about Instagram so far.

Maybe there is something planned?

```
"title": "The Corona-Warn-App project on Instagram",
 "url": "https://www.instagram.com/rki_fuer_euch/",
```

image

dsarkar:
@MikeMcC399 Thanks for the comment. We will check if this is the correct link.

MikeMcC399:
@dsarkar

I can't see anything new and specific to the Corona-Warn-App on https://www.instagram.com/rki_fuer_euch/.

I did notice though that there is an Instagram site https://www.instagram.com/covidwarnapp/ but that seems to have quite old content. It is also unclear who owns this account. Maybe RKI should follow up the same way that a follow-up was done for the Twitter account https://twitter.com/coronawarnapp which was originally not created or owned by RKI? See https://github.com/corona-warn-app/cwa-wishlist/issues/364.

PR #2089, opened 2021-11-16
Add eu_dcc_leaked

jucktnich:
This is an early staged FAQ-entry for the process if a DCC is leaked. If the team is ok with the general changes i will put real work into it.
Thanks @fynngodau for the email text.


MikeMcC399:
@jucktnich 

It's usually better to start out with an enhancement request issue for something like this, rather than create a PR directly. You can then get feedback about whether it would be welcomed and if there are any other suggestions.

Also for better readability, it helps if you post a screenshot of the text as it would appear on the website.

The primary authority in Germany for digital Covid certificates is https://digitaler-impfnachweis-app.de/ so if RKI thinks the advice you propose should be published it might be better on https://digitaler-impfnachweis-app.de/faq.

jucktnich:
Ah ok, yeah makes sense 😅
Should I leave this pr open here, or how should I proceed?

MikeMcC399:
@jucktnich 
I would wait for a comment from the Open Source Team first before putting more effort in.

jucktnich:
@MikeMcC399 Yeah this for sure, but should I close this and open a new one at the Digitaler-Impfnachweis repo

MikeMcC399:
See https://github.com/corona-warn-app/cwa-documentation/issues/751#issuecomment-970637358 regarding a similar point.

dsarkar:
image

@jucktnich @MikeMcC399 Thank you very much.

@jucktnich Do you have knowledge that this has occurred and therefore is the need of this FAQ?

jucktnich:
@dsarkar yes this happend. There are multiple instances (mostly the BR 😅) on the slack server

jucktnich:
The #1 reason for this FAQ entry would be a trust anchor for such mails

MikeMcC399:
There is a related post https://github.com/eu-digital-green-certificates/dgca-validation-service/issues/76#issuecomment-1068032796 which says that leaked certificates are not a security issue.

It seems in any case that this subject is not best covered in the Corona-Warn-App. I would suggest closing this PR.

Ein-Tim:
I second the suggestion of @MikeMcC399 to close this PR now.

dsarkar:
As suggest by @Ein-Tim and @MikeMcC399 closing this PR

PR #1699, opened 2021-09-03
Event multi QR code generator

svengabr:
Continue of
https://github.com/corona-warn-app/cwa-website/pull/1698

---
Internal Tracking ID: [EXPOSUREAPP-8495](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8495)

PR #2105, opened 2021-11-22
Reintroducing multiQR functionality

brifemu:
In this PR, I have solved the problems occurred from #1699 

- Dependence moment.js was replaced by day.js to solve issues using Safari.
- Code refactored (+100 lines optimized).
- Now generate a PDF, not pages like images.
- The PDF generated can be seen at the site like a slider.
- Working in all browsers.


src/data/eventregistration_de.json
```diff
@@ -50,7 +50,7 @@
                 "wrongfileheaders": "Die Header der .csv-Datei entsprechen nicht den Headern der entsprechenden Vorlage, Sie können sie <a href='. /event-qr-code-guide/'>hier überprüfen.</a>",
                 "wrongfileformatfields": "Einige Felder in Ihrer CSV-Datei sind nicht korrekt ausgefüllt. Bitte überprüfen Sie Ihre .csv und versuchen Sie es erneut",
                 "filewithnodata": "Ihre .csv Felder sind nicht ausgefüllt",
-                "fileoverloaded": "Die Erstellung von QR-Codes wurde auf 100 . begrenzt"
+                "fileoverloaded": "Die Erstellung von QR-Codes wurde auf 100 begrenzt."
```
Ein-Tim, line 53: Wouldn't it make more sense to write "Die Erstellung von QR-Codes ist auf 100 begrenzt." "wurde begrenzt" means that it was possible for some time to create more than 100 QR codes but now it was limited to 100. This is not the case here. 
Same in English.

MikeMcC399:
@dsarkar 
When I execute `npm install` on this branch it makes a change to `package-lock.json`.

image

so it seems that the `package-lock.json` in this branch is not quite in sync with the `package.json` from the branch.

dsarkar:
@MikeMcC399 Yes, it changes package-lock.json, since there are new dependencies.

MikeMcC399:
@dsarkar 
I understand that there are new packages defined in `package.json`
```
"dayjs": "^1.10.7",
"jspdf": "^2.4.0",
"papaparse": "^5.3.1",
```

which lead to changes in `package-lock.json`. My comment was just about a minor difference, not about most of the changes.

Can I suggest the following?
Execute this in `cuisines/cwa-website`:
```
git switch master
npm ci
```
which deletes and re-installs `node_modules` exactly according to the `master `branch version of `package-lock.json`

then
```
git switch brifemu-eventregistration-refactor
npm install
git diff
```

and I think you will see a small difference in `package-lock.json` which should be committed.

When I carried out these steps I got:
```
$ git diff
diff --git a/package-lock.json b/package-lock.json
index 062a3829..a1b7488d 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -13488,7 +13488,8 @@
       "dependencies": {
         "ansi-regex": {
           "version": "5.0.0",
-          "resolved": ""
+          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-5.0.0.tgz",
+          "integrity": "sha512-bY6fj56OUQ0hU1KjFNDQuJFezqKdrAyFdIevADiqrWHwSlbmBNMHp5ak2f40Pm8JTFyM2mqxkG6ngkHO11f/lg=="
         },
         "ansi-styles": {
           "version": "4.3.0",
```

MikeMcC399:
@dsarkar 
One other minor comment,

if you add `papaparse` from the command line

`npm install papaparse@^5.3.1 --save`

npm will add it to `package.json` in alphabetical order in the dependencies:

All the other dependencies are in alphabetical order in  `package.json` which is helpful for maintenance. It does not affect the functionality however.

src/data/eventregistration_de.json
dsarkar: fixed in this commit https://github.com/corona-warn-app/cwa-website/pull/2105/commits/c56895e8cbcccada09a59da130e0423820a908d0

dsarkar: alright, you are right, it is about the tense.

Ein-Tim: Thank you @dsarkar!

dsarkar: thanks for review

svengabr:
@MikeMcC399 thanks for your input! I did execute your suggested changes.

src/data/eventregistration_de.json
Ein-Tim: You're very welcome!

MikeMcC399:
@svengabr 
> I did execute your suggested changes.

Perfect 👍🏻 

That resolves the issues I pointed out. Thank you!

benbender:
Is there any ETA when this PR will land?
I'm asking especially as there might be a growing demand for this functionality as the constraint to deploy the luca-app may drop in several federal countries...

dsarkar:
@benbender We will have a final look at this, then we should merge it.

Issue #2098, opened 2021-11-18
GitHub new default Nodejs 16 will break deploy starting Nov 29, 2021


MikeMcC399:
## Where to find the issue

[deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml)

## Describe the issue

According to the announcements in https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
"[all OSs] Default Nodejs version will be set to 16 on November, 29"

This will break the deployment of cwa-website unless measures are taken to lock the deployment to Nodejs 14.

> ### Breaking changes
> We are going to set Nodejs 16 as the default version across all VM images.
> 
> ### Target date
> Image deployment is starting on November 29 and will take 3-4 days.
> 
> ### The motivation for the changes
> Node.js 16 has moved to Active LTS status and is ready for general use. We would like to provide images with the latest stable updates as default ones.
> 
> ### Possible impact
> If your builds depend on the default Nodejs 14 or npm 6 they can be broken.
> 
> ### Virtual environments affected
> * [x]  Ubuntu 18.04
> * [x]  Ubuntu 20.04
> * [x]  macOS 10.15
> * [x]  macOS 11
> * [x]  Windows Server 2016
> * [x]  Windows Server 2019
> * [x]  Windows Server 2022
> 
> ### Mitigation ways
> The recommended way is to pin your workflow to the desired nodejs version using:
> 
> * https://github.com/actions/setup-node action for GH
> * https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/node-js?view=azure-devops for AzDO

## Current setup

[deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml) contains

```
runs-on: ubuntu-latest
    run: npm install
    run: npm run build
```

https://github.com/actions/virtual-environments defines `ubuntu-latest` as [Ubuntu 20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) currently (Nov 18, 2021)

- Node 14.18.1
- Npm 6.14.15

## Result of running on Nodejs 16

On Ubuntu 20.04.03
```
$ nvm use lts/gallium
Now using node v16.13.0 (npm v8.1.0)
```

The command
`npm install`
first updates package-lock.json
and after some deprecation messages it fails trying to install `node-sass`.

[...]

PR #2106, opened 2021-11-22
Prevent break the deployment of cwa-website

brifemu:
PR to solve issue #2098


MikeMcC399:
@brifemu 
Thank you for picking up this issue and using my suggestion from #2098!

Is there some way to check if node `lts/fermium` is actually being selected and used? I could not find this in the log files.

According to https://nodejs.org/en/about/releases/ `lts/fermium` is equivalent to node v14.18.1. See https://nodejs.org/download/release/latest-fermium/. 

According to https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md `Node 14.18.1` is the default version for `ubuntu-latest` at the moment in any case.

Maybe we just have to monitor and make sure everything still works next week (after the planned migration on Nov 29)?

svengabr (approved):

dsarkar (approved):

Issue #2114, opened 2021-11-22
Scrolling behavior is broken under Safari on iOS on FAQ page


Ein-Tim:
## Where to find the issue
DE: https://www.coronawarn.app/de/faq/
EN: https://www.coronawarn.app/en/faq/

## Describe the issue
The scrolling behavior of the FAQ page is really broken on the mobile Safari browser: 

https://user-images.githubusercontent.com/67682506/142935138-0fb8b47b-9a7e-426f-8e63-90080f7eed31.mp4

https://user-images.githubusercontent.com/67682506/142935185-66fae5c2-d6cc-40c4-bd85-2501291385a3.mp4

## Suggested change
Fix the issue ASAP.

## Additional information
I was not able to reproduce this under Android using the Chrome browser. I did not check under iOS with another browser. IHMO this is a high priority issue.

dsarkar:
@Ein-Tim Thanks, we will have a look at this.

dsarkar:
@Ein-Tim Thanks for reporting. PR https://github.com/corona-warn-app/cwa-website/pull/2116

Ein-Tim:
Thanks @dsarkar & @matriceria!

Confirm fixed! 

Closing.

PR #2116, opened 2021-11-23
Fixed scroll in IOS devices

matriceria:
We discovered that the issue #2114  was caused because of the event listener that we added in pr #2053. We added a new function so that the event listener only works in desktop devices.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -3,15 +3,27 @@ import throttle from 'lodash.throttle';
 import 'slick-carousel';
 
 window.jQuery = $;
+const deviceType = () => {
+    const ua = navigator.userAgent;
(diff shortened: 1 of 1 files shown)


dsarkar (approved):

src/assets/js/app.js
```diff
@@ -3,15 +3,27 @@ import throttle from 'lodash.throttle';
...
-		if(hash) {
-            const contPos = $(hash).offset().top;;
-            $(window).scrollTop(contPos);
-        }
-	});
+    if(deviceType() === "desktop"){
+        $(window).on('resize', function(){
+            const { hash } = window.location;
+            if(hash) {
+                const contPos = $(hash).offset().top;;
```
svengabr, line 22: Please remove the double semicolon

dsarkar: fixed

Issue #2084, opened 2021-11-15
Misleading description of number of "Empfangene Warnungen"


OlympianRevolution:
## Where to find the issue
https://www.coronawarn.app/en/analysis/
Plot Empfangene Warnungen

## Describe the issue
Neither the plot subtitle nor the info text mention that this is a lower bound. It is a lower bound since the number of warnings given is only known for participants of PPA.(about half of users)

## Suggested change
I would change the title from:
"Empfangene Warnungen"
To:
"Mindestanzahl empfangener Warnungen"
AND
Change the text "Gesamt (Anzahl bis gestern" to "Gesamt (Mindestsumme aus Daten der freiwilligen Datenspende)" or similar

AND OR

Change the info text from:
"Empfangene Warnungen
Anzahl roter und grüner Risikowarnungen sowie deren Summe in der Corona-Warn-App."
To:
"Mindestanzahl Empfangener Warnungen
Anzahl roter und grüner Risikowarnungen bei Teilnehmern der PPA sowie deren Summe in der Corona-Warn-App." or similar

MikeMcC399:
@OlympianRevolution

Thanks for your suggestion!

I agree that it should be clearer for "Empfangene Warnungen" what this really means.

I suggest changing the title displayed on the graph to
"Empfangene Warnungen (nur Datenspende)"

and in the ℹ️ Info text explain that these warnings are only collected from users who have enabled in the app to share data (German: Datenspende), currently approximately half of users. 

Casual readers may not understand the significance of the text:

**Datenquellen**
Privacy-Preserving-Analytics (PPA)

---
The science blog https://www.coronawarn.app/de/science/2021-10-15-science-blog-4/#2-das-wichtigste-in-k%C3%BCrze-zahlen-daten-fakten says:

- Täglich spenden etwa 12 Mio. Nutzende ihre Daten (insgesamt gibt es 9.638.099.766 Datensätze).
- Der Anteil der Datenspendenden liegt bei etwa 49%.

which is completely in agreement with your statement about the data coming from about half of users. This puts the data into perspective.

Ein-Tim:
#2169 will fix this.

MikeMcC399:
@OlympianRevolution 

The texts you referred to have now been changed:

- https://www.coronawarn.app/en/analysis/ "Warnings received by data donors"
- https://www.coronawarn.app/de/analysis/ "Von Datenspendenden empfangene Warnungen"

[...]

PR #2128, opened 2021-11-26
Update analyse_en.json

HelTob:
Fixed infobox "Received warnings".


MikeMcC399:
@HelTob 
Would you also take a look at issue https://github.com/corona-warn-app/cwa-website/issues/2084 to make it clearer that the count of Received Warnings is underestimated because it only counts users who have subscribed to Data Donation?

MikeMcC399:
@HelTob 

> We're on issue #2084 and will change the description in the near future.

I'm happy to hear that! Thank you.

PR #2131, opened 2021-11-26
Update backup for analyseData.json

brifemu:
In this PR due to today's technical issue, I have made a change on the server deploy, updating the backup file used to start the project in localhost without internet connection.

The file had been added to the .gitignore file because for testing is not necessary to keep the file updated, just is important for the deployed site.

The file name was changed because IDK why, with the last name .gitignore ignored ignore the file.


dsarkar (approved):

MikeMcC399 (changes requested):
@brifemu
When I execute
`npm run build`
in this branch, it creates an untracked file
`src/data/analyse_data_no_internet.json`

so I think something needs to change in this PR.

`npm run build` should not cause any changes that git status would report about.

brifemu:
> @brifemu When I execute `npm run build` in this branch, it creates an untracked file `src/data/analyse_data_no_internet.json`
> 
> so I think something needs to change in this PR.
> 
> `npm run build` should not cause any changes that `git status` would report about.

You are right. I forgot to change the path, sorry. Already fixed.

MikeMcC399:
@brifemu 
There is no problem with `git status` now, after your latest change. 🙂

svengabr (approved):

Issue #2132, opened 2021-11-26
check_links fails on testbuchen.de


MikeMcC399:
## Where to find the issue

[check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_links.js)

## Describe the issue

[check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_links.js) produces a 403 (http forbidden) error when checking the link 

https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999

When opened manually, this link shows no error.

## Steps to reproduce

`npm run test:open`
select `check_links.js`

When the test fails, then view the file `cypress/logs/broken_links_result.txt` which contains:

```
==================== Broken links ====================

[RESPONSE 403] https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999 on 'http://localhost:8000/de/blog/2021-05-11-how-to-rapid-test-integration' 
[RESPONSE 403] https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999 on 'http://localhost:8000/de/blog/2021-05-02-corona-warn-app-version-2-1' 
[RESPONSE 403] https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999 on 'http://localhost:8000/en/blog/2021-05-11-how-to-rapid-test-integration' 
[RESPONSE 403] https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999 on 'http://localhost:8000/en/blog/2021-05-02-corona-warn-app-version-2-1' 
```

I don't know what is causing this error and this is a new problem.

MikeMcC399:
It seems that the site testbuchen.de has added network security to stop it being automatically accessed.

![Error 1020](https://user-images.githubusercontent.com/66998419/143623929-b05ac688-a0a5-4238-b349-214cb94a0415.png)

So we need some way in check_links.js to make an exception for websites like testbuchen.de which don't allow link checking automatically.

This is similar to a problem we had with markdown-check-links on the cwa-documentation site (see https://github.com/corona-warn-app/cwa-documentation/issues/542#issuecomment-795571682).

I solved that by adding:
```
<!-- markdown-link-check-disable --> disables markdown link check. 
<!-- markdown-link-check-enable --> reenables markdown link check.
```
[...]

Issue #952, opened 2021-03-06
FAQ: Firewall setting for Event-driven User Surveys (EDUS)


MikeMcC399:
## Where to find the issue

**[Google/Android]: Firewall/Router configuration: CAUSE: 4000, error during web request, HTTP status 901**
https://www.coronawarn.app/en/faq/#access-list

**[Google/Android]: Firewall/Router Konfiguration: URSACHE: 4000, error during web request, HTTP status 901**
https://www.coronawarn.app/de/faq/#access-list

## Describe the issue

The domains listed in the Firewall advice (see links above) are:

- *.coronawarn.app
- *.t-online.de

The domain *.rki.de is not listed and it seems this is necessary for the Event-driven User Surveys (EDUS) according to the verbose logging I'm seeing in CWA Android 1.13.2 Logcat, which shows a configuration parameter
`surveyOnHighRiskUrl=https://befragungen.rki.de/CWABasisbefragung` 

The parameter is also to be seen in the server configuration file [application.yaml](https://github.com/corona-warn-app/cwa-server/blob/main/services/distribution/src/main/resources/application.yaml).

## Suggested change

Please check that access to `https://befragungen.rki.de` is necessary.

Add 
`*.rki.de`
to the list of domains in
- https://www.coronawarn.app/en/faq/#access-list
- https://www.coronawarn.app/de/faq/#access-list

Also, test what happens if a firewall blocks access to this URL. If the error 
`CAUSE: 4000, error during web request, HTTP status 901`
is not the outcome, then it might be necessary to reword the document including the title.

---
Internal Tracking ID: [EXPOSUREAPP-1906](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-1906)

Ein-Tim:
@MikeMcC399 

The site https://befragungen.rki.de/CWABasisbefragung checks your browser with https://www.howsmyssl.com/ (see https://befragungen.rki.de/browsercheck), so probably this site would also have to be whitelisted, or is this process completely independent from the app?

Just btw, is the survey only in German?

MikeMcC399:
@Ein-Tim 
Good point about https://www.howsmyssl.com/ ! I didn't notice this. I can see this URL also being used when I look at the network traffic. Perhaps the prerequisites for using the survey need to be a new separate FAQ?

> Just btw, is the survey only in German?
[...]

PR #2136, opened 2021-11-29
Create URL allow list  for check_links test

brifemu:
Due to issue #2132 I have create a allow list to avoid errors like the one like the mentioned issue. For add more URLs just have to insert it into the new arrays.


Ein-Tim:
I did not test whether the changes work, however, I want to add that the word "whitelist" should not be used anymore, see https://github.com/corona-warn-app/cwa-website/issues/952#issuecomment-791924528 and the related issue https://github.com/corona-warn-app/cwa-documentation/issues/580.

I suggest you change it to "allowlist" or something similar that fits here.

MikeMcC399:
@brifemu 

I executed 
`npm run test:open` then 
`check_links.js`
which showed 31 passed tests and 0 failed tests.

Thank you for solving this issue!

dsarkar:
@MikeMcC399 Thanks for testing and feedback.
@Ein-Tim Thanks for your comment.
@brifemu See comment above and change the wording also in the code to "allowlist"

brifemu:
Thanks, I've already learned it for next time. I'm sorry.

dsarkar (approved):

PR #2153, opened 2021-12-03
Delete large v4_Teaser_CWA_Testkampagne.mp4

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/2152 "Large (95 MB) unused mp4 file" by deleting the unused MP4 video file [science/2021-06-15-science-blog-1/v4_Teaser_CWA_Testkampagne.mp4](https://github.com/corona-warn-app/cwa-website/blob/master/science/2021-06-15-science-blog-1/v4_Teaser_CWA_Testkampagne.mp4).

## Verification

Ensure that the Cypress test suite runs without error after deleting this file.

Issue #1349, opened 2021-06-10
Example of EU Digital COVID vaccination certificate half upside down


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/faq/#vac_cert_invalid
https://www.coronawarn.app/de/faq/#vac_cert_invalid

link to image https://user-images.githubusercontent.com/6367738/121354638-59aba880-c92f-11eb-98bb-9dcbc42d6742.jpg

## Describe the issue

The top half of the image used in the FAQ is upside down and not easily readable by non-acrobats.

![121354638-59aba880-c92f-11eb-98bb-9dcbc42d6742](https://user-images.githubusercontent.com/66998419/121585358-c8255f00-ca32-11eb-85bd-712bd771e580.jpg)

## Suggested change

Provide an image such as the one attached where the text is all readable from one viewing position without having to print out the graphic or have the user stand on their head.

![EU Digital Covid Vaccination Certificate Demo](https://user-images.githubusercontent.com/66998419/121585483-f014c280-ca32-11eb-83b1-38ef4a74719b.jpg)

MikeMcC399:
@dsarkar 

Is this what you intended?

![image](https://user-images.githubusercontent.com/66998419/121649196-9e9f1e80-ca98-11eb-8b30-10fe5d122806.png)

MikeMcC399:
Maybe this was a refresh issue? Anyway it is working now. So ignore previous comment. 🙂

dsarkar:
@MikeMcC399 Thanks! File was missing.

PR #2156, opened 2021-12-05
Replace GitHub picture with local one

Ein-Tim:
This PR replaces a picture which was shown using the GitHub usercontent pictures with a local one.
It also removes src/assets/img/faq/EU_digital_vaccination_certificate_2.jpg, which is a duplicate of src/assets/img/faq/EU_digital_vaccination_certificate.jpg from the repository.


MikeMcC399:
@Ein-Tim 
On https://www.coronawarn.app/en/faq/#vac_cert_invalid clicking on the image of the certificate shows a larger version which is more readable. Does your change preserve this behavior?

Ein-Tim:
@MikeMcC399 

HI Mike, thanks for your question! No, this PR changes the behavior. One would now have to do a right click on the picture and then select "Open picture in new tab". 

I can try changing this, let me see.

Ein-Tim:
@MikeMcC399 Restored the behavior of clicking on the picture with https://github.com/corona-warn-app/cwa-website/pull/2156/commits/65d4214aa91b3171ad0a07277d597ef252a7aff6. I made the picture a link which links to itself, thus, if clicked, the picture can be viewed in a larger version. 

Hope this is what you meant?

MikeMcC399:
@Ein-Tim

Perhaps you didn't notice, I manipulated the images on the current page https://www.coronawarn.app/en/faq/#vac_cert_invalid so that all the text is readable, not upside down. @dsarkar took this and posted it.

See https://github.com/corona-warn-app/cwa-website/pull/1351.

The following image comes from my account (`EU Digital Covid Vaccination Certificate Demo`)  and from the issue https://github.com/corona-warn-app/cwa-website/issues/1349. I thought that @dsarkar would copy and rename it when he used it, but he just linked to my account instead:

EU Digital Covid Vaccination Certificate Demo

If you want to clean this up by making the image independent of my GitHub account, you could copy the jpg image to the web `img` directory and link to it, leaving everything else the same.

A cleaner solution would be to separate the top and bottom half of the form and show each half in a readable orientation.

Another option would be not to change anything, since it is not broken.

Ein-Tim:
@MikeMcC399 

> Perhaps you didn't notice, I manipulated the images on the current page https://www.coronawarn.app/en/faq/#vac_cert_invalid so that all the text is readable, not upside down.

I can change this back, although I'm not too happy as the actual certificate which is received by the user is upside down too.

> A cleaner solution would be to separate the top and bottom half of the form and show each half in a readable orientation.

Yes, but this is a separate task. This PR aims to get remove the dependency of GitHub here.

> Another option would be not to change anything, since it is not broken.

Currently not, no, but 
1. There was an issue with this picture for a short amount of time during the major GitHub outage last months.
2. If GitHub has an outage, I as a user, would not expect this to influence coronawarn.app. This is why I propose this change.

MikeMcC399 (changes requested):
Change suggested to use `EU_digital_vaccination_certificate_2.jpg` only.

Ein-Tim:
@MikeMcC399 

> If you use the original EU_digital_vaccination_certificate_2.jpg instead of changing it to src/assets/img/faq/EU_digital_vaccination_certificate.jpg you will keep the original intention of displaying a readable certificate.

Okay, yes, I can do this.

> You say that the files are identical, but this is not true.

You're right, sorry, due to the similar/same names of the files I thought they were the same. I will restore this. 

> You can take out rel='noopener' from the hyperlink because this is only needed on external links to _blank targets. You are changing the hyperlink to an internal link, therefore there is no longer a need. (It is suggested to avoid a security link in older browsers when accessing external sites.)

Will do so.

Ein-Tim:
Sorry for merging master again, this was not my intention. 😕

@MikeMcC399 I implemented your suggested changes in [`aa732fb ](https://github.com/corona-warn-app/cwa-website/pull/2156/commits/aa732fb164e05164b4d34a05bb2fc9fb8684d9dd). Please take another look.

MikeMcC399 (changes requested):
@Ein-Tim 

It seems that you have deleted the original image source file and renamed the used image file to have the same name as the original image source file used.

I would advise not to change the contents or names of the two images files.

- the original image file might be needed again, so I would want to see that file kept and not deleted. Although I made a private copy of the file in June 2021 when I rotated the text, it should probably continue to be available to other people independently of my private copy.
- renaming the altered image file is confusing for the git / GitHub history and it introduces an change unnecessary to achieving the goal of removing the dependency on https://user-images.githubusercontent.com

Ein-Tim:
> I would advise not to change the contents or names of the two images files.

Why? If you'd advise not do so because it's not the subject of this PR, ok, then I will revert the changed to the files and create a new PR, only for these changes.

> the original image file might be needed again, so I would want to see that file kept and not deleted. 

So you think unused material should be kept in the source code? I think unused images should be deleted, they can be added back when needed. You also recently deleted a video in PR #2153. I don't see any reason for treating this image in a different way then the video deleted in #2153! 

> renaming the altered image file is confusing for the git / GitHub history and it introduces an change unnecessary to achieving the goal of removing the dependency on https://user-images.githubusercontent.com

Ok, yes. So your problem is that these changed are not the direct subject of this PR? If this is the problem, I can revert the changes and create a new PR. But as this is all somewhat related, it would save work & commits.

MikeMcC399:
@Ein-Tim 
Thank you for considering my comments.

The reason to delete the video was that it was extremely large (95MB) and it exceeded the GitHub maximum recommended size.

Ein-Tim:
@MikeMcC399 

And don't get me wrong: I'm happy that this video was deleted. Still I think unused material should not be kept in the repo, as it also consumes space, etc.

So please, @dsarkar or anybody else from the Open-Source Team, decide whether an unused picture should stay in the codebase or if it should be removed.

Thank you!

Ein-Tim:
Hey @dsarkar, could you take a look here? Thank you!

dsarkar:
@Ein-Tim @MikeMcC399 
- let's keep the original file for the moment, although it is not being used.
- lets's not change the filenames.

Thanks!

Ein-Tim:
Please, take another look now, @MikeMcC399. Thanks!

MikeMcC399:
@Ein-Tim 
> Please, take another look now, @MikeMcC399. Thanks!

You'll need to ask @dsarkar, since he wrote:
> lets's not change the filenames.

and you have changed a filename from:

`src/assets/img/faq/EU_digital_vaccination_certificate_2.jpg` to
`src/assets/img/faq/EU_digital_vaccination_certificate-2.jpg`

Ein-Tim:
Thanks Mike, fixed in [`0be26d9` (#2156)](https://github.com/corona-warn-app/cwa-website/pull/2156/commits/0be26d9074948903ffd7a1d85c108eb548e28095).

MikeMcC399 (approved):
The PR now achieves the goal of removing the dependency in [#vac_cert_invalid](https://www.coronawarn.app/en/faq/#vac_cert_invalid) on image content from https://user-images.githubusercontent.com without introducing other side-effects.

dsarkar:
@Ein-Tim @MikeMcC399 Thanks!

Issue #2146, opened 2021-12-02
build and deploy workflows use different node versions


MikeMcC399:
## Where to find the issue

[build.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) workflow in [cwa-website](https://github.com/corona-warn-app/cwa-website).

## Describe the issue

The node versions used by [build.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) (14.7.0) and [deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml) (14.18.1) are different which may cause unexplained differences in results the workflows produce.

This issue replaces the discussion started in https://github.com/corona-warn-app/cwa-website/issues/2142.

### Details

The [.github/workflows/build.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) workflow includes the line:

`container: cypress/browsers:node14.7.0-chrome84`

which provides the environment consisting of node 14.7.0, npm 6.14.7 and chrome 84.

The [.github/workflows/deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml) is however now tied to the official Long Term Support node version `lts/fermium`. In practice GitHub currently sets this equivalent to node 14.18.1 and npm 6.14.15.

(See PR https://github.com/corona-warn-app/cwa-website/pull/2106 which changed [deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml), https://github.com/nodejs/Release#release-schedule which defines all node releases and https://nodejs.org/download/release/latest-fermium/ which links to the latest LTS fermium release. Node 14.18.2 was released on Nov 30, 2021 as latest-fermium release.)

## Discussion

The available Cypress docker images are listed on https://github.com/cypress-io/cypress-docker-images.

1. There is currently no Cypress docker brower image in https://github.com/cypress-io/cypress-docker-images/tree/master/browsers which uses node 14.18.2 (lts fermium version). The latest is [node14.17.0-chrome91-ff89](https://github.com/cypress-io/cypress-docker-images/tree/master/browsers/node14.17.0-chrome91-ff89).

[...]

PR #2157, opened 2021-12-05
Harmonize build workflow to match deploy workflow

MikeMcC399:
This PR addresses issue https://github.com/corona-warn-app/cwa-website/issues/2146 and harmonizes the selection of Nodejs version v14 to run in the [build](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) workflow to match the [deploy](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml) workflow of [cwa-website](https://github.com/corona-warn-app/cwa-website).

## Node LTS Fermium (v14)

The Nodejs version is selected to be v14 according to the Long Term Support version alias `lts/fermium` (see [Nodejs release schedule](https://nodejs.org/en/about/releases/)) through the following section in [.github/workflows/build.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml):

```
    - name: Setup Node.js environment (v14)
      uses: actions/setup-node@v2
      with:
        node-version: 'lts/fermium'
```

`actions/setup-node` uses the cached version of `lts/fermium` defined in [Ubuntu 20.04 README.md: nodejs](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#nodejs) which is currently 14.18.2.

### GitHub cache alignment to Node.js

The GitHub virtual environment may show the same patch version or an older patch version of `lts/fermium` compared to the definition from [Nodejs - latest-fermium](https://nodejs.org/download/release/latest-fermium/). Right now they are identical. To override the version defined in GitHub and use the latest from https://nodejs.org, it would be necessary to add 
`check-latest: true` under 
`node-version: 'lts/fermium'`.  

I think it is better to align to the stable version cached by GitHub's virtual environment definitions, so I didn't add these extra statements.

## Use of Cypress Docker

The previous use of a Cypress Docker image through
`container: cypress/browsers:node14.7.0-chrome84` in the [build](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) workflow, which tied the build environment to an older minor version of Nodejs (14.7.0), is removed. More than one year ago the Cypress test suite was run as part of the build workflow. That was found to be problematic and so the execution of the test suite was removed, leaving the unnecessary Cypress Docker image behind. This is now corrected.

## Comments and labels

Comments in [.github/workflows/build.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) are tidied up and corrected.

Additionally the key `name: Setup Node.js environment (v14)` is set identically in [.github/workflows/deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml) for consistent logging in the two affected workflows. The incorrect comment `# This workflow contains a single job called "build"` has been corrected. The job is called `"prod"`.


MikeMcC399:
@dsarkar / @svengabr 

Although this PR is not essential, I believe it makes the workflows more maintainable and so it would be worthwhile to implement.

dsarkar:
@MikeMcC399 Thanks. We will check it!

svengabr (approved):
I have tested your changes locally. Everything ran smoothly and looking good. Approval from my side. Thanks for your effort @MikeMcC399 !

dsarkar (approved):

PR #2159, opened 2021-12-06
Fix white space - Update faq_de.json

dsarkar:
image


MikeMcC399:
@dsarkar 
You might like to take a look at #2163 regarding a possible change to the related partials file which might alleviate this problem and allow `&nbsp;` to be used in FAQ title text.

Issue #2163, opened 2021-12-07
nbsp HTML entity cannot be put into FAQ title


MikeMcC399:
## Where to find the issue

cwa-website FAQ titles

## Describe the issue

If the HTML entity `&nbsp;` is put into the title of an FAQ article to insert a non-breaking space it appears literally as `&nbsp;` instead of being interpreted. I had noticed this in the past and recently it had to be corrected by @dsarkar in PR https://github.com/corona-warn-app/cwa-website/pull/2159.

## Steps to reproduce

In [src/data/faq.json](https://github.com/corona-warn-app/cwa-website/blob/master/src/data/faq.json) add a new FAQ article:

```
                    {
                        "title": "Test&nbsp;title",
                        "anchor": "test_title",
                        "active": true,
                        "textblock": [
                            "Test&nbsp;text"
                        ]
                    }
```

Execute `npm run build`

View `public\en\faq\index.html` in web browser and in source code

```
              <div id="test_title-div">
                <dt class="accordion-header active">
                  <h3 class="accordion-item-title" id="test_title">Test&amp;nbsp;title</h3>
                  <span class="accordion-item-icon"></span>
                </dt>
                <dd class="accordion-body">
                  <div class="accordion-item-content">
                    <p>Test&nbsp;text.</p>
                    <p><a href="#test_title" class="faq-anchor">Link to this answer</a></p>
                    <p><a href="#top">To the top</a></p>
                  </div>
                </dd>
              </div>
```

In the title the `&nbsp;` is displayed literally, whereas in the text block it is used as intended and displays a non-breaking space.

![FAQ Test title](https://user-images.githubusercontent.com/66998419/145063727-7cd68785-f3a0-4da3-812a-8707d9cf8cb0.jpg)

## Investigation

The partials [src/partials/page-faq.html](https://github.com/corona-warn-app/cwa-website/blob/master/src/partials/page-faq.html) page, used by Panini to covert the faq.json file to HTML, deals with the title source link differently to the textblock.

https://github.com/corona-warn-app/cwa-website/blob/7a12ee369d8d015579afda5726cc6e40b2ab9e19/src/partials/page-faq.html#L37
https://github.com/corona-warn-app/cwa-website/blob/7a12ee369d8d015579afda5726cc6e40b2ab9e19/src/partials/page-faq.html#L44

Whereas the title uses double curly brackets, the textblock uses triple curly brackets.

[...]

PR #2165, opened 2021-12-07
Prevent escaping in FAQ title

MikeMcC399:
This PR implements the change suggested in issue https://github.com/corona-warn-app/cwa-website/issues/2163 "nbsp HTML entity cannot be put into FAQ title".

It changes the [src/partials/page-faq.html](https://github.com/corona-warn-app/cwa-website/blob/master/src/partials/page-faq.html) partials page which processes FAQ json input through Panini, so that the contents of the "title" field are not escaped.

This means that HTML entities such as `&nbsp;` are passed through one-to-one into the HTML. Before this change the `&` character was being escaped in title text with the effect that `&nbsp;` appeared literally in the title instead of being used as a non-breaking space.

Text in the "textblock" field was already set up not to be escaped, so this change brings the two fields - "title" and "textblock" into alignment with each other.

### Reference

https://handlebarsjs.com/guide/#html-escaping

## Verification

1. Run the Cypress test suite
2. Revert commit https://github.com/corona-warn-app/cwa-website/commit/7a12ee369d8d015579afda5726cc6e40b2ab9e19 from PR https://github.com/corona-warn-app/cwa-website/pull/2159 and confirm that:
    - `&nbsp;` is  **not** displayed in the title of https://www.coronawarn.app/de/faq/#vac_cert_multiple_scan_possible
    - when the browser width is changed, the text `z. B.` is never split on two lines and always remains together

### DIFF
--- a/src/partials/page-faq.html
+++ b/src/partials/page-faq.html
@@ -34,7 +34,7 @@
       {{#each accordion}}
         <div id="{{anchor}}-div">
           <dt class="accordion-header{{#if active}} active{{/if}}">
-            <h3 class="accordion-item-title"{{#if anchor}} id="{{anchor}}"{{/if}}>{{title}}</h3>
+            <h3 class="accordion-item-title"{{#if anchor}} id="{{anchor}}"{{/if}}>{{{title}}}</h3>
             <span class="accordion-item-icon"></span>
           </dt>
           <dd class="accordion-body">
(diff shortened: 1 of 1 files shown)


MikeMcC399:
The VS Code editor automatically trims trailing whitespace according to the GitHub settings for this repository (see [.vscode/settings.json](https://github.com/corona-warn-app/cwa-website/blob/master/.vscode/settings.json) so I added this commit as well.

dsarkar (approved):

dsarkar:
Good morning @MikeMcC399, thank you very much for solving this issue and this PR.

MikeMcC399:
@dsarkar 
Good morning to you and the Open Source Team!
> thank you very much for solving this issue and this PR.

My pleasure!

Issue #2171, opened 2021-12-09
Analysis page: Issue with title


dsarkar:
**Where:**
Second graph on:
[0] https://www.coronawarn.app/de/analysis/
[1] https://www.coronawarn.app/en/analysis/

**Issue**
Short moment one sees new title, then the old one.

MikeMcC399:
@dsarkar 
Good morning!
Could you explain a bit more about this issue as I don't understand it? I couldn't see any problem when I opened either of the pages.

Is it perhaps happening only with a particular browser?

HelTob:
@MikeMcC399 

Good morning,

the title of the quadrant "Registrierte Tests" (DE) or "Registered test" (EN) has been changed to "Übermittelte und registrierte Testergebnisse" (DE) or "Transmitted and registered test results" (EN) but the old title is still shown.

If you hit F5 on the analysis page (DE and EN) the new title is shown for a short moment but then changes on an instant to the old one.

Tested with Mozilla Firefox 91.3.0 and Microsoft Edge 96.0.1054.43.

Best regards
Tobias

MikeMcC399:
@HelTob 
Now I see it. Thank you for taking the trouble to explain.

dsarkar:
![image](https://user-images.githubusercontent.com/1409741/145384064-ef344593-af69-48e0-a01c-ca379bb1b56b.png)

![image](https://user-images.githubusercontent.com/1409741/145384083-db4f5fab-6c60-4717-9abb-30cafdb01f6e.png)

brifemu:
FYI: https://github.com/corona-warn-app/cwa-website/pull/2172

dsarkar:
@brifemu Thanks.

PR #2172, opened 2021-12-09
Fix analyse title updated

brifemu:
Hello, due to issue #2171 I have created this pull request. The problem was that the title was first loaded from translations section from the json at line 140 and then is replaced from the board's title, at line 81.


Ein-Tim:
This PR seems to include an unrelated commit: https://github.com/corona-warn-app/cwa-website/commit/92e54dabc495aff0f3b77a80bdf487c97d2266eb

brifemu:
Close for git issue. Solving it.

PR #2173, opened 2021-12-09
Fix analyse page title

brifemu:
Hello, due to issue #2171 I have created this pull request. The problem was that the title was first loaded from translations section from the json at line 140 and then is replaced from the board's title, at line 81.


dsarkar (approved):

dsarkar:
@HelTob FYI

Issue #2154, opened 2021-12-03
English Science blog 1 graphs reverted to German format


MikeMcC399:
## Where to find the issue

On [Science Blog 1 - EN](https://www.coronawarn.app/en/science/2021-06-15-science-blog-1/)
- Fig. 1: Tests registered with the CWA and their results.
- Fig. 2: Test results shared via the CWA (TAN and QR code).
- Fig. 3: The number and risk level (Transmission Risk Level, TRL) of the daily keys shared via the CWA.
- Fig. 4: Users who provided or received warnings via the CWA.

## Describe the issue

The figures use German-language dates in the English-language science blog.

- dot after the day of the month
- name of month in German

for example "5. Okt" which should be "5 Oct".

This is a follow-on from the closed issue:
- https://github.com/corona-warn-app/cwa-website/issues/1980

which was solved by @brifemu at the time. 

## Suggested change

It looks like one of the other later changes to the graphs has overwritten earlier changes. The locale has gone back to "de". I don't know however what influences the dot after the day of the month though. It is almost impossible in git to see the changes because the whole graph data is on one extremely long line.

MikeMcC399:
@brifemu / @matriceria 

I think this is an issue for one of you.

dsarkar:
@MikeMcC399 Thanks, we will check this.

brifemu:
FYI: https://github.com/corona-warn-app/cwa-website/pull/2174

MikeMcC399:
@brifemu 
#2174 fixes the locale but not the date format.

brifemu:
@MikeMcC399 
Fixed here https://github.com/corona-warn-app/cwa-website/pull/2176, thanks for your feedback as always!

MikeMcC399:
@brifemu 
Thank you.

All is now fixed.

PR #2174, opened 2021-12-09
Fix date locale for figs

brifemu:
Due to issue #2154 I have edited the locales to print the correct language.


MikeMcC399:
@brifemu 
Can you also remove the dot in the dates?
For instance on 
https://www.coronawarn.app/en/science/2021-06-15-science-blog-1/ and 
Fig. 1: Tests registered with the CWA and their results.
the first date on the left is
"29. June"
which should be
"29 June"
without the dot.

dsarkar (approved):

dsarkar:
@MikeMcC399 So sorry... merged it already. Ping @brifemu

PR #2193, opened 2021-12-15
Updated the FAQ entry #vac_cert_red_warning

thorbenkuro:
Updated the FAQ entry according to issue #2192


MikeMcC399:
@dsarkar 
I assume that you meant to apply the 2.15 label, not the 2.5 label.

Issue #2239, opened 2021-12-22
browser-sync does not refresh browser


MikeMcC399:
## Where to find the issue

cwa-website

## Describe the issue

This is a testing issue only. It does not affect the production website.

When making source changes to cwa-website, `browser-sync` announces "Reloading Browsers", however it does not refresh the browser contents as it should.

## Steps to reproduce

Execute 

```
npm start
```

Terminal shows
```
[Browsersync] Serving files from: ./public
Finished 'server' after xx ms
Starting 'watch'...
Finished 'watch' after xx ms
Finished 'start-server' after xxx ms
Finished 'default' after xx s
```

Web browser displays http://localhost:8000/en/

Open `src/data/index.json` in an editor.
Make a change, such as changing the title in line 5 and save the change.

```
Starting 'resetPages'...
Finished 'resetPages' after xx ms
Finished 'pages' after x.x s
[Browsersync] Reloading Browsers...
```

Note that browser contents have not changed.
Press `F5` (refresh) in browser
Now the changes which were made are displayed.

## Expectation

When browser-sync displays "Reloading Browsers..." then the contents of the browser should be refreshed.

This is documented in [README - Start server](https://github.com/corona-warn-app/cwa-website/blob/master/README.md#start-server) as:

"The web server uses public as its base directory. It also watches for file changes and refreshes the browser contents if necessary."

It is also a documented function of [Browsersync](https://browsersync.io/).

Synchronized scrolling between multiple browsers also does not work for cwa-website as documented in the Browsersync documentation.

## Workaround

To refresh the page after making a change to the source, press `F5` (refresh) in the browser window.

If there is no solution to this issue, then this hint should be added to the [README - Start server](https://github.com/corona-warn-app/cwa-website/blob/master/README.md#start-server) section.

MikeMcC399:
I am trying to find out the reason for this not working, but I don't know if I will get to the bottom of it.

It works with a simple test page, so there is something about the complexity of the cwa-website which is causing a problem.

MikeMcC399:
PR https://github.com/corona-warn-app/cwa-website/pull/1784 "Analysis page" with commit https://github.com/corona-warn-app/cwa-website/commit/7e7d2217f04eba716e0981b9508552c5a8187ae0 merged on Oct 1, 2021 introduced a regression concerning the automatic reload of the browser window by `Browsersync` when web site source files were changed.

[...]

PR #2247, opened 2021-12-27
Restore [Browsersync] Reloading Browsers functionality

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/2239 "browser-sync does not refresh browser".

It adds back `'unsafe-inline' 'unsafe-eval'` in the Content Security Policy `default-src` fetch directive in places where it was removed.

The CSP of https://www.coronawarn.app/en/analysis/ already includes the two directive values `'unsafe-inline' 'unsafe-eval'`, so there is already a precedent for allowing these values on the production website.

## Verification

Execute
```
npm ci
npm start
```

When the web browser starts, go to http://localhost:8000/en/faq/.

Edit `src\data\faq.json` and save the change.
For instance make a change in the title on line 4.

Wait for `[Browsersync] Reloading Browsers...` to appear in the terminal window.
Notice that the contents of the browser has been automatically updated and includes the change made above.


browkin (approved):
@MikeMcC399 Tested locally. Thank you for the fix.

mtb77:
Hi @MikeMcC399 - thanks for the suggestion. I think it's a good idea which makes sense especially for text changes, so I would like to approve this.
However, this was added to test and see new upcoming issues with JS and the server-based CSP. Would you be so kind and also add a description for testers how to add the policy back again to test according to the server policies?
Thanks so much!

Sascha

MikeMcC399:
Hi @mtb77 

Thank you for your inputs and your comments that you would like to approve.

Are the server policies which you refer to published anywhere? I don't think there is anything documented in the README here, so I don't know how this is supposed to work.

Are the CSP settings produced by the `npm run build` command used by the production nginx server https://www.coronawarn.app or are they ignored or overwritten? I see for the science blogs there is an explicit `--remove-header=Content-Security-Policy` so that makes it look like they might be transferred by default from the build result. Have any or all of the Content Security Policy settings on the production server been defined through manual AWS S3 console commands?

I think I would need to understand the above first before I could get close to adding a description for testers to test in a stricter environment. An alternative would perhaps be to add a parameter to the `npm run start-server` task to make it run with stricter CSPs.

---

The fallback for this PR would be not to implement it and simply document in the README that it is necessary to refresh the browser manually (F5) to get an update. In practice this means for multiple updates of text, you would be pressing Ctrl-S to save in an editor, then F5 to view, each time you wanted to view the results of a change. This is the way it is working right now.

I also found https://github.com/BrowserSync/browser-sync/issues/1395 which discusses using `nonce` to allow Browsersync to refresh, which could be an alternative to `'unsafe-inline' 'unsafe-eval'`. Unfortunately this is past my level of expertise, so if that were a viable alternative I would have to pass it on to somebody else to implement.

MikeMcC399:
@mtb77
If you could take a look at 
- https://github.com/corona-warn-app/cwa-website/issues/2258

and comment, that would also help me understand. This would be an example of a JS issue which might be hidden in a test environment if this PR were to be merged as it stands today.

I do not yet understand how the CSPs are set in the production environment though.

mtb77:
Hi @MikeMcC399  - i have approved this PR right now - we can discuss the changes for the README.md in the issue #2258  which you mentioned!

Thank you,

Sascha

MikeMcC399:
Hi Sascha (@mtb77)
Thanks for your approval and merge!

We can continue in #2258 as you suggest.

Issue #2251, opened 2021-12-27
Wrong use of narrow no-break space (U+202f) in FAQ (DE)


MikeMcC399:
## Where to find the issue

- `src\data\faq_de.json` displaying on
- https://www.coronawarn.app/de/faq/#val_service_no_valid_dcc "Warum erhalte ich die Meldung, dass kein geeignetes Zertifikat vorhanden ist?" 

## Describe the issue

The `src\data\faq_de.json` FAQ source file contains multiple instances where the Unicode character [U+202f](https://en.wikipedia.org/wiki/Whitespace_character#NNBSP) is used in incorrect places.

It is particularly noticeable in https://www.coronawarn.app/de/faq/#val_service_no_valid_dcc "Warum erhalte ich die Meldung, dass kein geeignetes Zertifikat vorhanden ist?" where it looks like "Sienur" is one word.

![val_service_no_valid_dcc](https://user-images.githubusercontent.com/66998419/147475772-96057037-7c41-4d90-b188-f3b7f0945d73.png)

## Suggested change

Review each use of narrow no-break space (`U+202f`) in `src\data\faq_de.json` and correct if necessary.

MikeMcC399:
PR #2252 submitted to resolve this issue.

dsarkar:
@MikeMcC399 Thanks for reporting.

PR #2252, opened 2021-12-27
Replace narrow nbsp formatting in FAQ DE where not needed

MikeMcC399:
This PR resolves https://github.com/corona-warn-app/cwa-website/issues/2251 "Wrong use of narrow no-break space (U+202f) in FAQ (DE)".

Almost all uses of the narrow no-break space (U+202f) character in `src\data\faq_de.json` look like they were inserted by mistake.

Except where a narrow no-break space (U+202f) character is used before a telephone number all other uses of narrow no-break space (U+202f) are replaced by a regular space character.

The following articles are affected:

- https://www.coronawarn.app/de/faq/#val_service_no_valid_dcc
- https://www.coronawarn.app/de/faq/#glossary_enf
- https://www.coronawarn.app/de/faq/#glossary_recovery_certificate


dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks for the fix!

PR #2261, opened 2021-12-29
FAQ Collapse

browkin:
- collapsed all articles in faq for visibility and speed


dsarkar:
@browkin 
Keep the "mini-indexes" open:
- https://www.coronawarn.app/en/faq/#check_in_index
- https://www.coronawarn.app/en/faq/#test_cert_index
- https://www.coronawarn.app/en/faq/#vac_cert_index
- https://www.coronawarn.app/en/faq/#vac_cert_booster

Ein-Tim (approved):
Works as intended! Thanks!

I think it would be nice if the titles of the FAQ entries would have a blue background, even if collapsed. 

At the moment it looks like this: 



It would be nice if the titles would have a blue background, like when the entries are opened:
Bildschirmfoto 2021-12-29 um 15 00 32

dsarkar:
@browkin See comment above, I guess have a look at the css.

browkin:
@Ein-Tim Thanks for the suggestion. Added!

Ein-Tim (approved):
LGTM now! Thanks for making the change with the background color!

MikeMcC399:
This is looking good now and I think this is a major improvement in how the FAQs are displayed.

I also checked it out on a mobile device and it looks good there too.

dsarkar (approved):

Issue #2240, opened 2021-12-23
EU Digital COVID Certificate press release changes validity and booster coding


MikeMcC399:
The press release [EU Digital COVID Certificate: Commission adopts binding acceptance period of nine months for vaccination certificates](https://ec.europa.eu/commission/presscorner/detail/en/ip_21_6837) from Dec 21, 2021 may affect two aspects documented in the FAQs:

- binding acceptance period of 270 days for travel
- new coding for booster vaccinations

## Booster changes

"In addition, today the Commission has also adapted the rules for the encoding of vaccination certificates. This is necessary to ensure that vaccination certificates showing completion of the primary series can always be distinguished from vaccination certificates issued following a booster dose.

Boosters will be recorded as follows:

- 3/3 for a booster dose following a primary 2-dose vaccination series.
- 2/1 for a booster dose following a single-dose vaccination or a one dose of a 2-dose vaccine administered to a recovered person"

---
Note that this is an EU announcement. The official German government position needs to be monitored and considered also.

---
EU Regulations are:

- [(EU) 2021/2288](https://eur-lex.europa.eu/legal-content/EN-DE/TXT/?uri=CELEX:32021R2288) (Dec 21, 2021) "It shall apply from 1 February 2022. / Sie gilt ab dem 1. Februar 2022."

- [(EU) 2021/2301 ](https://eur-lex.europa.eu/legal-content/EN-DE/TXT/?uri=CELEX:32021D2301) (Dec 21, 2021)

MikeMcC399:
The new coding for booster vaccinations is described in an FAQ update through PR https://github.com/corona-warn-app/cwa-website/pull/2264.

MikeMcC399:
The legal documents referred to in the press release [EU Digital COVID Certificate: Commission adopts binding acceptance period of nine months for vaccination certificates](https://ec.europa.eu/commission/presscorner/detail/en/ip_21_6837) from Dec 21, 2021  are still in draft status:

- [Commission Delegated Regulation on the validity of vaccination certificates issued under the EU Digital COVID Certificate framework](https://ec.europa.eu/info/files/commission-delegated-regulation-validity-vaccination-certificates-issued-under-eu-digital-covid-certificate-framework_en)

[...]

PR #2264, opened 2021-12-30
Added an paragraph about the commission adopted rules to FAQ #vac_booster_jj

browkin:
- Added to DE & EN

**Deutsch:**
comissionDE

**English:**
commissionEN


MikeMcC399 (changes requested):
@browkin 

I suggest that you write at the top:
"UPDATE December 30, 2021" in bold
like used on https://www.coronawarn.app/en/faq/#minimum_requirements
(avoid stating dates numerically in English because they are locale dependent).

Also to follow the example of https://www.coronawarn.app/de/faq/#minimum_requirements write
"AKTUELL 30. Dezember 2021" in bold

----
For the external links with `target='_blank'` add `rel='noopener noreferrer'` for security with older browsers.

MikeMcC399:
This is related to issue #2240.

MikeMcC399:
This is the best practice reference:
https://web.dev/external-anchors-use-rel-noopener/

It probably should be added to the website documentation so that all contributors are creating hyperlinks consistently.

MikeMcC399 (approved):
@browkin 
Thank you for taking account of my suggestions.
It looks fine now!

dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks for reviewing !

Issue #2287, opened 2022-01-05
Wrong "Daily count" on the figures "Test results - PCR tests" and "Test results - Rapid antigen tests" on the analysis page


Ein-Tim:
## Your Question

* Website URL Path: https://www.coronawarn.app/en/analysis/
* Line / Paragraph: "Test results" and then "Test results - PCR tests" and/or "Test results - Rapid antigen tests" 

What does "Daily count" mean? It is shown above the positive tests and I would guess that this means that on this date x persons were tested positive. But why isn't "Daily count" shown above "Invalid" and "Negative" too? This does not really make sense to me, as all the data shown there is the "Daily count", or? 

When I switch to the "Sorted by calendar week" view this makes even less sense to me. 

Here's a screenshot: 

![Bildschirmfoto 2022-01-05 um 23 37 36](https://user-images.githubusercontent.com/67682506/148299562-fcd780b8-9dfa-4cde-99d3-d828770ee40f.png)

MikeMcC399:
@Ein-Tim 

It's a formatting and labelling bug.

## Steps to reproduce

1. Open https://www.coronawarn.app/en/analysis/
2. View "Transmitted and registered test results" graph using default selections (Last 28 days, Daily)
3. Click on "PCR tests"
4. Hover with the mouse over the graph
5. Note the data formatting

The label "Daily count" is below a separating line and above the value and label for "Positive"

## Expected display

The "Daily count" label should not be displayed.
The Positive value should be directly below the Negative value without any separating line.

Ein-Tim:
@MikeMcC399 

I also thought that this might be a bug! Thanks for confirming. 

@dsarkar Please, take a look at this. Thanks.

MikeMcC399:
This is an area that @brifemu has worked on in the past.  He probably has a public holiday today though.

brifemu:
@Ein-Tim 
FYI https://github.com/corona-warn-app/cwa-website/pull/2290

MikeMcC399:
@Ein-Tim 

It's up to you if you want to change the description, but now the bug has been identified, a title such as:

Wrong "Daily count" label on Analysis / Test result page

would be closer to describing the real issue.

Ein-Tim:
@MikeMcC399 I changed the title, thanks.

I can confirm that this issue has been fixed, so closing this.

PR #2212, opened 2021-12-20
Create header and separator for '7 day moving' in Analysis page

brifemu:
Due to issue #2149 I have added a header and a separator for '7 day moving' to split it from 'daily'

**English version**

Screenshot 2021-12-20 at 09 38 49

**German version**

Screenshot 2021-12-20 at 10 01 07

PR #2290, opened 2022-01-07
Fix analysis tag bug

brifemu:
I created this PR to solve issue #2287, that tag shouldn't exist. It appears from PR #2212, the modification I made to separate the data. Now it's fixed, sorry for not realizing that mistake.

Some screenshots with the error fixed:

graph1

graph2

graph3

### DIFF
--- a/src/assets/js/analyse/chart/options.js
+++ b/src/assets/js/analyse/chart/options.js
@@ -131,8 +131,8 @@ export default {
 
 			seriesArray = seriesArray.map((e,i) => {
 				const value = e.data[dataPointIndex]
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@brifemu 

- The fix in this PR resolves the issue https://github.com/corona-warn-app/cwa-website/issues/2287, so it is fine on its own.

- Could you please also take a look at the issue https://github.com/corona-warn-app/cwa-website/issues/2291 which I noticed when I was reviewing this PR. Since it is a different issue I submitted a separate description with screenshot. It is based on viewing the page using this PR.

dsarkar (approved):

PR #2312, opened 2022-01-12
Release 2.16 Screenshots

larswmh:
DE:
image

EN: 
image


MikeMcC399:
Regarding "Suppress security note:" I suggest to add a comment that this applies to Android only (not iOS).

larswmh:
> Regarding "Suppress security note:" I suggest to add a comment that this applies to Android only (not iOS).

Thanks for your suggestion @MikeMcC399. In the latest commit I have inserted the sentences "This feature only affects Android devices" / "Diese Funktion betrifft nur Android Geräte"

PR #2317, opened 2022-01-12
Harmonize term "Contact Journal" / "Kontakt-Tagebuch" in Blog, FAQ, Glossary and Event Registration

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/2315 "Inconsistent use of terms for Contact Journal"

The app uses the terms:
- Journal / Contact Journal (EN)
- Tagebuch / Kontakt-Tagebuch (DE)

For reference, view the screenshots on:
- https://www.coronawarn.app/en/screenshots/#ios_journal
- https://www.coronawarn.app/en/screenshots/#android_journal
- https://www.coronawarn.app/de/screenshots/#ios_journal
- https://www.coronawarn.app/de/screenshots/#android_journal

The Blogs, FAQ, Glossary and the Event Registration form have been aligned to use only the terms used in the app.
Text relating to screenshot labelling before the current 2.15 version has not been changed.

The terms  "Contact Journal" (EN) and "Kontakt-Tagebuch" (DE) replace other equivalent terms:

- Contact diary (used in English texts)
- Kontakttagebuch (alternative spelling, without hyphen, used in German texts)

Although the term "Contact diary" is used internally in the source code, even the first version with this functionality (CWA 1.10.1) was released displaying the names as "Contact Journal" and "Kontakt-Tagebuch" on the main screen.

PR #2318, opened 2022-01-12
Fix Cypress test, change from diary to journal

MikeMcC399:
Fix Cypress failure in Event Registration test caused by PR https://github.com/corona-warn-app/cwa-website/pull/2317.


dsarkar (approved):

dsarkar:
Good morning @MikeMcC399! Thanks for the fix PR.

brifemu:
@MikeMcC399 
Tested and working without problems, thank you!

Issue #2320, opened 2022-01-13
Error in deselect daily count / 7-day moving average analysis graphs


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/analysis/ "Corona-Warn-App (CWA): Key performance indicators"
- https://www.coronawarn.app/de/analysis/ "Corona-Warn-App (CWA): Kennzahlen"

## Describe the issue

When clicking on the "Daily count" items under any of the graphs on https://www.coronawarn.app/en/analysis/ viewing in "7-day moving average" mode, the display of the "7-day moving average" values is switched off and on instead of affecting the display of "Daily count" values.

## Steps to reproduce

1. Open https://www.coronawarn.app/en/analysis/ in browser
2. Take default "Last 28 days"
3. Select "7-day moving average"
4. View graph "Warnings received by data donors (7-day-moving average)

![Warnings all](https://user-images.githubusercontent.com/66998419/149296455-599cea9d-b316-493c-bc01-cb802f133683.jpg)

5. Click on Daily count: Red risk
6. Click on Daily count: Green risk
7. Note that the solid-line display of "7-day moving average" has been removed although the "Daily count" items below the graph are shown as de-selected

![Warnings daily count deselected](https://user-images.githubusercontent.com/66998419/149296488-b7043583-0022-41bf-9d93-eb6de9f058e6.jpg)

## Expected behavior 

Clicking on the "Daily count: Red risk" or "Daily count: Green risk" under the graph should remove the display of these values (shown with dotted lines) and not remove the display of the values for "7-day moving average" (shown with solid lines).

## Suggested change

Link clicking on the Daily count items with switching the display of the Daily item values off and on.

brifemu:
@MikeMcC399 
FYI https://github.com/corona-warn-app/cwa-website/pull/2322

MikeMcC399:
- This issue is resolved and thanks to @brifemu for PR #2322!

PR #2322, opened 2022-01-13
Fix deactivate legend issue in Analysis

brifemu:
Due to issue #2320 I have created this PR to solve it. The problem was that the name of the options are equal, so just I needed to add a white space at the end of the string to difference them.


MikeMcC399:
@brifemu 
This solves the problem for me. Thank you!

dsarkar (approved):

PR #2326, opened 2022-01-14
Hide/show titles depending on search result in FAQ page

brifemu:
I have created an enhancement to the FAQ page that deals with hiding the section titles that have no questions to show. Below, I will show proof of this improvement through searching for the term 'j&j'.

### Actual FAQ search

search-old

### New FAQ search

search-new


brifemu:
@Ein-Tim
Thanks for your suggestion! I applied it, and also I have hidden the expand button when are not sections to show. Here some pics.

### English

image


### German

image

dsarkar (approved):

Issue #1776, opened 2021-09-29
Name for analysis page is verb not noun


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/

## Describe the issue

All other headings on the website https://www.coronawarn.app/ are English language nouns.

The heading "Analyse" could be an English language verb or it could be a German language noun.

![image](https://user-images.githubusercontent.com/66998419/135259802-e5da080b-1deb-4d68-a4c8-4e2262f0ee48.png)

## Suggested change

Change the heading from "Analyse" to "Analysis" to make it an English language noun, like the other headings.

(If it was supposed to be an English verb, then "Analyze" would conform to the rest of the website which uses US English, not UK English spelling.)

## References
https://dictionary.cambridge.org/dictionary/english/analyse
https://dictionary.cambridge.org/dictionary/english/analysis

---
Internal Tracking ID: [EXPOSUREAPP-9698](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9698)

dsarkar:
@FabioFuschi-SNK FYI

MikeMcC399:
PR https://github.com/corona-warn-app/cwa-website/pull/1777 removes "Analyse" from the menu on https://www.coronawarn.app/en/, however if anybody already bookmarked

https://www.coronawarn.app/en/analyse/ or
https://www.coronawarn.app/de/analyse/

the menu bar still shows "Analyse", as before. 

Also, under

https://www.coronawarn.app/en/analysis and
https://www.coronawarn.app/de/analysis/

the heading of the page is still "Analyse".  I'm not sure that that heading is needed, directly above the text "Corona-Warn-App (CWA): Kennzahlen". If it is needed there, then consider whether it should be changed to "Analysis" there as well.

Ein-Tim:
FYI: https://github.com/corona-warn-app/cwa-website/pull/1774 has removed the analysis page from the website again.

MikeMcC399:
Assuming that the analysis page will be published again, leaving this issue open.

MikeMcC399:
@dsarkar / @FabioFuschi-SNK / @mtb77 

PR https://github.com/corona-warn-app/cwa-website/pull/1793 has added the navigation name "Analyse" back.

Originally it was "Analyse", then it was changed to "Analysis", now it is "Analyse" again,

Is that what you want it to be?

This issue pointed out the inconsistency of using a verb in that place when everything else in the menu is a noun.

[...]

PR #2331, opened 2022-01-15
Rename "Analysis" to "Analyse" in German menu (Alignment to page name)

Ein-Tim:
This PR renames "Analysis" to "Analyse" in the German menu of the page.

Analysis is a German word, but it is not correct in this context, see https://www.duden.de/rechtschreibung/Analysis:

> Teil der Mathematik, in dem mit Grenzwerten gearbeitet, die Infinitesimalrechnung angewendet wird

OR

> Voruntersuchung beim Lösen geometrischer Aufgaben

I tested this change locally & everything was still correct. 

---

Here is a preview of this change:

Unknown

### DIFF
--- a/src/data/global.json
+++ b/src/data/global.json
@@ -370,7 +370,7 @@
                 },
                 {
                     "id": "analysis",
-                    "title": "Analysis",
+                    "title": "Analyse",
                     "url": "/de/analysis/"
                 },
                 {


MikeMcC399:
@Ein-Tim 

- The naming has been confusing from the beginning and it was agreed in https://github.com/corona-warn-app/cwa-website/issues/1776 to make it an English language word like the other top menu items.

It's unfortunate that the English and German words are so similar but not exactly the same.

The alternative design change would be to rename all the menu items to German when displaying any German page e.g. instead of "Science" write "Wissenschaft". At this point in time that would be quite disruptive though.

In short, I recommend no change to the menu.

Ein-Tim:
@MikeMcC399 

I disagree here, it is OK to have menu items "Science" or "Community" as they are widely known and do not have another meaning in Germany.

This is **not** the case for "Analysis" which has a different meaning in German than in English. 

Also, after clicking on "Analysis", the page already shows "Analyse" at the top, so this is only an alignment of the menu item to the name of the page.

Ein-Tim:
I have changed the title of this PR to make clearer that this is an alignment and nothing else.

MikeMcC399:
@Ein-Tim 

- I have provided an alternative suggestion to resolve the problem you highlighted in this PR. The new issue is https://github.com/corona-warn-app/cwa-website/issues/2352 which does not involve changing the language of a menu item, nor does it involve any formatting changes, so it is very easy to implement. You are welcome to comment there.

Ein-Tim:
Closing in favour of #2390.

PR #2351, opened 2022-01-18
Update FAQ admission_policy: Fix typo

larswmh:
Before:
image

With this PR:
image


src/data/faq_de.json
```diff
@@ -342,7 +342,7 @@
...
                             "Ein Testzertifikat bestätigt ein negatives Testergebnis offiziell in Form eines QR-Codes in der Corona-Warn-App. Sie können das offizielle Testzertifikat für PCR- und Schnelltests anfordern, wenn sie einen Test in der Corona-Warn-App registrieren. Dafür können sie den universellen QR-Code-Scanner in der Mitte der Registerkarte der CWA öffnen und den QR-Code zur Test-Registrierung scannen. Den erhalten sie entweder schon während der Buchung des Tests oder vor Ort beim Testen, sofern der Anbieter an die Corona-Warn-App angeschlossen ist. Nach dem Scannen des QR-Codes öffnet sich automatisch ein Fenster, in dem sie das Testzertifikat anfordern können. Mehr Informationen finden Sie in diesem Blog: <a href='../blog/2021-06-24-cwa-version-2-4' target='_blank'>Projektteam veröffentlicht Online-Schnelltestpartnersuche und Corona-Warn-App Version 2.4 mit digitalem Testzertifikat</a>",
```
MikeMcC399, line 346: In this paragraph there are multiple instance of "sie" written with lower case where "Sie" should be capitalized.

MikeMcC399 (changes requested):
Apart from the in-line comment about "sie" / "Sie" there is a full-stop missing in

"einen 2G+-Nachweis zu erbringen Beide QR-Codes"

between "erbringen" and "Beide"

I can't comment in-line on this issue, since it wasn't changed (so far) in this PR.

src/data/faq_de.json
larswmh: Fixed in the latest commit, thanks!

larswmh:
@MikeMcC399 in the latest commit I have capitalized "Sie" and added a full stop at the mentioned location in both the German and English version of the FAQ

MikeMcC399:
@larswmh 

In the English language text, there is an error in the paragraph starting "A test certificate officially confirms a negative test result in the form of a QR code in the Corona-Warn-App."

Maybe that wouldn't have happened if the German had been written correctly!

However in sentences such as "They can request the official test certificate for PCR and rapid tests when they register a test in the Corona-Warn-App." 

"They / they" is supposed to be "You / you".

MikeMcC399 (commented):
The changes done so far look good.

One whole English paragraph is wrong (using they instead of you), which was not part of this PR, but it would be good to take the opportunity to correct it here.

dsarkar:
@MikeMcC399 Thanks for the review.

dsarkar (approved):

MikeMcC399:
@dsarkar 

My comment https://github.com/corona-warn-app/cwa-website/pull/2351#issuecomment-1015484556 and

> One whole English paragraph is wrong (using they instead of you), which was not part of this PR, but it would be good to take the opportunity to correct it here.

was not addressed. I will submit a PR for this. You can take a look to see if you agree with it.

MikeMcC399:
- See https://github.com/corona-warn-app/cwa-website/pull/2357

PR #2357, opened 2022-01-18
Change "they" to "you" in #admission_policy (EN)

MikeMcC399:
This PR follows on from https://github.com/corona-warn-app/cwa-website/pull/2351#issuecomment-1015484556 correcting the paragraph starting "A test certificate officially confirms a negative test result in the form of a QR code in the Corona-Warn-App. They can request the official test certificate for PCR and rapid tests when they register a test in the Corona-Warn-App.  ..." in https://www.coronawarn.app/en/faq/#admission_policy

It seems like typos in the German text which wrote "sie" instead of "Sie" caused a translation error into "they" instead of "you".

The way it is written does not make sense, therefore all occurrences of "they" in this paragraph are changed to "you".

Corrected to:

image


dsarkar (approved):

dsarkar:
@MikeMcC399 Thank you very much!

Issue #2321, opened 2022-01-13
Analysis graph full-screen display too large


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/analysis/ "Corona-Warn-App (CWA): Key performance indicators"
- https://www.coronawarn.app/de/analysis/ "Corona-Warn-App (CWA): Kennzahlen"

## Describe the issue

Full-screen display of any graph on https://www.coronawarn.app/en/analysis/ sometimes displays the graph larger than the available browser window.

## Steps to reproduce

1. Open browser and resize to take one half of screen height and one half of screen width (this effectively uses one quarter of the screen's available area)
2. Open https://www.coronawarn.app/en/analysis/ in the existing browser window
4. Scroll down to Downloads graph
5. Click on the full-screen icon of the Downloads graph
6. Note that the graph extends vertically outside of the viewable area

Reproducible with:
- Google Chrome
- Mozilla Firefox

## Expected behavior

The full-screen display option in the analysis graphs should display the full graph inside the available browser window display area.

## Suggested change

Change the graph resizing to use the available browser area.

MikeMcC399:
- PR #2323 improves the situation. The displayable part of re-sized graph now fits in the window, 
- however I am seeing a lot of white graph below the legend:

![image](https://user-images.githubusercontent.com/66998419/149338587-9563aff5-a1e9-4124-b25e-3adc16ec7d27.png)

This is on both Google Chrome and Mozilla Firefox.

dsarkar:
related issue:
- vertical screen
- OS: Windows 10 Enterprise (Version 10.0.18363 Build 18363)
- Browser: Firefox (Version 91.4.0esr (32-Bit))
- Auflösung: 1080 x 1920

![image](https://user-images.githubusercontent.com/1409741/149752445-3d7ead07-cf6c-43e8-8a34-9b0db7c738d6.png)

MikeMcC399:
It is also possible to select full screen, which can display a graph with no scroll-bar, depending on the size / geometry of the browser window. Then if the browser window size is reduced the graph does not resize, it is cut off and there is no scroll-bar to correct the situation.

MikeMcC399:
- PR #2350 does not fix the remaining issue mentioned in https://github.com/corona-warn-app/cwa-website/issues/2321#issuecomment-1012137062. This is reproducible using the steps in the original post.

- Additional steps to test in Windows 11:
[...]

PR #2360, opened 2022-01-20
Fix responsive graph in Analysis

brifemu:
Due to issue #2321 I have created this PR. After several tries, I finally get to resolve this issue.


dsarkar (approved):

MikeMcC399:
@brifemu 

This is a big improvement. 👍🏻 

There are still however some issues. In Google Chrome with the browser window full height and 1/3 screen width:
1. Open http://localhost:8000/en/analysis/
2. Scroll down to Downloads
3. Click full-screen icon
4. Graph fills window and displays correctly. There is no vertical scroll-bar.
5. Drag bottom of window upwards to resize to smaller height.
6. The graph does not resize.
7. Click full-screen icon twice to recover.

If the browser window is resized when the display is in full-screen mode the graph should be redrawn.

dsarkar:
@MikeMcC399 Thanks for checking, we will have look at your comments.

brifemu:
@MikeMcC399 
Thank you so much for your comment! This happens because with that width gets smartphone breakpoint. That breakpoint uses `fullscreen.js` and desktop uses `expand.js`. I will implement the same resize functionality in another PR soon.

MikeMcC399:
@brifemu 

> This happens because with that width gets smartphone breakpoint. That breakpoint uses `fullscreen.js` and desktop uses `expand.js`. I will implement the same resize functionality in another PR soon.

Thank you for the explanation! With Windows 11 I make a lot of use of the built-in templates for window resizing, which are quite neat. If you haven't seen them, they look like this:

image

PR #2368, opened 2022-01-22
Remove wrong information reg. TAN hotline

Ein-Tim:
This PR removes the wrong information that the TAN hotline is not available on German national holidays.

---

Addresses https://github.com/corona-warn-app/cwa-hotline/issues/14#issuecomment-1019152190.

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -95,7 +95,7 @@
                         "active": false,
                         "textblock": [
                             "The technical hotline (<a href='#international_phone_numbers' target='_blank' >click here to see the phone numbers</a>) will help you with technical questions about the Corona-Warn-App, for instance if you have problems with exposure logging. You can reach the technical hotline from Monday to Saturday from 7 a.m. to 10 p.m. (except on German national holidays), and it is free of charge for you within Germany.",
-                            "The TAN hotline (<a href='#international_phone_numbers' target='_blank' >click here to see the phone numbers</a>) will help you to enter the TAN if you have a positive PCR test result and if you have not received a TAN or document with a PCR-Test QR code. The TAN hotline is available Monday to Sunday from 7 a.m. to 8 p.m. German local time (except on German national holidays). The call is free of charge from within Germany.",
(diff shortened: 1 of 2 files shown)


MikeMcC399:
@Ein-Tim 
Perhaps where you were considering a full-stop, use a colon before the list of National / International phone numbers?

Ein-Tim:
@MikeMcC399 

Yes, in German there was one `.` at the end but in English not, so I removed the `.` from the German entry. Let me know if you think that a `.` should rather be added to the English text as well.

MikeMcC399:
@Ein-Tim 
I saw that it was previously written inconsistently. I would suggest replacing the full-stop with a colon, or adding a colon where there was none previously. For example:

"The TAN hotline is available Monday to Sunday from 7 a.m. to 8 p.m. German local time:

**National:** 0800 7540002"

You can check style guides in Internet. They recommend using a colon after a complete sentence and before a list in English. See https://www.grammarbook.com/blog/colons/colons/.

I think that is the same in German according to https://www.duden.de/sprachwissen/rechtschreibregeln/doppelpunkt

dsarkar (approved):

Ein-Tim:
The Twitter account says that the Hotline is not available on national public holidays: https://twitter.com/coronawarnapp/status/1490275898481856513?s=20&t=CHBK6YjX_IEJrkZ7ZQEZlQ
What is correct?

MikeMcC399:
@Ein-Tim 

Maybe you could ask on the Twitter channel?

It contradicts the updated information on https://www.coronawarn.app/de/faq/#international_phone_numbers (as you know).

Ein-Tim:
@MikeMcC399 
I asked them for checking in https://twitter.com/EinTim2/status/1490292875715874824?s=20&t=O440Fm2rlS4I4LnJESS-gA.

Ein-Tim:
Ok this was an error from the Twitter-Team: https://twitter.com/coronawarnapp/status/1490293098349580289?s=20&t=O440Fm2rlS4I4LnJESS-gA

PR #2377, opened 2022-01-26
Update FAQ: Effective validity of vaccination decreased to 270 days

larswmh:
Decreases the effective validity mentioned in glossary and faq entry from 1 year to 270 days (starting on February 1st)


Ein-Tim:
Is this really true? I think in Germany vaccination certificates are still valid one year only for intra-EU traveling the effective validity was reduced to 270 days, see https://github.com/corona-warn-app/cwa-wishlist/issues/726.

MikeMcC399:
@Ein-Tim 
> Is this really true? I think in Germany vaccination certificates are still valid one year, only for intra-EU traveling the effective validity was reduced to 270 day

There are no special German vaccination certificates. The only certificates being used are EU digital COVID vaccination certificates. If the EU regulations change the validity to 270 days (see regulations https://github.com/corona-warn-app/cwa-website/issues/2240#issue-1087441547) then automatically in Germany that is the same, also for EU certificates issued by other countries being used internally in Germany.

I'm sure a lawyer experienced in EU law and how EU regulations apply to member countries can explain it better, but I think what I write is basically correct.

Ein-Tim:
@MikeMcC399 

I asked because https://ec.europa.eu/commission/presscorner/detail/en/ip_21_6837 especially states:

> Today, the Commission adopted rules relating to the EU Digital COVID Certificate, establishing a binding acceptance period of 9 months (precisely 270 days) of vaccination certificates **for the purposes of intra-EU travel.**

Are you sure that this also affects the validity in Germany?

MikeMcC399:
@Ein-Tim 

> I personally read this as a change only affecting the validity for entering another country.
> 
> But for sure, neither you nor me have to understand this, but only the project team.

To be sure, it would need a statement from the German government.

MikeMcC399:
@Ein-Tim 
The EU legal text is the [Commission Delegated Regulation (EU) 2021/2288 of 21 December 2021 amending the Annex to Regulation (EU) 2021/953 of the European Parliament and of the Council as regards the acceptance period of vaccination certificates issued in the EU Digital COVID Certificate format indicating the completion of the primary vaccination series (Text with EEA relevance)](https://eur-lex.europa.eu/eli/reg_del/2021/2288/oj). The website allows you to read it in English, German or any of the official EU languages.

Ein-Tim:
@MikeMcC399 

Thanks. I read it and it confused me even more. 

In (10) it is written:

> To avoid diverging and disruptive measures, it is thus necessary to establish, **for the purpose of travel**, a standard acceptance period of 270 days for vaccination certificates indicating the completion of the primary vaccination series. This takes into account the guidance of the European Centre for Disease Prevention and Control regarding the administration of booster doses as of six months after completion of the primary vaccination series, and provides for an additional period of three months to ensure that national vaccination campaigns can adjust and citizens can have access to the administration of boosters. **To ensure a coordinated approach, Member States should not accept vaccination certificates indicating the completion of the primary vaccination series if more than 270 days have passed since the administration of the dose indicated therein. At the same time, and in order to ensure a coordinated approach, Member States should not, for the purpose of travel, provide for an acceptance period shorter than 270 days.** Within this standard acceptance period, vaccination certificates indicating the completion of the primary vaccination series should continue to be accepted by a Member State even if it is already administering booster doses.

However, then later in 16 it states:

> The standard acceptance period should not be included as a new data field in the vaccination certificate, but should be applied at the level of verification, by adapting the mobile applications used to verify EU Digital COVID Certificates. **If a relevant vaccination certificate indicating a date of vaccination exceeding the acceptance period of 270 days is presented to the verifier, the mobile application used for verification should indicate the certificate as expired.** Applying the standard acceptance period at the level of verification allows for easier follow-up of further evolution in scientific evidence than if a set expiry date is included in the certificates. For the purpose of applying the standard acceptance period at the level of verification, the data field on the date of vaccination should be modified. Doing so is preferable to adding a new data field specifically on the expiry date of a vaccination certificate. Adding a new data field would imply the need either to re-issue already issued vaccination certificates or to establish technical systems capable of interpreting, at the same time, already issued vaccination certificates without an expiry date and newly issued vaccination certificates featuring an expiry date. To ensure its uniform application, the standard acceptance period of vaccination certificates should be incorporated into the verification applications of all Member States.

Here, there is no mention of the purpose of travel anymore. So either check apps will need a "border" mode where all vaccination certificates which are older than 270 days (but booster vaccination certificates) are rejected or this in the end means that regardless of the purpose the certificate is not accepted anymore after 270 days.

What do you say?

@dsarkar Is this being cross checked with the lawyers?

MikeMcC399:
@Ein-Tim 
> What do you say?

My opinion is not what counts here. Only an official statement has value.

Ein-Tim:
@MikeMcC399 Definitely! I assume the team is already clarifying this.

MikeMcC399:
@larswmh 

The discussion has drifted into a legal one here, which would maybe better take place in a separate issue, instead of a PR discussion.

Are you implementing a German decision into the FAQs or are you implementing the EU regulation into the FAQs? If you are implementing the EU regulation, then I suggest to review the wording, which uses Germany as a reference a lot.

For instance in: "This vaccination took place on a date X and is valid from then on e.g. in Germany for 270 days after February 1st 2022, according to the current state of scientific knowledge." it makes it sound like it is only Germany (not the EU) and the reference to scientific knowledge should probably be replaced by a reference to the regulation, if indeed the EU regulation is what is triggering this change.

Ein-Tim:
After reading the tagesschau article [EU-Impfnachweis Ohne Booster nur noch 9 Monate gültig](https://www.tagesschau.de/ausland/europa/impfnachweis-eu-neun-monate-101.html), which says:
> Die EU-Impfnachweise sind ohne Auffrischungsimpfung nur noch neun Monate gültig. Nach Ablauf dieser Frist werden Menschen ohne Booster bei Grenzübertritten wie Ungeimpfte behandelt.

I'm quite certain that this only affects the validity for traveling but not the validity in Germany.

MikeMcC399:
@Ein-Tim

I didn't read the Tagesschau article quite that way, however [ntv: EU-Impfpass nur 270 Tage gültig](https://www.n-tv.de/panorama/Reisen-ohne-Booster-wird-schwieriger-article23096383.html) says explicitly:

"In Deutschland gibt es zurzeit keine Regelung, die die Anerkennungsdauer von Impfnachweisen begrenzt."

dsarkar:
continued here: https://github.com/corona-warn-app/cwa-website/pull/2408

PR #2402, opened 2022-02-02
New FAQ #vc_2of1

browkin:
- Added new vaccination article to the FAQ

**PREVIEW:** 

**DE**:
21DE

**EN**:
21EN


Ein-Tim:
Please also update https://www.coronawarn.app/de/faq/#vac_cert_jj or at least add a reference.

dsarkar:
@Ein-Tim Thanks for the suggestion, yes, we should reference/link these two articles mutually.

src/data/faq.json
```diff
@@ -573,6 +573,15 @@
                             "For more information on the STIKO vaccination recommendations and all about COVID-19 and vaccination, see <a href='https://www.rki.de/DE/Content/Infekt/Impfen/ImpfungenAZ/COVID-19/COVID-19.html' target='_blank' rel='noopener noreferrer'>RKI - Impfungen A - Z - COVID-19 und Impfen (German)</a>."
                         ]
                     },
+                    {
+                        "title": "2/1 vaccination certificates",
+                        "anchor": "vc_2of1",
+                        "active": false,
+                        "textblock": [
+                            "The Android app and iOS app versions 2.17.1 differ with regard to the display of the 2/1 vaccination certificate after a first vaccination with Johnson and Johnson: In the iOS app, the complete vaccination status is currently displayed immediately. For the Android app, this is currently not the case until day 15. We will immediately adjust this with the Android app so that immediate full immunization is also displayed here with a 2/1 vaccination certificate.",
```
MikeMcC399, line 581: "We will immediately adjust this with the Android app so that immediate full immunization is also displayed here with a 2/1 vaccination certificate."

"immediately adjust" would mean in the next second or two

instead

"We are urgently in the process of adjusting ..."

also the word order  in "immediate full immunization" could be improved, so I suggest the full changed sentence as:

**"We are urgently in the process of adjusting this with the Android app so that full immunization is also immediately displayed here with a 2/1 vaccination certificate."**

MikeMcC399 (changes requested):
Suggestion for rewording in English text. See inline comment.

Ein-Tim:
@dsarkar 

>  Thanks for the suggestion, yes, we should reference/link these two articles mutually.

Will you take care of this?

src/data/faq.json
```diff
@@ -573,6 +573,15 @@
                             "For more information on the STIKO vaccination recommendations and all about COVID-19 and vaccination, see <a href='https://www.rki.de/DE/Content/Infekt/Impfen/ImpfungenAZ/COVID-19/COVID-19.html' target='_blank' rel='noopener noreferrer'>RKI - Impfungen A - Z - COVID-19 und Impfen (German)</a>."
                         ]
                     },
+                    {
+                        "title": "2/1 vaccination certificates",
```
MikeMcC399, line 577: Suggest to capitalize "Vaccination".

It's the only article which starts with a digit, so there is no comparative article to know what the convention is.
Articles which start with letters all have their first word capitalized in English though.

MikeMcC399:
@dsarkar 

The merged article uses the term "immediate" (sofort). When will the fix be released? If it is not today, then I suggest changing the wording.

src/data/faq.json
dsarkar: alright, thank you!

PR #2420, opened 2022-02-07
changed link markdown to html

svengabr:
Fixes [#2410](https://github.com/corona-warn-app/cwa-website/pull/2410#issuecomment-1031311837)

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -2703,7 +2703,7 @@
                 {
                     "term": "Effective validity",
                     "anchor": "effective_validity",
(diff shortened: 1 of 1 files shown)


src/data/faq.json
```diff
@@ -2703,7 +2703,7 @@
...
+                    "description": "We refer to the duration of vaccination protection as 'effective validity'. The effective validity of the vaccination thus counts from the last vaccination date. The effective validity of the vaccination protection may vary from country to country. According to the <a href='https://www.rki.de/DE/Content/InfAZ/N/Neuartiges_Coronavirus/Genesenennachweis.html' target='_blank' rel='noopener noreferrer'> RKI </a>, the maximum effective validity of a <a href='https://www.coronawarn.app/en/faq/#glossary_recovery_certificate' target='_blank' rel='noopener noreferrer'>recovery certificate</a> is 90 days."
```
MikeMcC399, line 2706: It would be better to use a relative link.
`rel='noopener noreferrer'` is not necessary for references a location on the same web and can be removed.

So this would be better and shorter:
`<a href='#glossary_recovery_certificate' target='_blank'>recovery certificate</a>`

This is the way the link is done in faq_de.json.

MikeMcC399 (changes requested):
Suggest to change link to relative.

src/data/faq.json
svengabr: Done

PR #2428, opened 2022-02-11
update/browser-compatibility-faq

browkin:
- Added an FAQ-Article for the error "Your browser is not compatible." under Resolved -> Problems with the QR code/link for event/test registration

### **Preview:**

**EN:**
browserEN

**DE:**
browserDE

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -2245,6 +2245,18 @@
                 "id": "resolved",
                 "accordion": [
                     {
+                        "title": "Problems with the QR code/link for event/test registration",
+                        "anchor": "browser_not_compatible",
+                        "active": false,
+                        "textblock": [
+                            "If you are unable to register the event or test with your link or QR code, please follow these steps:",
+                            "<b>iOS:</b>",
+                            "<li>Make sure you have the app installed</li> Try using one of the following workarounds to register the test or event: <li>Long press on the link and choose \"Open in Corona-Warn\" </li> <li>Alternatively: Open the link in Safari </li><br/> ",
+                            "<b>Android:</b>",
+                            "<li>Make sure you have the app installed</li> Try using one of the following workarounds to register the test or event: <li>Long press on the link and choose \"Open link in external App\"</li> <li>Open the link in Chrome</li> <li>Enable the setting 'Open links in other apps'. Example for a Samsung A21 device (the exact process depends on the device): <br/> Settings -> Apps -> Default Apps -> Opening Links -> Corona-Warn -> open supported links in this app \"Always allow\" "
+                        ]
+                    },
+                    {
                         "title": "[Google/Android]: Fixed security vulnerability in the Exposure Notification Framework",
                         "anchor": "google_security_vulnerability",
                         "active": false,
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -2252,6 +2252,18 @@
                 "title": "Geklärt",
                 "accordion": [
                     {
+                        "title": "Probleme mit dem QR-Code/Link zur Event-/Testregistierung",
+                        "anchor": "browser_not_compatible",
+                        "active": false,
+                        "textblock": [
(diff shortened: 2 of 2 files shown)


MikeMcC399:
@browkin 

For Samsung Internet Version 16.0.6.23
1. Open Samsung Internet browser
2. Tap three-bar menu icon
3. Tap Settings > Useful features > Open links in other apps (enable)

however this still does not solve the issue.

I scanned an event code on Samsung Galaxy A5 (2017) (no CWA installed, Samsung Internet is default browser) using a 3rd-party QR code scanner, after scanning I selected open for the link https://e.coronawarn.app?v=1#and so on. I still got the error message: "Your browser is not compatible."

MikeMcC399:
I checked also for Firefox on Android 8.0 
Firefox three-dot symbol > Settings > Open links in app (enable)

Enabling this option makes no difference. If CWA is not installed and an event QR code is scanned, using the built-in QR code scanner of Firefox, the error message "Your browser is not compatible." is output.
(To access the QR code reader in Firefox, tap the search box, then tap "Scan".)

Ein-Tim:
It seems like that under Android only the Chrome Browser is compatible.

MikeMcC399:
I suggest to hold back on this PR. Looking at this more closely there are multiple issues with the website display https://e.coronawarn.app when no app is installed.

I will submit several issues to https://github.com/corona-warn-app/cwa-event-landingpage/issues describing what I found. That will take me a little while due to combinations of browsers, Android versions and locales.

Ein-Tim:
Suggestion to add the https://github.com/corona-warn-app/cwa-website/labels/on%20hold label.

MikeMcC399:
There are so many different combinations of QR code scanner, browser and Android version that it is practically impossible to test them all.

In my limited tests I found that if CWA is installed then Google Chrome, Firefox and Samsung Internet were all able to handle a scanned event QR code and open CWA if each one of them was set as default browser in turn and the Android 12 built-in QR code scanner or camera was used.

I did find though that using Firefox's built-in QR code scanner (tap the search field, tap Scan) that it showed the incompatibility message, whether or not "Open links in apps" was enabled in Firefox's settings. It is still possible to get Firefox to open the app, by going to Firefox settings and selecting "Open in app". In the end that does not work, because CWA rejects this with the error "Invalid QR code".

The simplest advice for Android users, if scanning a QR event code using the device's camera, built-in QR code scanner or third-party QR code scanner does not call up the CWA app and allow to check in, is to:

- ensure that the Corona-Warn-App is installed
- scan the event QR code using the Scan button of the Corona-Warn-App

The same probably applies for RAT QR codes. I couldn't test that though.

MikeMcC399:
@dsarkar 

I apologise for misreading the instructions about enabling links. I did actually find the settings. I also found that they were enabled by default.

I still think the best advice though is for users to scan using CWA. Now that there is one scan button for everything in CWA, it is the simplest and most reliable alternative.

luisoos (approved):

Ein-Tim:
Dear Open-Source-Team (@dsarkar & @larswmh),

there was no activity on this PR for a prolonged period of time. Could you share your plan how to proceed here with us? Thanks!

larswmh:
@Ein-Tim @MikeMcC399 @luisoos thanks for your participation.

The reason we planned to add this FAQ is that users occasionally receive links per mail instead of QR codes to scan. While we recommend to scan QR codes if applicable, we still want to cover the rare cases where this is not an option and provide help for this in the FAQ. 
However, we have come to the conclusion that there are some textual issues we would like to address before merging this into FAQ. Thanks for your patience

Ein-Tim:
This PR is heavily out of sync with the `main` branch, which would require major manual work to make this PR ready to merge.
I therefore request this PR to be closed & if the changes should be applied, this should be done via a new PR.

GisoSchroederSAP:
I agree with the statements above. Just close this issue. We may split the tasks in "smaller" PRs, if necessary.

GisoSchroederSAP:
I agree to close tis issue for now. If necessary, we may go forward with new PR's step-by-step

PR #2438, opened 2022-02-17
Event-Registration: Error field focus and validation logic fix

thorbenkuro:
Focus is now set to the first error field on button press and validation logic regarding invalid dates is fixed to display the correct error message for every error case.


svengabr (approved):

svengabr:
Changes looking good, thanks!

PR #2439, opened 2022-02-17
Event-Registration: Changed error-message placement

thorbenkuro:
Error messages are now displayed above the input-fields inside the label-tag


svengabr:
Looking good, thanks @thorbenkuro

PR #2440, opened 2022-02-17
Dashboard: Changed html-tag for analysis board headings to <h3>

thorbenkuro:


svengabr:
Please change the layout so the heading will look the same like before. The newdlines are a little big too big. Thanks!

diff1

src/partials/page-analyse.html
```diff
@@ -137,7 +137,7 @@ <h1 class="headline headline-heavy">
 						<div class="analyseBoard-head">
 							{{#if title}}
 								<div class="analyseBoard-title">
-									<span class="analyseBoard-title-title">{{title}}</span> 
+									<h3 class="analyseBoard-title-title">{{title}}</h3>
```
svengabr, line 140: Please add some css to make the new <h3> look the same as the <span> before

thorbenkuro:
I changed the text styling back to the original version.

svengabr (approved):

PR #2441, opened 2022-02-17
Insert aria-live on match count in FAQ page

brifemu:
This PR makes FAQ count result readable for screen lectors.


svengabr:
Changes are looking good, thanks @brifemu !

svengabr (approved):

PR #2442, opened 2022-02-17
Ser area-current on global nav and asides

brifemu:
This PR makes nav selected pages and aside options selected readable for screen lectors.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -37,8 +37,8 @@ $(document).ready(function(){
 
     $('.js-menu .js-scroll-navigate a').on('click tap', function(){
         $(this).parents('.js-menu').first().removeClass('active');
-        $(this).parents('.js-menu').first().find('a').removeClass('active');
-        $(this).addClass('active');
+        $(this).parents('.js-menu').first().find('a').removeClass('active').removeAttr('aria-current');
+        $(this).addClass('active').attr('aria-current', $(this).text());
     });
 
     const anchors = Array.from(document.querySelectorAll('.js-anchor'));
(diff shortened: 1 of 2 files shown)


src/assets/js/app.js
```diff
@@ -37,8 +37,8 @@ $(document).ready(function(){
 
     $('.js-menu .js-scroll-navigate a').on('click tap', function(){
         $(this).parents('.js-menu').first().removeClass('active');
-        $(this).parents('.js-menu').first().find('a').removeClass('active');
-        $(this).addClass('active');
+        $(this).parents('.js-menu').first().find('a').removeClass('active').removeAttr('aria-current');
+        $(this).addClass('active').attr('aria-current', $(this).text());
```
svengabr, line 41: According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current , the aria-current attribute has to contain a special keyword. In this case, I think `aria-current="true"` is what we are aiming for. all other entries need to have it set to `false`

src/partials/navlist-item.html
```diff
@@ -9,7 +9,13 @@
         {{#if
         item.external}} nav-link_external{{/if}}{{#if item.disabled}}
         disabled{{/if}}" {{#if item.external}} target="_blank"
-        aria-label="{{title}}" rel="noopener noreferrer" {{/if}}>
+        aria-label="{{title}}" rel="noopener noreferrer" {{/if}}
+        {{#unless item.external}}
+        {{#ifequal @root.page-name item.id}} aria-current="{{title}}"{{/ifequal}}
```
svengabr, line 14: According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current , the aria-current attribute has to contain a special keyword. In this case, I think `aria-current="page"` is what we are aiming for. All other menu entries need to have the `aria-current` attribute removed

svengabr:
Rejected for now, please see my comments on top

svengabr (approved):

PR #2445, opened 2022-02-18
Dashboard time range dialog: Copied action buttons to the top.

thorbenkuro:
This is how it looks now: New time range dialog


MikeMcC399:
@thorbenkuro 
At the top, the order of the buttons "Übernehmen" and "Abbrechen" is different to the bottom where "Abbrechen" is on the left with "Übernehmen" on the right.

Is that desirable?

svengabr:
For me, it's ok if the button order is changed as it makes (maybe) more sense for blind people to use the form. The main problem was that blind people need to hit [Tab] like 30 times until they go through the date picker to reach the submit button. I won't merge this yet and leave it up for discussion. Maybe someone knows a more pretty solution.

svengabr:
Will merge the branch for now due to accessibility reasons. If there is a need, it will be changed later. At the moment I think this should be fine.

PR #2446, opened 2022-02-18
Footer: Added aria-label to social media links

thorbenkuro:


svengabr:
Changes are approved, thanks @thorbenkuro !

PR #2450, opened 2022-02-18
Layout images alt text

ibexad:


svengabr (approved):

svengabr:
The background is: Remove alt text from images that serve only a layout purpose and add nothing to the content.

Changes are approved, thanks @ibexad !

PR #2463, opened 2022-02-22
order list added to how the app works section

ibexad:


svengabr:
Changes approved, thanks @ibexad !

PR #2467, opened 2022-02-23
[32] Add focus at some class elements

brifemu:
This PR adds focus prop to some elements that couldn't be visible focused by pressing the tab key on the site.


svengabr:
Can not approve as there are diffrerences shown in Windows Firefox, Chrome and Safari on Mac all looking different. Please find a way to make focus clearly visible with a minimum contrast of 3:1 in every common browser.

firefox
chrome

svengabr:
Changes approved, thanks @brifemu !

PR #2473, opened 2022-02-23
Grouped form inputs and added separate labels

thorbenkuro:
Here are two screenshots from the event-qr-code generator form with the changes:
Event-Form1
Event-Form2


svengabr (approved):
Changes approved, thanks @thorbenkuro !

PR #2505, opened 2022-03-04
changing text colors in dashboard page

ibexad:


svengabr:
This PR contributes to EXPOSUREAPP-11867 but won't resolve it.

PR #2510, opened 2022-03-07
skip to main link added for header and side menu

ibexad:


svengabr:
Changes approved, thanks @ibexad !

PR #2523, opened 2022-03-10
Remove italic effect from science blog text

larswmh:
image
With this PR
image

### DIFF
--- a/src/data/science.json
+++ b/src/data/science.json
@@ -2,7 +2,7 @@
   "en": {
     "headline": {
       "title": "Science Blog",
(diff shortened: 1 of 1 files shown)


dsarkar (approved):

MikeMcC399:
Can I suggest also removing the text "Over the next few weeks" since in fact the science blog spans a period of months rather than weeks.

The changed sentence would be "We will be using this blog to publish answers to these questions as well as results from the research undertaken by the Robert Koch Institute." which would be more accurate and not tied to a time-period of weeks.

larswmh:
Thanks for the suggestion @MikeMcC399. It has been removed with https://github.com/corona-warn-app/cwa-website/pull/2523/commits/2633c2a44230156a903a2a08b52007731a893cc9

PR #2526, opened 2022-03-10
Add FAQ: fault_tolerant_name_handling

larswmh:
image
image

---
Internal Tracking ID: [EXPOSUREAPP-12153](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12153)

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -417,7 +417,7 @@
                         "textblock": [
                             "As of version 2.19 of the Corona-Warn-App, the assignment of certificates to persons is more fault-tolerant. Small differences in the spelling of first and last names are ignored so that the certificates are still automatically assigned to the same person.",
                             "The Corona-Warn-App can recognize that certificates belong to one person if the dates of birth match exactly and: <ul><li>the first and last names match exactly,</li><li>at least one first name and at least one last name match or</li><li>the first and last names are swapped.</li></ul>",
-                            "In case your name contains umlauts, they will be converted automatically. The academic doctorate now no longer leads to the distinction of names.",
+                            "If a name contains umlauts, they will be converted automatically. A doctorate in a name will now be ignored for the purpose of comparing names.",
                             "If the Corona-Warn-App changes anything in the grouping of your certificates as a result of this adjustment, you will be informed about it with a notification.",
                             "For more information, visit the blog: <a href='../blog/2022-03-16-cwa-version-2-19/' target='_blank'>CWA 2.19 improves the assignment of certificates to persons</a>"
                         ]
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -417,7 +417,7 @@
                         "textblock": [
                             "As of version 2.19 of the Corona-Warn-App, the assignment of certificates to persons is more fault-tolerant. Small differences in the spelling of first and last names are ignored so that the certificates are still automatically assigned to the same person.",
                             "The Corona-Warn-App can recognize that certificates belong to one person if the dates of birth match exactly and: <ul><li>the first and last names match exactly,</li><li>at least one first name and at least one last name match or</li><li>the first and last names are swapped.</li></ul>",
-                            "In case your name contains umlauts, they will be converted automatically. If you have a title of \"Dr.\", it will be ignored.",
+                            "In case your name contains umlauts, they will be converted automatically. The academic doctorate now no longer leads to the distinction of names.",
                             "If the Corona-Warn-App changes anything in the grouping of your certificates as a result of this adjustment, you will be informed about it with a notification.",
                             "For more information, visit the blog: <a href='../blog/2022-03-16-cwa-version-2-19/' target='_blank'>CWA 2.19 improves the assignment of certificates to persons</a>"
                         ]
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -409,6 +409,18 @@
                             "According to the decision of the federal and state health ministers, vaccinated or recovered persons who have received a booster vaccination currently do not have to provide a recent test in case of a 2G+ regulation. As of version 2.18 of the Corona-Warn-App, the proof status '2G+' is displayed if a certificate of a booster vaccination is stored in the app.",
                             "See also: <ul><li><a href='#admission_policy' target='_blank'>General information about the status proof display</a></li><li><a href='#admission_policy_2gplus' target='_blank'>What does the 2G+ status proof mean in concrete terms?</a></li><li><a href='#admission_policy_2gplus_not_shown' target='_blank'>Why is the 2G+ status proof not displayed even though I have a vaccination certificate or certificate of recovery and a negative test?</a></li><li><a href='#admission_policy_2gplus_howto' target='_blank'>How do I use the combined 2G+ display?</a></li></ul>"
                         ]
+                    },
+                    {
+                        "title": "What happens if the names of my certificates do not exactly match?",
+                        "anchor": "cert_fault_tolerant_name_handling",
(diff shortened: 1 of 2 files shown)


dsarkar:
We might still add some details about the rules of the fault tolerant mechanism. Stand by.

MikeMcC399:
Looks good so far!

The PR depends on the blog article `blog/2022-03-16-cwa-version-2-19` which is currently absent. Until this is provided there will be a dead link.

src/data/faq.json
```diff
@@ -416,8 +416,8 @@
...
-                            "When comparing two certificates, the Corona-Warn-App can now break down the first and last names into their components in each case and thus recognize that two certificates belong to one person if: <ul><li>the first and last names either match exactly or at least one first name and at least one last name match and</li><li>the dates of birth match exactly.</li></ul>",
-                            "The precondition that the dates of birth in the certificates must match exactly is retained.",
+                            "The Corona-Warn-App can recognize that certificates belong to one person if the dates of birth match exactly and: <ul><li>the first and last names match exactly,</li><li>at least one first name and at least one last name match or</li><li>the first and last names are swapped.</li></ul>",
+                            "In case your name contains umlauts, they will be converted automatically. If you have a doctorate, it will be ignored.",
```
MikeMcC399, line 420: "doctorate" could be a Ph.D. so I would say "a title of `Dr`. in your certificate name" (or whatever the actual text is which is getting filtered out)

larswmh: Changed to "If you have a title of \"Dr.\", it will be ignored." in https://github.com/corona-warn-app/cwa-website/pull/2526/commits/60a3277c3ac3cce573c6ae0bc6539788535bce4c, thanks!

MikeMcC399:
@larswmh 

It seems in Android release/2.19.x, that the literal text "DR" is ignored. I'm not sure whether "Dr." (with a dot is also ignored) - probably yes, if "." is converted to "<" though. Maybe worth checking because the test cases don't include "DR." (with a dot)?

See https://github.com/corona-warn-app/cwa-app-android/blob/a74dc73a8df7d164eac438f7210ed80658e2c015/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificatePersonIdentifier.kt#L101-L120

```
/*
Function sanitize names according to the following requirements:

- dots `.` and dashes `-` shall be replaced by `<`
- the string shall be converted to upper-case
- German umlauts `Ä/ä`, `Ö/ö`, `Ü/ü` shall be replaced by `AE`, `OE`, `UE`
- German `ß` shall be replaced by `SS`
- the string shall be trimmed for leading and training whitespace
- the string shall be trimmed for leading and trailing `<`
- any whitespace in the string shall be replaced by `<`
- any occurrence of more than one `<` shall be replaced by a single `<`

Name fields shall be split into components as follows:

- the string shall be split by `<`
- components with the value `DR` shall be filtered out
 */
private fun String.sanitizeName(): List<String> {
    val filteringList = listOf("DR")
    return uppercase()
```

- from PR https://github.com/corona-warn-app/cwa-app-android/pull/4858

larswmh:
@MikeMcC399 As we understand it, "Dr." is also ignored because, for example, at the beginning of the comment it is mentioned that the dot is removed, the string is converted to uppercase, and then later "DR" is removed. We think that the current wording is sufficient

src/data/faq.json
```diff
@@ -417,7 +417,7 @@
                         "textblock": [
                             "As of version 2.19 of the Corona-Warn-App, the assignment of certificates to persons is more fault-tolerant. Small differences in the spelling of first and last names are ignored so that the certificates are still automatically assigned to the same person.",
                             "The Corona-Warn-App can recognize that certificates belong to one person if the dates of birth match exactly and: <ul><li>the first and last names match exactly,</li><li>at least one first name and at least one last name match or</li><li>the first and last names are swapped.</li></ul>",
-                            "In case your name contains umlauts, they will be converted automatically. If you have a title of \"Dr.\", it will be ignored.",
+                            "In case your name contains umlauts, they will be converted automatically. The academic doctorate now no longer leads to the distinction of names.",
```
MikeMcC399, line 420: "The academic doctorate now no longer leads to the distinction of names." 

I wonder why this was changed? It seems less easy to understand than the previous text. I'm not even sure that the term "academic doctorate" is accurate. Depending on a country's legal framework there is also a "professional doctorate" which allows the holder to title themselves "Dr.".

The sentence now sounds like a literal translation of the German "Der akademische Doktortitel führt nun nicht mehr zur Unterscheidung von Namen." which doesn't read well in English.

Let me know if you want a suggestion for improvement, otherwise I will leave it be.

larswmh: @MikeMcC399 thanks for your feedback. We would be happy to hear your suggestion.

src/data/faq_de.json
```diff
@@ -408,6 +408,18 @@
...
                         ]
+                    },
+                    {
+                        "title": "Was passiert, wenn die Namen meiner Zertifikate nicht exakt übereinstimmen?",
+                        "anchor": "cert_fault_tolerant_name_handling",
+                        "active": false,
+                        "textblock": [
+                            "Ab Version 2.19 der Corona-Warn-App ist die Zuordnung von Zertifikaten zu Personen fehlertoleranter gestaltet. Kleine Unterschiede in der Schreibweise von Vor- und Nachnamen werden ignoriert, sodass die Zertifikate trotzdem derselben Person automatisch zugeordnet werden.",
+                            "Die Corona-Warn-App erkennt, dass Zertifikate zu einer Person gehören, wenn die Geburtsdaten exakt übereinstimmen und: <ul><li>Vor- und Nachnamen exakt übereinstimmen,</li><li>mindestens ein Vorname und mindestens ein Nachname übereinstimmen oder</li><li>Vor- und Nachnamen vertauscht sind.</li></ul>",
+                            "Für den Fall, dass Ihr Name Umlaute enthält werden diese automatisch umgewandelt. Der akademische Doktortitel führt nun nicht mehr zur Unterscheidung von Namen.",
```
MikeMcC399, line 419: "Für den Fall, dass Ihr Name Umlaute enthält werden diese automatisch umgewandelt."

In this sentence there is a change to using the 2. person plural, which is a break with the rest of the description. Also there is a comma missing.

Suggest changing to:
"Für den Fall, dass ein Name Umlaute enthält, werden diese automatisch umgewandelt."

src/data/faq.json
MikeMcC399: @larswmh 

I noticed a grammar issue in the German paragraph, regarding the sudden use of "you" in the middle of the text, so I'm also suggesting to change it in the English version:

"If a name contains umlauts, they will be converted automatically. A doctorate in a name will now be ignored for the purposes of comparing names.",

MikeMcC399 (changes requested):
Some suggestions regarding the text about Umlauts and Doctor titles.

src/data/faq_de.json
larswmh: Changed it in https://github.com/corona-warn-app/cwa-website/pull/2526/commits/3d76e6f74e60bd94c4530492103dd75055b0f04d, thanks for the suggestion

src/data/faq.json
larswmh: Changed it in https://github.com/corona-warn-app/cwa-website/pull/2526/commits/3d76e6f74e60bd94c4530492103dd75055b0f04d, thanks for the suggestion

dsarkar:
@MikeMcC399 Thanks for review!

PR #2528, opened 2022-03-14
Update FAQ 2.19 #rat_tan_hotline: Mention rapid PCR tests

larswmh:
imageimage

---
Internal Tracking ID: [EXPOSUREAPP-12118](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12118)

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -938,11 +938,11 @@
                         ]
                     },
                     {
-                        "title": "Can I share my positive rapid antigen test result via the TAN hotline?",
+                        "title": "Can I share my positive rapid antigen test or PCR rapid test result via the TAN hotline?",
                         "anchor": "rat_tan_hotline",
                         "active": false,
                         "textblock": [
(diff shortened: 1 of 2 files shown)


Ein-Tim (changes requested):
I suggest to think about adjusting the reasoning.

1. Rapid PCR tests are always performed by trained personnel, never by users (there is no such thing as a "self rapid PCR tests")
2. [Testifly](https://testifly.de/en/), is connected to the CWA. It offers the service that users carry out a self rapid antigen test at home, record themselves by doing so and then sending the video to testify. There a staff member of Testifly looks at the video and if everything is OK, the result of the test is certified.
While one could argue that this is according to the dual control principle, it was definitely not performed by trained personnel.

dsarkar:
@Ein-Tim Good point at this stage. We will remove last sentence for the time being. Thanks for the review.

Ein-Tim (approved):
Thank you, looks good to me!

dsarkar:
ok, lets hope it will not result in a merge conflict....

Ein-Tim:
@dsarkar I think it should not as I identically copied the content from the 2.19 branch. 

I opened #2535.

Issue #1137, opened 2021-04-24
Warn when creating new code with same data


fynngodau:
Regarding the QR code generator, users might not be aware that generating a new code with the same data is actually a very different event.

This is different from the mobile apps, where it is clear because the code is stored in the app, and deleting and re-creating an event makes it clear what one is doing.

Therefore, there should be a warning when generating a new code with the same data.

DerVogel2020:
That might be difficult, since in my understanding, no data is saved on any server.
The same problem occurs if I create an event with the same data as my colleague.
I think you're right, a warning should be always displayed. Always because my browser doesn't know of the data my colleague entered to generate a qr code.

fynngodau:
What I meant was to show a warning when the Create button was clicked multiple times in a row, as this would generate multiple different codes. Maybe the button should even be blocked until any of the input fields are changed.

rugk:
or… alternatively… hash the data with a password-hash and save the data in local storage.
This way they would be non-extractable (as long as the event name etc. is “random” enough), but you can compare whether the coe is the same.

If you’d want, maybe change the hash once a day  or save it only as in the session storage, so it is deleted fast enough or so… to minimize the attack surface even more.

But I agree that “disabling” the UI for some seconds or so would already be a decent solution for now.

Ein-Tim:
@fynngodau Can you confirm that this issue is resolved, now with the fix #2565 merged?

Ein-Tim:
@dsarkar I suggest to close this issue now.

fynngodau:
Thanks for implementing this, @Crystalskulls.

PR #2565, opened 2022-03-23
Disable create button on Event QR-Code Page until the qr form changes

Crystalskulls:
Fix for Issue #1137

After generating a QR Code, the 'create' Button will be disabled until the qr form changes. This prevents generating of different QR Codes for the same input data. Same behaviour for the multi-qr-code-generator.

The form of the single-qr-code-generator and the form of the multi-qr-code-generator had the same ID. This means, that changes to form fields of the single-qr-code-generator affects the buttons of the multi-qr-code-generator. So i decided to change the form ID of the multi-qr-code-generator to prevent side effects.


dsarkar:
@Crystalskulls Thanks again... !
@brifemu Please review and test.

Crystalskulls:
@dsarkar It's a pleasure.

PR #2554, opened 2022-03-18
Add onlinelibrary.wiley.com to allowlist

brifemu:
Due to issue #2490, I had included 'https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1' into the current allow list.

The current allow list could bring us problems in the future because we are not controlling an exact error code and if for example tomorrow this site is closed we would not get notices because the current allow list doesn't difference between 404 or 503 error.

I will create another PR in the next weeks about this, after finish the new FAQ and more pending stuff.

---
Internal Tracking ID: [EXPOSUREAPP-12411](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12411)

Issue #2490, opened 2022-02-26
Cypress fails testing link to Wiley Online Library


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/blog/2021-08-05-statistictiles/ "Explanation of the statistic tiles in the Corona-Warn-App"
- https://www.coronawarn.app/de/blog/2021-08-05-statistictiles/ "Erläuterung der Statistikkacheln in der Corona-Warn-App"

## Describe the issue

The Cypress test `check_links.js` fails with an HTTP 503 error (Service Unavailable) when testing the link to https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1.

```
[RESPONSE 503] https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1 on 'http://localhost:8000/de/blog/2021-08-05-statistictiles' 
[RESPONSE 503] https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1 on 'http://localhost:8000/en/blog/2021-08-05-statistictiles'
````

The site https://onlinelibrary.wiley.com appears to be now using DDoS protection by [Cloudflare](https://www.cloudflare.com/5xx-error-landing/). This is a new situation. Previously checks to this website were not causing an error response.

## Steps to reproduce

Open a web browser and check that the link https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1 is available when accessed manually.

Execute
`npm run test:open`
In Cypress console select `check_links.js` to execute this test.

Note 2 failed tests and examine contents of log file
`cwa-website/cypress/logs/broken_links_result.txt`

## Suggested change

Options:

1. Ignore HTTP 503 (Service Unavailable) as a hard error. Instead, treat HTTP 503 as a temporary error and show success, or
2. Do not test https://onlinelibrary.wiley.com

---
Internal Tracking ID: [EXPOSUREAPP-12411](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12411)

MikeMcC399:
Perhaps @brifemu could look at the best way to resolve this issue?

brifemu:
Hi @MikeMcC399  it is a pleasure to be able to communicate with you again after such a long time. I have created a PR for this problem, it is a temporary PR. I have left a small comment in the PR about how to fix this error in the future. Right now I am quite busy with several projects, so I will personally take care of this in depth in a few weeks.

#2554

[...]

PR #2434, opened 2022-02-16
FAQ page rework

brifemu:
In this PR I will include all the features where we were working lately.

- FAQ home page includes:

1. Search bar with topic filter (including glossary).
2. Section with topic and subtopics filter.
3. Link to Glossary.

image

- FAQ results page includes:

1. Right filter menu by topic and subtopic (with and without search). 
2. Search bar to make or modify a FAQ search (including glossary). 
image
Screenshot 2022-02-16 at 13 57 05
3. Breadcrumbs under search bar to navigate inside or outside a search result. 
image
4. Clickable titles and subtitles with same functionality as right filter menu.
5. In case of search: 
6.1 Highlighted subtopics on the right filter menu
image
6.2 Show the count results found in the breadcrumbs
image
6.3 Clean button on search input to reset the search.
image
6.4 Highlighted searched word in FAQ title, FAQ content and Glossary. 
image

**Page without a search:**
image

**Page with a search:**
image

PR #2574, opened 2022-03-25
Reapply checklinks exception for wiley.com

MikeMcC399:
- This PR resolves the regression in PR https://github.com/corona-warn-app/cwa-website/pull/2434 which reverted the fix from PR #2554.
- The re-opened issue https://github.com/corona-warn-app/cwa-website/issues/2490 is resolved by this PR.

After applying this PR the Cypress test `check_links.js` no longer fails with

```
[RESPONSE 503] https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1 on 'http://localhost:8000/de/blog/2021-08-05-statistictiles' 
[RESPONSE 503] https://onlinelibrary.wiley.com/doi/abs/10.2307/3315826.n1 on 'http://localhost:8000/en/blog/2021-08-05-statistictiles'
```

in `broken_links_result.txt`. Instead all tests pass.

## Verification

```
npm run test:prepare
npm run test:open
```
select `check_links.js` in Cypress console

Ensure that all tests pass and that there are no failures i.e. -- next to red X.

image

### DIFF
--- a/cypress/integration/check_links.js
+++ b/cypress/integration/check_links.js
@@ -33,7 +33,8 @@ context("Check for broken links", () => {
                   '/en/sign-language/'
                 ]
     const allowlist = [
-      'https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999'
+      'https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999',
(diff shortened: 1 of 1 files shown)


dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks!

PR #2576, opened 2022-03-25
Adding spans to define english terms to fix german screen reader pronunciation

ibexad:
The update will add a `<span>` defining English languages for some English terms. Making it easier for german screen readers to pronounce some terms better.


svengabr:
Due to huge amount of non related changes, PR will be closed

PR #2578, opened 2022-03-25
creating english-texts.json file and replacing the texts by gulp func…

ibexad:
…tion


svengabr:
The update will add a <span> defining English languages for some English terms. Making it easier for german screen readers to pronounce some terms better.

svengabr (approved):

PR #2581, opened 2022-03-25
FAQ delta_submission

dsarkar:
EN
image

DE
image

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -2607,6 +2607,14 @@
                         "indexed": true,
                         "accordion": [
                             {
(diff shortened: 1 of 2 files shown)


vaubaehn:
@dsarkar 
Very nice, thank you!
Just to prevent misunderstandings - the link to Google Play Store review in the other issue https://github.com/corona-warn-app/cwa-app-android/issues/4753#issuecomment-1079172676: At the time of writing I found that you already took care of the reviews, so I felt to submit an extra-thank-you 😉

PR #2583, opened 2022-03-25
include json files while parsing version numbers

Crystalskulls:
Fix for Issue 'Missing substitution in duplicated FAQ #2575'

The function **replaceVersionNumbers()** in the gulp workflow considers only html files in the public directory. But the duplicate FAQ entries are not placed as html files in the public directory at compile time. So we have to include also json files to catch all placeholders at compile time.


dsarkar:
@Crystalskulls Thanks for this PR.
@brifemu Please review.

brifemu:
@dsarkar / @svengabr LGTM. Tested and working. Thank you so much @Crystalskulls !

PR #2615, opened 2022-03-29
R 2.20 FAQ update admission_policy

dsarkar:
DE
image

EN
image

Internal Tracking ID: [EXPOSUREAPP-12289](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12289)


svengabr (approved):
Contextual changes approved. The text states `March 30`, which would be tomorrow.

Issue #2601, opened 2022-03-28
FAQ search input boxes do not resize dynamically


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/faq/

## Describe the issue

If the FAQ https://www.coronawarn.app/en/faq/ page is reduced from wide to narrow, then the boxes are not reorganized and the content is not readable.

## Steps to reproduce

1. Open https://www.coronawarn.app/en/faq/ in Google Chrome Desktop using full-screen width

![FAQ search 1](https://user-images.githubusercontent.com/66998419/160383030-7545fff3-ca37-405a-b404-460c49f5d5c1.jpg)

2. Drag the side of the browser window to reduce the width to the minimum

![FAQ search 2](https://user-images.githubusercontent.com/66998419/160383064-ee215a86-1942-4b98-9956-19d05bee41f1.jpg)

3. Note that the hint text in the search field is no longer readable, nor is the default text in the category field readable.

## Suggested change

If the window is dynamically resized, then adjust the placement of the search boxes and button accordingly, so the contents is always readable.

## Workaround

Press F5 (refresh) to redraw the screen.

![FAQ search 3](https://user-images.githubusercontent.com/66998419/160383089-7a442b48-cbcd-45b8-8024-332a4d37bf36.jpg)

svengabr:
Closed as I am not fine with the changes made to the layout. 

The changed version looks like this on mobile:
![2022-03-30 14_00_04-Open-Source-Projekt Corona-Warn-App – FAQ](https://user-images.githubusercontent.com/4686874/160837988-301cba54-d2b3-480e-8e00-933e9b92dc6a.png)

I expect a mobile-first solution like this:
![2022-03-30 14_28_00-Open-Source-Projekt Corona-Warn-App – FAQ](https://user-images.githubusercontent.com/4686874/160838088-0316a796-88a0-4d98-b9f3-3a9d50ab883d.png)

MikeMcC399:
@svengabr 

- I also tried out the PR #2638 before you closed it and I found it did not solve the issue. I was just about to re-check that I had not made a mistake when I tested it, but you got there first!

MikeMcC399:
- PR https://github.com/corona-warn-app/cwa-website/pull/2646 is resolving the issue for https://www.coronawarn.app/en/faq/results/.

PR #2621, opened 2022-03-29
Fix FAQ home page responsive of search's form

10delin:
Due to issue #2601, I delete the matchmedia condition in app.js and delete bootstrap classes in FAQ home page then I add a new 3 ids on and a media query in faq.scss. 
This is the issue's url : https://github.com/corona-warn-app/cwa-website/issues/2601

There is a video with the changes: https://user-images.githubusercontent.com/56647378/160608078-aedf1f56-3f20-4f53-a69b-c5394b731db8.mp4

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -360,13 +360,6 @@ $(document).ready(function(){
     if (document.querySelector(".page-faq")) {
         const searchForm = document.getElementById("faq-search-form");
         const { hash } = window.location;
-
-        if(window.matchMedia("(max-width: 767px)").matches) {
-            $("#faq-search-form").removeClass("w-50").addClass("w-100");
-            $("#faq-search").removeClass("w-50").addClass("w-100").removeClass("mr-3");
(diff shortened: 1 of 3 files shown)


MikeMcC399:
@10delin 

This is better than it was, however it is not quite fixed.

The first screenshot is from shrinking the width with no search text entered. In this case the hint text is truncated. F5 (refresh) recovers from the situation.

image

The second screenshot is from shrinking the width with a search text already entered. A separate problem here is that the ❎ is on the edge of the box. Also in this case F5 will redraw with a better layout.

image

10delin:
@MikeMcC399 This changes was in FAQ's home page, however we will do it in the FAQ's result page tomorrow in a Pull Request.

MikeMcC399:
@10delin 
> we will do it in the FAQ's result page tomorrow in a Pull Request.

Thank you. I will look out for a new PR.

svengabr:
I will close this PR in favor of https://github.com/corona-warn-app/cwa-website/pull/2646 where you already changed the detail FAQ screen. Please apply the same changes to the home page as well using bootstraps grid system. Thank you very much!

Issue #2594, opened 2022-03-27
FAQ search result not expandable


MikeMcC399:
## Where to find the issue

FAQ search

https://www.coronawarn.app/en/faq/results/?search

## Describe the issue

Search results cannot be expanded on a mobile device in portrait orientation in order to read the text.

## Steps to reproduce

1. Start Google Chrome (Desktop Version 99.0.4844.84 (Official Build) (64-bit))
2. Press F12 for Developer tools
3. Use device toolbar to select Samsung Galaxy A51/71 (412 x 914)
4. Enter https://www.coronawarn.app/en/faq/ into browser address field
5. Enter "real-time" into FAQ search field and click "Search"
6. Note that there are two results found

![real-time](https://user-images.githubusercontent.com/66998419/160279397-6cf0c6bc-4f91-41dd-9eba-c892b4dad739.jpg)

7. Click on "App Features" to expand the section
8. Click on "Why do I receive a risk warning several days after the risk encounter?"

![real-time expanded](https://user-images.githubusercontent.com/66998419/160279410-fdd83191-20b7-4f1c-81c5-460ca3587a5b.jpg)

9. Note that nothing happens

## Expected results

After clicking on "Why do I receive a risk warning several days after the risk encounter?" the FAQ article should be expanded and be readable.

## Workaround

Change the browser orientation to landscape before attempting to search the FAQs.

MikeMcC399:
- This is actually an unintentional duplicate of https://github.com/corona-warn-app/cwa-website/issues/2592. I suggest to leave it though because there are more detailed repro steps and screenshots.

Ein-Tim:
No need for the https://github.com/corona-warn-app/cwa-website/issues?q=is%3Aopen+label%3Aduplicate+sort%3Aupdated-desc label here. I closed my issue #2592.

MikeMcC399:
@brifemu said in https://github.com/corona-warn-app/cwa-website/pull/2606#issuecomment-1080813049 that this issue will be covered in another PR.

brifemu:
Hello everyone! A short summary today about this issue, we have found that the error is generated by the functionality of highlight words. Once we have found the source, we hope to have this resolved ASAP.

brifemu:
@MikeMcC399 Morning! FYI: https://github.com/corona-warn-app/cwa-website/pull/2629

MikeMcC399:
@brifemu 

Good morning!
[...]

PR #2629, opened 2022-03-30
Fix FAQ search result on mobile after search

brifemu:
Due to issue #2594 I created this PR. The problem was that the content was moved to the mobile container before highlighting. Moving the content after the highlighting, this problem is solved.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -352,6 +352,32 @@ $(document).ready(function(){
                         }
                     }
                 })
+                if(window.matchMedia("(max-width: 767px)").matches) {
+                    $("#faq-container-mobile").removeClass("d-none")
+                    $("#glossary-container-mobile").removeClass("d-none")
+                    //Adjust form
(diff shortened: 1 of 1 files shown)


MikeMcC399:
- The article can be opened, which solves the issue, however it introduces another issue that too many sections are shown. In the steps to reproduce from #2594 only two results should be found. The other sections, starting with "Troubleshooting", have no matches and should not be shown:

image

brifemu:
@MikeMcC399 Thanks Mike for your quick reply! I'm starting to research that issue, this will be covered in this PR

brifemu:
With this last update the issue is fixed, the issue was that as in the mobile view there is a container inside another, the code was confused and also detected the first dropdown and that should not happen. With this change, it only analyzes the correct dropdown, the one inside the topic.

MikeMcC399:
@brifemu 

I tried out the changes and it looks good now.
Thank you!

svengabr (approved):

PR #2639, opened 2022-03-30
Use term recycle bin in #dcc_replacement

MikeMcC399:
This PR changes the term "bin" to "recycle bin" in https://www.coronawarn.app/en/faq/results/#dcc_replacement which is then in alignment with the term as used in the app.

image

See https://www.coronawarn.app/en/screenshots/#android_recycle_bin

image


MikeMcC399:
Perhaps @HannaHei / @svengabr could take a look?

svengabr (approved):

svengabr:
Looks good, thanks @MikeMcC399 !

PR #2642, opened 2022-03-30
Fixed typo in english version of newly released 2.20 blog post

### LINKED ISSUE #2640: Incomplete sentence in 2.20 blog (EN) for certificate reissue
## Where to find the issue

https://www.coronawarn.app/en/blog/2022-03-30-cwa-2-20/
First paragraph.

## Describe the issue

The following sentence is cut off before it is complete:

"In addition, since March 29, 2022, users will be able to issue new vaccination certificates directly via the"

image

## Suggested change

image

Suggest to change:

1. "issue new" to "reissue" and reflect the fact that the user only requests a reissue. It is RKI, not the user, which does the issuing.
2. complete the sentence with "app."

so the sentence becomes:

"In addition, since March 29, 2022, users will be able to request vaccination certificates to be reissued directly via the app."

svengabr:
Closes #2640

### DIFF
--- a/blog/2022-03-30-cwa-2-20/index.md
+++ b/blog/2022-03-30-cwa-2-20/index.md
@@ -7,7 +7,7 @@ author: CWA-Team, 10 am
 layout: blog
 ---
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
It should say that the user **requests** the certificate to be reissued. Compare to German "ausstellen lassen" and not "ausstellen".

svengabr:
There seems to be an increase in typo spelling errors lately...
Thanks for your support @MikeMcC399 !

PR #2644, opened 2022-03-31
WCAG fixes

browkin:
- Fixed several duplicate ids across the website
- Removed not needed attributes and added necessary attributes for accessibility (e.g. alt="", aria-hidden="true")
- Closed a lot of <inputs>
- Added css class sr-only for navbar links "sign language" and "simple language" and a <span> for screenreader functionality


src/partials/page-analyse.html
```diff
@@ -193,13 +193,13 @@ <h3 class="analyseBoard-title">{{title}}</h3>
 							{{/if}}
 
 							<button class="btn analyseBoard-iconBtn analyseBoard-expand" data-id="chart{{@index}}">
-								<img src="{{../root}}assets/img/icons/expand.svg" aria-label="{{expand}}">
-								<img src="{{../root}}assets/img/icons/collapse.svg" aria-label="{{collapse}}">
+								<img src="{{../root}}assets/img/icons/expand.svg" alt="" aria-hidden="true">
```
svengabr, line 196: Buttons are now completely unusable by blind people. Either add an aria-label to the <button> or add a .sr-only <span> with a description inside

svengabr (approved):

PR #2648, opened 2022-03-31
Reponsive of FAQ's form fixed

10delin:
Responsive of FAQ's form fixed with bootstrap grid


MikeMcC399:
- Is this PR supposed to solve https://github.com/corona-warn-app/cwa-website/issues/2601?

svengabr (approved):

svengabr:
Yes exactly @MikeMcC399 

I made one addition in my latest push, made the box on mobile bigger.

**Before**

before

**After**

after

MikeMcC399:
@svengabr 

The box width of /en/faq/ looks good when the window width is dynamically decreased, however at full width screen on a desktop the box does not fully display the text.

image

svengabr:
Yes, you are right. I made another small adjustment and I think it turned out much better. More space now.

en
de

MikeMcC399 (approved):
I can't see any more issues when re-sizing, so it looks good to go! 👍🏻

svengabr (approved):

Issue #2652, opened 2022-03-31
Search result counter is flapping if you click on a section header


Crystalskulls:
## Where to find the issue

https://www.coronawarn.app/en/faq/
https://www.coronawarn.app/de/faq/

## Describe the issue
The search result counter in the FAQ page is flapping. If you search for something and click on a section header the counter shows the correct number for a second and then shows an incorrect number.
<img width="1552" alt="flapping_search_result" src="https://user-images.githubusercontent.com/25111875/161119499-5c90157c-39fa-4446-a096-632e62e1bdf8.png">

## Steps to reproduce
1. Open https://www.coronawarn.app/en/faq/
2. Type **rule** in the search field, category **All**
3. Click on the search button
4. Click on the section header **Technical Contenxt** in the side menu
5. Notice, that the string **Search result: 2 results found** is present for a second and than changes to **Search result: 5 results found**

## Suggested change
I would like to create a PR for this Issue

PR #2654, opened 2022-03-31
use current topic in on-click event listener for section headers

Crystalskulls:
Fix for Issue #2652

The on-click listener of the section headers use the URL parameter **topic** for updating the results, but if you click on a section-header the URL parameter doesn't changes to the clicked topic. So if somebody search in **All** topics and then click on a section header the result counter is calculated wrong.

So i pass the clicked topic the the **updateResult()** funtion.


dsarkar (approved):

dsarkar:
Hi @Crystalskulls And again: Thanks a lot for this fix PR! Best wishes, DS

Issue #2613, opened 2022-03-28
Cypress check_anchor_links.js misses issues in FAQ


MikeMcC399:
## Where to find the issue

[cypress/integration/check_anchor_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_anchor_links.js)

## Describe the issue

The Cypress test [cypress/integration/check_anchor_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_anchor_links.js) does not find anchor link issues in the new FAQ structure under https://www.coronawarn.app/en/faq/results/ and https://www.coronawarn.app/de/faq/results/

## Steps to reproduce

Manually introduce a bad anchor link into `/src/data/faq.json`.

`npm run test:open`
Select `check_anchor_links.js`
Note that the bad anchor link is not found

## Suggested change

Add the directories:

- `/en/faq/results/` and
- `/de/faq/results/`

to 

[cypress/integration/check_anchor_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_anchor_links.js)

This is however not sufficient as then the test fails with a syntax error:

```
  1) Check for broken anchor links
       "/de/faq/results/" - Check for broken anchor links:
     Error: Syntax error, unrecognized expression: #
```

MikeMcC399:
This issue needs assistance from the Open Source Team (@brifemu / @larswmh ?) regarding the syntax error preventing an update to cover testing the new directories `/faq/results`.

ahodzic2:
@MikeMcC399 I can try to take a gander at it and see if I can assist. Not exactly my cup of tea either but I've got some time later and a willingness to try/learn ;)

MikeMcC399:
@ahodzic2 
If you learn Cypress you will not be alone. See https://www.npmtrends.com/cypress for download trends.

ahodzic2:
@MikeMcC399 I was taking a look at it yesterday and I suspect that it may be having an issue with a particular anchor. From the test results it does successfully get and test a number of anchor links under en/faq/results/ and de/faq/results but then throws that syntax error.

Will continue to play around with it today/tomorrow and see if I can pinpoint the issue.

MikeMcC399:
@ahodzic2 
You are right that it successfully executes some tests before it stops with the syntax error. 
It appears that it cannot deal with the html code `<a href='#'`.

ahodzic2:
[...]

PR #2656, opened 2022-04-01
Update check_anchor_links.js for new FAQ

ahodzic2:
Updated check_anchor_links.js to allow for testing of "/en/faq/results/" and "/de/faq/results/".
See https://github.com/corona-warn-app/cwa-website/issues/2613#issuecomment-1085418447

image


MikeMcC399:
@svengabr / @brifemu 

It would be great if you could take a look at this PR. Without it, we are currently not checking the anchor links in the FAQ articles after they moved down a level from `/faq` to `/faq/results`.

larswmh (approved):

dsarkar:
@ahodzic2 Many thanks for the PR.

Issue #16, opened 2020-05-30
[Discussion] Header/footer: color contrast ratio is too low


zichy:
## Where to find the issue
The issue can be found in the header and footer on every site.

## Describe the issue
The contrast ratio of foreground/background color is too low (1.77:1). This makes it hard for some people with visual impairment to read the text. According to [WCAG 2.1](https://www.w3.org/TR/WCAG21/#contrast-minimum), it should be at least 4.5:1 for normal text and 3:1 for large text.

See: https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=80CDEC

The actual contrast ratio is hard to determine since there's a color gradient. The ratio of white text on red background would be fine (6.27:1), but almost all of the content area is in brighter colors.

## Suggested change
Either background gradient or text should be in darker colors. Maybe it would be possible to use text shadow or a solid background color for some parts of the header/footer, e.g. main navigation.

MalteJ:
Is this just about the Corona-Warn-App logo? Every other asset has a higher contrast as far as I can see.

zichy:
It's about the other text as well. As I said before, the ratio is hard to determine since it's a color gradient. The ratios differ from 1.77:1 to 6:27:1 and depend on the viewport as well.

mojoaxel:
I think the is the most problematic part:
![image](https://user-images.githubusercontent.com/600565/86393265-136cb480-bc9d-11ea-8a6a-98ca38534027.png)

svengabr:
The colors for linear gradients were adjusted with the latest PR https://github.com/corona-warn-app/cwa-website/pull/2660 

The team is currently actively working on resolving all found WCAG issues.

PR #2660, opened 2022-04-01
changed background gradient to meet contrast requirements

svengabr:
Changed background gradient to meet contrast requirements affecting the header, footer and some design elements in the middle of the web page.

My used colors were approved by the UX team.

Before
before

After
after

### DIFF
--- a/src/assets/scss/_custom-variables.scss
+++ b/src/assets/scss/_custom-variables.scss
@@ -4,7 +4,7 @@ $theme-button-colors: (
 ) !default;
 $theme-button-shadow-color: $charcoal-black;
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@svengabr 
- Does this resolve the old issue https://github.com/corona-warn-app/cwa-website/issues/16?

svengabr:
This will increase the minimum contrast to 1:3 which should be the minimum required amount. I will link it to this very old issue which weirdly I can't even select since the issue is too old (I guess)

svengabr:
Closes https://github.com/corona-warn-app/cwa-website/issues/16

Issue #2661, opened 2022-04-01
Removal of "Questions and answers" articles


MikeMcC399:
In the previous FAQ layout there were several FAQ articles which were links to other articles:

- [/en/faq/#vac_cert_index](https://www.coronawarn.app/en/faq/results/#vac_cert_index) "Questions and answers about the vaccination certificate"
- [/en/faq/#vac_cert_booster](https://www.coronawarn.app/en/faq/results/#vac_cert_booster) "Questions and answers about the booster vaccination and notifications regarding my vaccination certificate"
- [/en/faq/#test_cert_index](https://www.coronawarn.app/en/faq/results/#test_cert_index) "Questions and answers about the test certificate"
- [/en/faq/#check_in_index](https://www.coronawarn.app/en/faq/results/#check_in_index) "Questions and answers about the event registration"

These link articles have been removed and the corresponding anchors no longer work.

Was there a decision to drop these anchors without preserving backwards compatibility, such as by providing a redirect? If not, should they be put back in as redirects, for instance?

- This is related to the suggested changes in PR https://github.com/corona-warn-app/cwa-website/pull/2657.

---

![vac_cert_index](https://user-images.githubusercontent.com/66998419/161232488-142fd96e-ded5-4b92-af07-d44bb9edcbbe.jpg)

![vac_cert_booster](https://user-images.githubusercontent.com/66998419/161232518-865c1b2d-54c0-4638-ad8c-5ade315ec2ee.jpg)

![test_cert_index](https://user-images.githubusercontent.com/66998419/161232549-953d2385-c47c-4c1d-a6c3-0556a9e9933f.jpg)

![check_in_index](https://user-images.githubusercontent.com/66998419/161232577-6fc253c5-8e11-4204-9016-19e2121cd10a.jpg)

MikeMcC399:
- A related issue https://github.com/corona-warn-app/cwa-website/issues/2666 has been resolved.

MikeMcC399:
## Suggested change

To cover the case where the FAQ Questions and Answers anchors have been bookmarked external to the cwa-website, I suggest adding FAQ redirects for the following removed anchors:

Removed Q&A anchor | Equivalent in new FAQ layout
--- | ---
#vac_cert_index |  #vac_cert
#vac_cert_booster | #vac_cert
#test_cert_index | #test_cert
#check_in_index | #check_in

MikeMcC399:
- PR #2681

[...]

PR #2681, opened 2022-04-04
Add FAQ Q&A index redirects

MikeMcC399:
- This PR implements the suggestion from https://github.com/corona-warn-app/cwa-website/issues/2661#issuecomment-1087690698

The following FAQ redirects are added
From anchor | To anchor
--- | ---
#vac_cert_index |  #vac_cert
#vac_cert_booster | #vac_cert
#test_cert_index | #test_cert
#check_in_index | #check_in

This means that for instance a bookmark https://www.coronawarn.app/en/faq/#vac_cert_index from the old FAQ format will redirect to https://www.coronawarn.app/en/faq/results/#vac_cert.

The original anchors were not carried over from the old FAQ format to the new FAQ format, because the corresponding content was deleted, so this ensures that any external bookmarks using these links will continue to direct the user to the most appropriate content.


larswmh (approved):

dsarkar:
@MikeMcC399 Thank you!

Issue #2670, opened 2022-04-04
Reduce "14" to "10" where it corresponds to the days of the risk calculation


DerVogel2020:
## Where to find the issue
e.g. https://www.coronawarn.app/de/faq/results/#new_device :
Falls Sie innerhalb der 14 Tage auf COVID-19 getestet wurden, tragen Sie den PCR-Test-QR-Code auf Ihrem neuen Gerät ein. Ist das Testergebnis positiv, sollten Sie das Testergebnis auf beiden Geräten teilen. Für das Teilen eines Testergebnisses auf mehreren Geräten beachten Sie bitte den FAQ Eintrag [Wie teile ich ein positives Testergebnis auf mehreren Geräten?](https://www.coronawarn.app/de/faq/results/#test_multiple_devices).

Nach 14 Tagen enthält Ihr altes Gerät keine relevanten Einträge für die Risiko-Ermittlung mehr. Sie brauchen den Risikostatus auf diesem Gerät dann nicht weiter zu beachten und können die Corona-Warn-App dort entfernen.

## Describe the issue
Only 10 days are relevant now.

## Suggested change
Change everywhere "14 days" to "10 days" (of course in both languages).

MikeMcC399:
@DerVogel2020 

Thanks for bringing this up!

- https://www.coronawarn.app/en/faq/results/#new_device unfortunately was not changed at all.
- https://www.coronawarn.app/de/faq/results/#new_device was already changed in the first paragraph:

"Damit Sie trotzdem über mögliche Risiko-Begegnungen informiert werden, empfehlen wir Ihnen, den Risikostatus auf dem alten Gerät noch mindestens 14 Tage (10 Tage ab Version 2.20) zu beobachten, nachdem Sie die Corona-Warn-App auf dem neuen Gerät installiert haben."

So I suggest that instead of changing 14 days to 10 days, it needs to list both. To simplify the text, the first mention defines the version differences and thereafter write `14 or 10 days` (EN)  `14 bzw. 10 Tage` (DE) without mentioning the version.

amouzayanni:
@MikeMcC399

Hello! I'm working with @ahodzic2 and I'd like to resolve this issue if you'd like it done.

MikeMcC399:
@amouzayanni 

@DerVogel2020 has contributed in the past, so might also want to provide a PR. Other than that I suggest waiting for feedback from the Open Source Team. I'm not the one to decide, since I am a community contributor with no official capacity.

[...]

PR #2691, opened 2022-04-05
updated days of risk calculation

amouzayanni:
This PR resolves Issue #2670 "Reduce "14" to "10" where it corresponds to the days of the risk calculation" by updating mentions of the number of days for risk calculation in the FAQ to include the number of days used since version 2.20.

The following FAQ entries have been updated:
English FAQ:
[What do I have to consider when switching to a new device?](https://www.coronawarn.app/en/faq/results/#new_device)
[Why is voluntary self-quarantine recommended?](https://www.coronawarn.app/en/faq/results/#voluntary_self_quarantine)
[I received the message that I had a risk encounter. The next day it was gone again. Shouldn't this be deleted after 14 days?](https://www.coronawarn.app/en/faq/results/#encounter_deletion)
German FAQ:
[Was muss ich beim Wechsel auf ein neues Gerät beachten?](https://www.coronawarn.app/de/faq/results/#new_device)
[Warum wird die freiwillige Selbstquarantäne empfohlen?](https://www.coronawarn.app/de/faq/results/#voluntary_self_quarantine)
[Ich habe die Meldung bekommen, dass ich eine Risiko-Begegnung hatte. Am nächsten Tag war sie wieder weg. Sollte das nicht erst nach 14 Tagen gelöscht werden?](https://www.coronawarn.app/de/faq/results/#encounter_deletion)


src/data/faq.json
```diff
@@ -1582,10 +1582,11 @@
...
-                                    "After 14 days, your old device will no longer contain any relevant entries for risk assessment. You don't need to pay any further attention to the risk status on this device and can remove the Corona-Warn-App there."
+                                    "If you are tested for COVID-19 during these 14 days (10 days since Version 2.20), enter the PCR-Test QR code on your new device. If the test result is positive, you should share the test result on both devices. For sharing a test result on multiple devices, please note the FAQ entry <a href='#test_multiple_devices' target='_blank' >How do I share a positive test result on multiple devices?</a>.",
+                                    "If you are using another compatible COVID-19 app which holds more than 10 days on your old device, you should keep the old device for 14 days regardless of your version of the Corona-Warn-App on your new device.",
```
Ein-Tim, line 1588: I did not yet translate this into German, as I'm unsure if this is necessary.

Feedback from @MikeMcC399 and the Open-Source-Team is welcome!

MikeMcC399: @Ein-Tim 

I think this should be left out. I will post my reasons in the issue. Thanks for questioning this!

src/data/faq.json
```diff
@@ -2696,7 +2697,7 @@
                                 ]
                             },
                             {
-                                "title": "I received the message that I had a risk encounter. The next day it was gone again. Shouldn't this be deleted after 14 days?",
+                                "title": "I received the message that I had a risk encounter. The next day it was gone again. Shouldn't this be deleted after 14 or resp. 10 days?",
```
MikeMcC399, line 2700: remove "resp." which is a German use of the word.

In English "respectively" is used to link two lists and it is not abbreviated like in German. For example you could say Version 2.19 and Version 2.20 use 14 and 10 days respectively. See https://dictionary.cambridge.org/dictionary/english/respectively for explanation.

MikeMcC399 (changes requested):
See in-line comments

src/data/faq.json
amouzayanni: I can handle this deletion if there aren't any objections.

src/data/faq.json
MikeMcC399: @amouzayanni 
I would recommend simply: "Shouldn't this be deleted after 14 or 10 days?"

Adding the version information in the title makes it more complicated. The text below explains about the version differences.

amouzayanni:
Just made the suggested changes. Please let me know if there are any issues.

MikeMcC399 (approved):
I have no more change requests for the text.

Please mark the reviews as resolved.

dsarkar (approved):

dsarkar:
Many thanks to @amouzayanni @Ein-Tim and @MikeMcC399 ! @amouzayanni Congrats to your first PR in this project!

Issue #2668, opened 2022-04-03
Missing FAQ anchor in browser address bar


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/faq/results/

## Describe the issue

The browser address bar shows only `https://www.coronawarn.app/en/faq/results/` after accessing a specific FAQ article.

This is a regression from the previous FAQ layout behaviour before PR #2434.

## Steps to reproduce

1. In Google Chrome open a new tab
2. Paste `https://www.coronawarn.app/en/faq/results/#where_to_get` into the address bar and press Enter
3. Note that the expected article [Where can I get the app?](https://www.coronawarn.app/en/faq/results/#where_to_get) appears. The address bar shows `https://www.coronawarn.app/en/faq/results/`.

![URL displayed](https://user-images.githubusercontent.com/66998419/161420936-c095f5ad-7b39-4c4b-8d14-b1bbddd311b1.jpg)

Compare this with the behaviour when accessing:

- [#howto](https://www.coronawarn.app/en/#howto)
- [README.md#about-this-repository](https://github.com/corona-warn-app/cwa-website/blob/master/README.md#about-this-repository).

FAQ previous layout:

- Also compare with the web behaviour before PR #2434 was applied.

![URL displayed V1](https://user-images.githubusercontent.com/66998419/161420945-29d8063f-dbc4-42a4-9d13-dc86df8bc2d9.jpg)

## Expected behaviour

After navigating to a specific FAQ article the browser address bar should show the complete link to the article, including the anchor.

In the above steps to reproduce, the address bar should show `https://www.coronawarn.app/en/faq/results/#where_to_get` and not just `https://www.coronawarn.app/en/faq/results/`.

MikeMcC399:
- PR #2707 fixes the issue according to the steps to reproduce.

PR #2707, opened 2022-04-07
Added FAQ anchor to url

browkin:
Adressing Issue #2668 by @MikeMcC399 

The browser address bar shows only https://www.coronawarn.app/en/faq/results/ after accessing a specific FAQ article.
I´ve added it back, so it behaves like the old FAQ.


larswmh (approved):

dsarkar (approved):

MikeMcC399 (approved):
- I checked with the repro steps from https://github.com/corona-warn-app/cwa-website/issues/2668 and this PR fixes the issue.

dsarkar:
@MikeMcC399 Thanks for feedback/review.

Issue #2714, opened 2022-04-08
Allow Cypress test suite to run on https://www.coronawarn.app without errors


MikeMcC399:
## Current Implementation

The [Cypress test suite](https://github.com/corona-warn-app/cwa-website/tree/master/cypress) for the [cwa-website](https://github.com/corona-warn-app/cwa-website/) is designed to run on a local pre-production instance of the web running on http://localhost:8000. See
https://github.com/corona-warn-app/cwa-website/blob/64d4f24367e85ca7091064e25a0fdf2975df6423/cypress.json#L2

It is possible to override the `baseUrl` parameter to request that the production web at https://www.coronawarn.app is tested, however one of the six tests fails, so the complete test is marked as unsuccessful.

Executing `npx cypress run -c baseUrl=https://www.coronawarn.app` from the locally cloned root directory of [cwa-website](https://github.com/corona-warn-app/cwa-website/) results in an error from the test [/cypress/integration/blog_spec.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/blog_spec.js). All the other tests run successfully.

```
       Spec                                              Tests  Passing  Failing  Pending  Skipped
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ √  applink_spec.js                          00:02        3        3        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ×  blog_spec.js                             00:06        2        -        2        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ √  check_anchor_links.js                    00:16       37       37        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ √  check_links.js                           08:13       37       37        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ √  hotline_spec.js                          668ms        1        1        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ √  pages/eventRegistration.js               00:14        6        6        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ×  1 of 6 failed (17%)                      08:55       86       84        2        -        -
```

[...]

PR #2719, opened 2022-04-08
Fix errors Cypress test suite

10delin:
Due to this issue #2714 I have fixed the issue related to the allow Cypress test suite.


10delin:
If @MikeMcC399  thinks it's too much of a stretch I'll change the js logic.

luisoos (approved):

MikeMcC399:
@dsarkar / @10delin 

I agree this is a little "brute force", however I don't think that the test page is necessary any more.

I do suggest though that all code lines which are commented out in `cypress/integration/blog_spec.js` should be deleted. The git history will always allow deleted lines to be reverted, and it is much clearer to remove them rather than leave them in, but commented out.

larswmh (approved):

dsarkar (approved):

PR #2730, opened 2022-04-11
[10] Make Event QR PDF's readables

brifemu:
With this PR, I will make Event QR PDF's readable for screen readers, replacing image's texts for real text.


svengabr:
Changes approved, thanks @brifemu !

MikeMcC399:
This PR causes the Cypress test `eventRegistration.js` to fail with the following output:
`expected '<canvas#eventqrcode.eventqr-preview>' to have attribute 'width' with the value '827', but the value was '1654'`. 

- I submitted PR https://github.com/corona-warn-app/cwa-website/pull/2740 to fix the Cypress test.

Issue #2271, opened 2022-01-01
Cypress eventRegistration test fails (canvas width)


MikeMcC399:
## Where to find the issue

Cypress test suite
[eventRegistration.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/pages/eventRegistration.js).

## Describe the issue

The Cypress test [eventRegistration.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/pages/eventRegistration.js) fails with the error:

`expected <canvas#eventqrcode.d-none.eventqr-preview> to have attribute width with the value '827', but the value was '1654'`

A related issue https://github.com/corona-warn-app/cwa-website/issues/2139 was already reported and resolved.

## Steps to reproduce

`npm run test:open`

In Cypress console select and run:

`eventRegistration.js`

## Suggested change

The Cypress test [eventRegistration.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/pages/eventRegistration.js) should be reviewed and updated to work with the current [Event Registration form](https://www.coronawarn.app/en/eventregistration/).

It looks like the changes from PR #2105 "Reintroducing multiQR functionality" broke the Cypress test.

Consider also making sure that the test covers the [single](https://www.coronawarn.app/en/eventregistration/#single) and the [multi QR code](https://www.coronawarn.app/en/eventregistration/#multi) tabs of the [Event Registration form](https://www.coronawarn.app/en/eventregistration/).

MikeMcC399:
Happy New Year to everybody!

I'm not planning to submit a PR for this. I did try to just change the width, but then another error came up:

```
#printCode
expected <button#printCode.btn.btn-secondary> not to have attribute disabled
```

It would be better for somebody who is familiar with the form and with Cypress tests to propose any changes.

brifemu:
@MikeMcC399 

Happy new year! I'm checking this and discovered the origin of

```
#printCode
expected <button#printCode.btn.btn-secondary> not to have attribute disabled
```

This because of PR #2260, the start date inserted by the test is invalid due to new date check. I'm working on it, will be a new PR soon.

dsarkar:
@MikeMcC399 @brifemu Happy new year. Thanks for report, analysis, and solution!

MikeMcC399:
@brifemu 
Thank you very much for picking this up! I wish you and @dsarkar a good start in the New Year!

MikeMcC399:
[...]

PR #2740, opened 2022-04-12
Remove QR code width check from eventRegistration.js

MikeMcC399:
This PR removes the QR width check in the Cypress `eventRegistration.js` test. Instead only the presence of the QR code is checked.

The test has proven to be flaky and has failed several times due to changes in the QR code generation. Although the code is generated correctly, the test fails.

This change in the Cypress test makes the test less sensitive to changes in the QR code generation.

It resolves the recurring issue of the test failing with the following output:
`expected '<canvas#eventqrcode.eventqr-preview>' to have attribute 'width' with the value '827', but the value was '1654'`. 

See also:
- https://github.com/corona-warn-app/cwa-website/issues/2271

The latest failure was caused by changes made in:
- PR https://github.com/corona-warn-app/cwa-website/pull/2730


larswmh (approved):

dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks.

Issue #2742, opened 2022-04-12
Overlapping text on FAQ overview (DE)


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/de/faq/

## Describe the issue

![image](https://user-images.githubusercontent.com/66998419/162962599-4d41d2fb-4658-4018-8adc-28c1a49dd873.png)

## Steps to reproduce

Open https://www.coronawarn.app/de/faq/ in web browser on desktop with wide screen.
Scroll down to "Themenbereiche"
Gradually reduce the width of the browser window and note that there are text overlaps.

## Suggested change

Add soft hyphens for long words

MikeMcC399:
If you wish, I can submit a PR to tackle this issue.

dsarkar:
@MikeMcC399 PR will be appreciated! Thanks!

MikeMcC399:
@dsarkar 

- I have submitted PR #2744 to resolve this issue.

PR #2744, opened 2022-04-12
Add soft hyphens to long DE FAQ  titles

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/2742 "Overlapping text on FAQ overview (DE)" by inserting soft hyphens into long titles using Duden rules for hyphenation of German words.

After this change, the text from one column does not run into the next, and each of the 5 columns stays at the same width compared to other columns as the width of the browser window is changed.

The narrowest view of https://www.coronawarn.app/de/faq/ which still shows columns across the page now looks like this:

image

The wide view, before any hyphenation due to narrowing of the browser window takes place, looks like this:

image

### DIFF
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -732,7 +732,7 @@
                         ]
                     },
                     {
-                        "title": "Statusnachweis",
+                        "title": "Status&shy;nachweis",
                         "id": "status_proof",
                         "indexed": true,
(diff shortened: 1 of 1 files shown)


dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks.

PR #2504, opened 2022-03-04
adding dynamic sitemap to read whole pages path and title

ibexad:

PR #2754, opened 2022-04-14
Revert old sitemap PR and create simple sitemap

larswmh:
This PR reverts #2504 and creates a less complicated sitemap. Among other things, many vulnerabilities that were caused by the previous solution have been terminated with this.
sitemap


MikeMcC399:
@larswmh 

Could you please check which version of `npm` you are using? For instance:
```
$ npm -v
8.5.0
```

The `package-lock.json` you committed has reverted to  `"lockfileVersion": 1`. It should be  `"lockfileVersion": 2`

It's best if you have Node.js 16 installed as recommended by https://github.com/corona-warn-app/cwa-website/blob/master/README.md#requirements.

MikeMcC399 (changes requested):
Testing locally there is a German footer on the English language page http://localhost:8000/en/sitemap/. The footer should be in the same language as the page itself.

larswmh:
@MikeMcC399 thanks for your feedback. 
I updated Node.js to the latest recommended version and `lockfileVersion` is now set back to 2.
The footer on the English sitemap has been changed to the English footer.

MikeMcC399 (approved):
I can't see any other issues, so LGTM!

svengabr:
Changes approved, thanks @larswmh !

PR #2755, opened 2022-04-14
Fix english imprint

larswmh:
Fixes #2752 
image

Issue #2752, opened 2022-04-14
Legal notice broken (EN) 


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/imprint/ "Legal notice"

## Describe the issue

The English-language version of the "Legal notice" https://www.coronawarn.app/en/imprint/  is broken

![image](https://user-images.githubusercontent.com/66998419/163351350-1e8fad30-9a44-4b8c-9b83-f84bf969f273.png)

The German-language version "Impressum" is not affected.

MikeMcC399:
- PR https://github.com/corona-warn-app/cwa-website/pull/2449 from @ibexad seems to have caused this issue. There is no comment text in this PR, so it is not clear what the goal was and why only the English file was changed, not the German one.

Comparing the two source files, they are formatted quite differently:

- https://github.com/corona-warn-app/cwa-website/blob/master/src/data/imprint.json (not working)
- https://github.com/corona-warn-app/cwa-website/blob/master/src/data/imprint_de.json (working)

MikeMcC399:
- PR #2755 has fixed this issue

PR #2774, opened 2022-04-19
Fix heading order

larswmh:
Follow-up to #2755 and #2752 

In order to make the pages more accessible, headings should follow a sequential, descending order, and shouldn't skip a level. With this PR, this will fix it for

-  https://www.coronawarn.app/en/terms-of-use/
-  https://www.coronawarn.app/en/imprint/
-  https://www.coronawarn.app/en/privacy/


svengabr (approved):

svengabr:
Changes approved, thanks @larswmh !

Issue #2783, opened 2022-04-19
Harmonize on app name in EN version announcement blogs


MikeMcC399:
## Where to find the issue

Recent blog announcements of new versions:

- https://www.coronawarn.app/en/blog/2022-04-19-cwa-2-21/ "Corona-Warn-App Version 2.21 enables test administration for family members"
- https://www.coronawarn.app/en/blog/2022-03-30-cwa-2-20/ "CWA 2.20: Corona-Warn-App Available in Ukrainian Language"
- https://www.coronawarn.app/en/blog/2022-03-16-cwa-2-19/ "CWA 2.19 improves the assignment of certificates to persons"

## Describe the issue

In the section of recent English language blogs the following sentence is used:

"The new version of the **Corona warning app** will be available to users here within the next 48 hours."

The corresponding German language blogs, for example https://www.coronawarn.app/de/blog/2022-04-19-cwa-2-21/, use the sentence:

"Hier steht Nutzer*innen die neue Version der Corona-Warn-App innerhalb der nächsten 48 Stunden zur Verfügung."

where the app is referred to by the name "Corona-Warn-App".

## Suggested change

Use the term "Corona-Warn-App" to refer to the app in English language version announcement blogs. This is in-line with the terminology use in the FAQs https://www.coronawarn.app/en/faq/results/, the name used in the German-language blogs and with the naming in the English language app stores e.g.  [Google Play Store - CWA - English](https://play.google.com/store/.apps/details?id=de.rki.coronawarnapp&hl=en).

So change the sentence to:

"The new version of the Corona-Warn-App will be available to users here within the next 48 hours."

MikeMcC399:
Please let me know if this change would be acceptable and I will then submit a corresponding PR.

dsarkar:
@MikeMcC399 Thanks for reporting, yes, PR would be appreciated and accepted.

MikeMcC399:
@dsarkar 

- Thanks for the approval to go ahead! I have submitted #2784 for these changes.

MikeMcC399:
The issue is now resolved.

PR #2784, opened 2022-04-19
Use standard name "Corona-Warn-App" in EN blogs

MikeMcC399:
This PR resolves issue #2783 "Harmonize on app name in EN version announcement blogs" by replacing "Corona warning app" by "Corona-Warn-App" in all English language blogs where the term is specifically referring to RKI's CWA app.

No change is made where "Corona warning app" is used in blogs in a generic sense.


dsarkar (approved):

dsarkar:
@MikeMcC399 Thanks for the fix!

Issue #2747, opened 2022-04-13
FAQ topics not vertically aligned


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/
- https://www.coronawarn.app/de/faq/

## Describe the issue

In a desktop browser the FAQ topics are shown side-by-side as if they were in a table, however the display does not behave like a table. Especially if the width of the browser is reduced, the initial alignment seen on a wide screen disappears and the entries are displayed irregularly. This is not pleasing to look at.

See the following screenshot where the entries are marked-up and linked by yellow lines:

![image](https://user-images.githubusercontent.com/66998419/163114117-2209e4a8-30bb-4f5d-bd6f-c3c00d17fb5a.png)

This topic has also been mentioned in https://github.com/corona-warn-app/cwa-website/issues/2733#issuecomment-1096762320.

## Suggested change

Align the entries vertically.

Note: I submitted this as a bug, however it may be considered as an enhancement.

brifemu:
Hey @MikeMcC399 I recorded a video with a possible solution for this. If I get your approval, I will make a PR with the fix.

https://user-images.githubusercontent.com/44208107/164683022-86542157-71fa-4e30-bfd2-8311c42bfa56.mp4

MikeMcC399:
@brifemu 

>  I recorded a video with a possible solution for this. If I get your approval, I will make a PR with the fix.

It looks promising, however it's difficult to tell from the video. Could you perhaps scroll the page down so that all the menu options are shown and record the video again? "Themenbereiche" could be at the top of the window.

brifemu:
@MikeMcC399 
I have created the PR directly so you can test it comfortably and if you like we can merge it. https://github.com/corona-warn-app/cwa-website/pull/2808

MikeMcC399:
@brifemu 

> I have created the PR directly so you can test it comfortably and if you like we can merge it.

- PR #2808 would be fine for me. The topics are always aligned vertically and the layout adapts to the width of the window.  I will put an approval in the PR and hopefully the others will agree!

MikeMcC399:
This issue has been resolved. Thanks to @brifemu for the improvements!

PR #2808, opened 2022-04-22
Adapt col width depend of the breakpoint in FAQ home

brifemu:
This PR is to give a possible fix to #2747


MikeMcC399 (approved):
I did a quick check with the English (http://localhost:8000/en/faq/results/) and German (http://localhost:8000/de/faq/results/) pages  , changing the width of the browser window.
I also simulated a mobile window size.

I couldn't see any issues with this and I think it is an improvement.

larswmh (approved):

dsarkar (approved):

PR #2819, opened 2022-04-27
changing heroku to auto updating standalone

svengabr:
Trying to fix

```
E: The repository 'https://cli-assets.heroku.com/apt ./ InRelease' is not signed.
```

in

https://github.com/corona-warn-app/cwa-website/runs/6194463380?check_suite_focus=true#step:3:59

### DIFF
--- a/.github/workflows/deploy-master.yml
+++ b/.github/workflows/deploy-master.yml
@@ -21,8 +21,7 @@ jobs:
     steps:
       - name: Install requirements
         run: |
-          curl https://cli-assets.heroku.com/apt/release.key | sudo apt-key add -
-          curl https://cli-assets.heroku.com/install-ubuntu.sh | sh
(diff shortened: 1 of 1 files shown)


MikeMcC399:
The key expired on Apr 26, 2022
See https://github.com/heroku/cli/issues
Issue 1990

I don't know whether `install-ubuntu.sh` would be preferable if it were working or whether it's ok to stay with `install.sh` permanently.

MikeMcC399:
`heroku` has fixed the issue and I was able to run the following command locally without errors:

`curl https://cli-assets.heroku.com/install-ubuntu.sh | sh`

```
heroku installed to /usr/bin/heroku
 ›   Warning: Our terms of service have changed: https://dashboard.heroku.com/terms-of-service
heroku/7.60.2 linux-x64 node-v14.19.0
```

(Just in case this is needed.)

PR #2824, opened 2022-04-29
Add FAQ #vac_status

larswmh:
image
image


Ein-Tim (commented):
Why use "vac_status" as anchor and not e.g. "vac_infofmation" or "vac_recommendations"?
IMO "vac_status" does not really fit here.

MikeMcC399:
The link in this new faq is also included in 

[#vaccination_recommendation](https://www.coronawarn.app/en/faq/results/#vaccination_recommendation) which is located in the section https://www.coronawarn.app/en/faq/results/#vaccination_recommendation

image

MikeMcC399:
@Ein-Tim 

> Why use "vac_status" as anchor and not e.g. "vac_infofmation" or "vac_recommendations"? IMO "vac_status" does not really fit here.

I agree that #vac_status does not fit the contents of the new faq.  "vac_infofmation" should probably be  "vac_information". "vac_recommendations" is too similar to the existing [#vaccination_recommendation](https://www.coronawarn.app/en/faq/results/#vaccination_recommendation) however.

larswmh:
@MikeMcC399 @Ein-Tim thanks for your quick feedback. `COVID-19 > General` truly fits better than `Troubleshooting > General` and the FAQ has been moved with https://github.com/corona-warn-app/cwa-website/pull/2824/commits/dbd784dd06117f052aa1bd155f46d8c6112b97f8.
About the anchor: It is predefined by stakeholders and can therefore not be changed. It is planned to be used in the app.

Ein-Tim:
@larswmh 

Thanks for the feedback, this doesn't make sense IMO but OK!

dsarkar (approved):

Issue #2832, opened 2022-05-04
Bad link to RATProfilesList_Empty.png in DE Screenshots 2.22


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/de/screenshots/#android_rat_profiles

## Describe the issue

The Screenshot section https://www.coronawarn.app/de/screenshots/#android_rat_profiles links to the file https://www.coronawarn.app/assets/screenshots/2-22/de/android/RATProfilesList_Empty.png however this file does not exist.

Instead there is a file https://www.coronawarn.app/assets/screenshots/2-22/de/android/RatProfilesList_Empty.png which has the same name, but different upper / lower case characters.

## Steps to reproduce

On Ubuntu execute
`npm run test:open`
select `check_links.js`

produces log:

```
==================== Broken links ====================

[RESPONSE 404] http://localhost:8000/assets/screenshots/2-22/de/android/RATProfilesList_Empty.png on '/de/screenshots/' 
```

OR

Open https://www.coronawarn.app/de/screenshots/#android_rat_profiles

Note "Schnelltest-Profile Bild 4 von 6" with link to https://www.coronawarn.app/assets/screenshots/2-22/de/android/RATProfilesList_Empty.png does not display any graphic.

![image](https://user-images.githubusercontent.com/66998419/166873403-192e68fc-4544-4e2a-ba03-3e356f35e2c8.png)

For comparison, the EN file https://www.coronawarn.app/assets/screenshots/2-22/en/android/RATProfilesList_Empty.png does exist.

## Suggested change

In the source directory `/src/assets/screenshots/2-22/de/android/` rename the file `RatProfilesList_Empty.png` to 
`RATProfilesList_Empty.png`.

MikeMcC399:
- PR #2833 submitted

PR #2833, opened 2022-05-04
Change case of RatProfilesList_Empty.png

MikeMcC399:
- Fix https://github.com/corona-warn-app/cwa-website/issues/2832


MikeMcC399:
The problem is not visible locally on Windows due to filename case insensitivity under this operating system.

I verified the fix locally on Ubuntu both manually and with Cypress.

http://localhost:8000/de/screenshots/#android_rat_profiles

image

larswmh (approved):

larswmh:
Good morning @MikeMcC399, thanks for your PR

---

Issue #2699, opened 2022-04-06
Heavy screenshot usage in repository (size and number of files)


MikeMcC399:
## Where to find the issue

https://github.com/corona-warn-app/cwa-website/tree/master/src/assets/screenshots

## Problem description

The [screenshots](https://github.com/corona-warn-app/cwa-website/tree/master/src/assets/screenshots) directory in the [cwa-website](https://github.com/corona-warn-app/cwa-website) repository uses a large amount of space and includes a high number of files.

Each app version release causes a complete set of images to be committed to the repository for the combinations iOS/English, Android/English, iOS/German and Android/German. Many of these files have not changed between versions, meaning that there is a high level of duplication.

Metric | Screenshots | Total web
--- | --- | ---
Size | 966 MB / 70% | 1.38 GB / 100%
Files | 11.296 / 88% | 12.881 / 100%

## Summary

Screenshots cause ...

- High use of space in total
- High number of files in total
- High number of files in each release PR
    - difficult to review
    - GitHub "Files changed" function in PRs does not work well with so many image files

## Suggested change

De-duplicate the screenshot images and use a method where only changed screenshots need to be published.

## Advantages

- Reduce the size and complexity of the repository
- If only changes need to be published with each release then reviewing the changes will be simpler. Currently more than 600 files are part of each related PR.
- GitHub "Files changed" for reviewing PRs will work better

Internal Tracking ID: [EXPOSUREAPP-14534](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14534)

MikeMcC399:
cwa-website now takes up a massive 5.3 GB space on disk.

git is not really suited to managing large numbers of image files as it tries to keep differences on each of them. The hidden `.git` folder is taking up more than 3GB. For comparison the same folder for the Android app is 590MB.

Please consider removing older screenshots from the website. I cannot copy the website over the network any more. I get an error:
```text
error: RPC failed; curl 56 Recv failure: Connection was resets
send-pack: unexpected disconnect while reading sideband packet
```
 I created a copy in order to test workflows separately, so I may not be able to contribute in this area any more.  🙁

dsarkar:
@MikeMcC399 ok, we will check, how to remove redundant image files.

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-14534](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14534)

[...]

PR #2841, opened 2022-05-06
Removed unused screenshots from assets

thorbenkuro:
Remove screenshots, that were not referenced from any of the src/data/screenshots*.json files and are not used on the website.


larswmh (approved):
Looks good to me now. This reduces the total screenshot size down to about 670MB #2699

PR #2854, opened 2022-05-06
going back to s3-sync-action

svengabr:


Ein-Tim:
@svengabr 

Hey Sven, we saw many PRs regarding the Deploy mechanism, could you please elaborate a bit why you've done and why?
To me it seems like deployment performance was improved with these changes, or?

PR #2860, opened 2022-05-09
Add note to README for case-sensitive testing

### LINKED ISSUE #2836: Website test on Windows does not find filename case issues
## Where to find the issue

[README.md](https://github.com/corona-warn-app/cwa-website/blob/master/README.md)

## Describe the issue

- The example of the issue https://github.com/corona-warn-app/cwa-website/issues/2832 shows that testing `cwa-website` manually using the Microsoft Windows operating system or with Cypress on Microsoft Windows does not show if there are filename case issues.

- The production web https://www.coronawarn.app/ runs under Ubuntu using a `nginx` web server. By default Ubuntu uses a case-sensitive file system, whereas Windows uses a case-insensitive, case-preserving file-system.

- This restriction is not mentioned in the [README.md](https://github.com/corona-warn-app/cwa-website/blob/master/README.md) file.

## Steps to reproduce

In clone of `cwa-website`, go back to state immediately after 2.22 was released by executing:

```
git checkout cd4889d7
npm run test:open
```
Select `check_links.js`

Execute the above test on Ubuntu and on Windows.

OS | result
--- | ---
Ubuntu | 38 pass, 1 fail in `/de/screenshots` 
Windows | 39 pass, 0 fail

## Suggested change

Add a note of caution to https://github.com/corona-warn-app/cwa-website#testing that testing on an operating system, such as Microsoft Windows, with a case-insensitive file-system may not detect case-dependent issues with filenames.

## References

Although it is theoretically possible to set a specific file-system directory on Windows to be case sensitive, it is probably not practical to try to integrate this into the operating-system-independent build procedure for `cwa-website`. See https://docs.microsoft.com/en-us/windows/wsl/case-sensitivity for more information.

MikeMcC399:
This PR resolves #2836 "Website test on Windows does not find filename case issues".

The PR is in draft state due to current deployment issues. Feedback on the text is however welcome.


larswmh (approved):

dsarkar (approved):

dsarkar:

PR #2907, opened 2022-05-23
Create check_videos.js Cypress Test

ahodzic2:
Create the cypress test to check video tags to ensure videos are playable.

**NOTE: This is not ready to merge and is a work in progress. As suggested by @dsarkar and @MikeMcC399, I am opening this PR to facilitate collaboration. The current test script is able to locate videos and confirm the video is paused initially and start the video, but I am having difficulty properly checking that the video is actually playable/playing. Hoping that someone else may have some suggestions or want to take a stab at it while I try to work through the kinks on my side.**

PR #2931, opened 2022-06-02
Cypress video test

brifemu:
I'm creating this PR to implement a Cypress test to check the videos' status from all the site. I want to thank @ahodzic2  for his PR #2907, which has helped me a lot to elaborate the test.

This new test is actually running without problems and has been tested by breaking the sources of some videos. The only drawback is that it doesn't save the broken URLs in a TXT file, but since there are so few videos on CWA website, I think we can avoid it.

### DIFF
--- a/cypress/integration/check_videos.js
+++ b/cypress/integration/check_videos.js
@@ -40,26 +40,20 @@ context("Check for broken videos", () => {
         cy.writeFile("cypress/logs/broken_videos_result.txt", "==================== Broken videos ====================\n")
     })
 
-    pages.forEach(page => {
+    pages.map(page => {
+        console.log(page);
         it(`"${page}" - Check for broken videos`, () => {
+            console.log("enter in test")
             cy.visit({log: false, url: page});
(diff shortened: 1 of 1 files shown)


cypress/integration/check_videos.js
```diff
@@ -0,0 +1,83 @@
...
+                    })
+                    .and('have.prop', 'paused', true)
+                    .and('have.prop', 'ended', false)
+                }
+            })
+        })
+    })
+})
+
+context("Check for broken links on entries", () => {
```
svengabr, line 61: I think the correct wording would be "Check for broken videos on blog/science entries"

brifemu: You are right, I'm gonna change it, thanks!

ahodzic2:
Replicated the new cypress test file and I can confirm this tests videos properly throughout the site.

Great job!

ahodzic2 (approved):

svengabr:
https://github.com/corona-warn-app/cwa-website/pull/2950

PR #2932, opened 2022-06-02
Rearrange Cypress test directories

MikeMcC399:
This PR moves the Cypress test file [cypress/integration/check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_links.js)  into a new directory `cypress/integration/hybrid` to make it easier to exclude this test programmatically from full test runs if desired. `check_links.js` also checks external links, which may produce varying results causing the test suite to be flagged as failing, even though there is no fault in the `cwa-website`.

At the same time the Cypress test file [cypress/integration/pages/eventRegistration.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/pages/eventRegistration.js) is moved up one level to the [cypress/integration/](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/) folder for consistency and so that it can also be selected programmatically more easily together with other tests.

## Verification

1. Before and after this PR, the command<br>
`npm test`<br>
should run all tests including [cypress/integration/check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_links.js)

2. After merging this PR, the command<br>
`npx cypress run --config baseUrl=https://coronawarn.app --spec 'cypress/integration/*.js'`<br>
should run without `check_links.js` and with `eventRegistration.js`<br>
Currently this runs locally in under one minute on a Windows 11 machine.


dsarkar:
@MikeMcC399 Thanks for PR, we will have a look at it.

svengabr:
Changes approved.
This will be merged into our ongoing cypress branch cypress-workflow and then into master later.

svengabr (approved):

Issue #2892, opened 2022-05-13
Add new GitHub Actions workflow caring out the cypress test suite


Ein-Tim:
## Feature description

A new GitHub Actions workflow caring out the cypress test suite should be added to this repository. The workflow should run for all opened PRs in this repository, like the "build and test" workflow does now. 

## Problem and motivation

- PR https://github.com/corona-warn-app/cwa-website/pull/2891 added the (strong) recommendation to carry out the cypress test suite before opening PRs in this repository.
It would be easier for everyone, the contributors as well as the Open-Source-Team if the tests are carried out automatically, without having to start them manually.

MikeMcC399:
@Ein-Tim 

It would be problematic to include the `check_links.js` Cypress test in a workflow, since this test depends on links to external websites and their state. `cwa-website` has no control over the external websites, so this could cause unpredictable failures.

@dsarkar's https://github.com/corona-warn-app/cwa-website/issues/1825#issuecomment-935668508 about the run time of the test being 10 minutes is also still relevant.

Ein-Tim:
@MikeMcC399 

> It would be problematic to include the `check_links.js` Cypress test in a workflow, since this test depends on links to external websites and their state. `cwa-website` has no control over the external websites, so this could cause unpredictable failures.

I don't think that this is a problem. The check should not be required to pass to be able to merge the PR, but rather feedback for the PR author. If the check fails, the PR author can check if this is an issue on our side or if this is an issue of an external website (which can then be fixed in another PR, by excluding the website). 

> @dsarkar's [#1825 (comment)](https://github.com/corona-warn-app/cwa-website/issues/1825#issuecomment-935668508) about the run time of the test being 10 minutes is also still relevant.

This argument seems outdated. If everyone follows the recommendation to run the Cypress test suite locally, it will take 10 minutes on their device. If it runs on GitHub, it takes 10 minutes on GitHub. There are no time savings either way.

MikeMcC399:
@Ein-Tim 

This suggestion could be implemented step-by-step. 

Some ideas:

- Add all Cypress tests, except [check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/check_links.js) into the [build](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) workflow.
- Also make the Cypress tests run periodically in a different workflow.
[...]

Issue #2909, opened 2022-05-23
Extend Cypress testing to coronawarn.app links used by CWA app


MikeMcC399:
## Feature description

Add testing to the Cypress suite in https://github.com/corona-warn-app/cwa-website/tree/master/cypress/integration to cover the links to https://coronawarn.app which are used by the CWA Android and iOS apps.

### Problem

1. In March 2022 PR https://github.com/corona-warn-app/cwa-website/pull/2434 implemented a redesign of the CWA FAQs on https://www.coronawarn.app/en/faq/ and https://www.coronawarn.app/de/faq/.
2. The redesign caused access from embedded links in the CWA on mobile devices to fail to display the requested FAQ article. Either the article was not displayed at all or it was not displayed expanded. See issue https://github.com/corona-warn-app/cwa-website/issues/2577.
3. The Cypress test suite, run using `npm test`, did not identify these issues prior to Move-To-Production.

The underlying issues have been fixed, however there is still no Cypress test available which would have flagged these issues.

- This is related to the suggestion https://github.com/corona-warn-app/cwa-website/issues/2585 which was more concerned with the FAQ display on a mobile device. It did not consider other links used by the app.

### Motivation

Running the Cypress test suite should give a high level of confidence that the website performs as needed and should provide good information to avoid publishing content which breaks the interface between CWA on the mobile device and content on https://coronawarn.app.

---
Internal Tracking ID: [EXPOSUREAPP-13210](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13210)

MikeMcC399:
Does it make sense to add this Cypress test functionality now? On the Android side I can only see [release/2.23.x](https://github.com/corona-warn-app/cwa-app-android/tree/release/2.23.x) as upcoming version. I can't see any 2.24.x or 3.0 so I don't know if there are going to be any more changes after the release of 2.23.

If there will be more changes on the app side and on the web site that would benefit from having a test in place, then it would be worthwhile to add a test.

I do have a new Cypress test prepared which does this test. I only want to submit it as a PR if it really adds value at this stage of the project. Do let me have feedback!

larswmh:
Thanks for your enhancement request @MikeMcC399. We have created an internal ticket for it and will raise this topic internally.
Internal Tracking ID: [EXPOSUREAPP-13210](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13210)

---
Corona-Warn-App Open Source Team

[...]

PR #2936, opened 2022-06-02
Add workflow for Cypress test on production with Chrome

MikeMcC399:
This draft PR partially implements the workflow requested in https://github.com/corona-warn-app/cwa-website/issues/2892 "Add new GitHub Actions workflow carrying out the cypress test suite".

It was suggested in https://github.com/corona-warn-app/cwa-website/issues/2892#issuecomment-1141760610 to run the test regularly.

It uses https://github.com/cypress-io/github-action to run Cypress tests on the production server https://coronawarn.app once a day. The tests selected are the ones in the top level of `cypress/integration`. The problematic test `check_links.js`, which is prone to failure, is already moved down one level to `cypress/integration/hybrid`, and is not selected.

Removing the line `spec: cypress/integration/*.js` would run all tests, if that is desired.

https://github.com/cypress-io/github-action can be used as the basis for other workflows, which might also be integrated into PR submission, and it allows build and test server start parameters to be added.

I only tested this so far by manually triggering it. I did not test the `cron` part.

### DIFF
--- a/.github/workflows/cypress-test-prod.yml
+++ b/.github/workflows/cypress-test-prod.yml
@@ -1,0 +1,25 @@
+name: Cypress Test Production
+# This workflow runs the Cypress test suite on the production server
+# https://coronawarn.app
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Run daily at 9:00 UTC
+    - cron: "0 9 * * *"
+
+jobs:
+  Cypress-Test-Production:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v3
+        with:
+          browser: chrome
+          headless: true
+          config: baseUrl=https://coronawarn.app
+          # Only run tests in top level of integration directory
+          spec: cypress/integration/*.js


svengabr (approved):

svengabr:
@MikeMcC399 thanks, we will have a look locally!

svengabr:
@MikeMcC399 I ran into some problems but I have fixed them in my latest commit. Now I am able to finish the workflow without errors.

https://github.com/corona-warn-app/cwa-website/commit/adb1ce970ca401a7a853f00e89d6eab6ca560839#diff-10d7404fdc00524a584d90f9b99b2e3d50d95659b03a58f21200449bbd8d43bb

2022-06-02 15_58_40-svengabr@DESKTOP-SVEN_ _mnt_c__dev_workspaces_github_cwa-website

svengabr:
@MikeMcC399 

It's not me getting it to work, it's the Docker image that is specified to run on Ubuntu Linux `runs-on: ubuntu-latest`. All the workflow code will be executed on that said image. This Ubuntu version has only Electron installed but if we want to use Chrome instead, we have to make an effort to install chrome/chromium before the tests are executed. 

Otherwise it will fail with this error message:
2022-06-03 08_33_20-svengabr@DESKTOP-SVEN_ _mnt_c__dev_workspaces_github_cwa-website

MikeMcC399:
Good morning @svengabr!

It's curious that you get this error message, as it worked for me with `chrome` on `ubuntu-latest`. I am testing natively on GitHub in a private copy of `cwa-website`.

Here is part of the log file using chrome:

image



According to https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers `Google Chrome 102.0.5005.61` is included in `ubuntu-latest` which is currently equivalent to [ubuntu-20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md). 

You mentioned in https://github.com/corona-warn-app/cwa-website/issues/2909#issuecomment-1143470698 that you were testing with https://github.com/nektos/act. I wonder if this explains the difference in test results? I haven't tried out `act` so I can't speak from personal experience. You mentioned that [ubuntu-20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md) is a docker image, however I couldn't find any reference about that, although `act` definitely does say in its documentation that it is using Docker technology.

I just submitted a tweak to the script in PR https://github.com/corona-warn-app/cwa-website/pull/2945 to make the raw log output from GitHub actions more readable.

MikeMcC399:
@svengabr 

> Act is running on docker as well, did you try again with my latest changes? Maybe I broke something while installing dependencies?
>I tried running on Windows 10 and on WSL2 Ubuntu and both resulted in the same problem. 

If you are having a problem with the Chrome browser I think that is most likely because your test environment does not emulate the production environment exactly. I did check again and I'm still able to run on Chrome successfully.

My test setup is a private copy of `cwa-website` where I pull the branch `cypress-workflow` from `cwa-website` and push it to my own copy. I have my private copy of the branch `cypress-workflow` set to be the default branch so I can test the workflow directly and natively. My copy of the branch `cypress-workflow` is up-to-date including https://github.com/corona-warn-app/cwa-website/commit/db153300ba162f1fcd670374d88c9cd75cd533bb.

I run the workflow through a browser to GitHub, so everything is running on GitHub. I do this from Windows 11, but that is not relevant, since the workflow runs remotely and directly in the GitHub runner.

I have attached a log file from a successful Chrome run. (The workflow is the same as the original, except I changed the browser from `electron` to `chrome`, added the environment variable `NO_COLOR` and disabled the `cron` schedule.)

[Successful Chrome run.zip](https://github.com/corona-warn-app/cwa-website/files/8831940/Successful.Chrome.run.zip)

MikeMcC399:
@svengabr 

> I wonder if electron has any downside I'm comparison to chrome

I found two problems running against `electron` on GitHub which do not occur every time.

They do not cause the test to fail. It just makes the test logs look suspicious. I could not find any obvious solution for either of these issues, which is why I chose to use `chrome` instead.

## Problem 1 with Electron

I found the same error in an old open issue from 2018 https://github.com/cypress-io/cypress-docker-images/issues/52

```
ALSA lib confmisc.c:767:(parse_card) cannot find card '0'
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1246:(snd_func_refer) error evaluating name
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5220:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM default
[3293:0603/110439.184841:ERROR:alsa_util.cc(204)] PcmOpen: default,No such file or directory
ALSA lib confmisc.c:767:(parse_card) cannot find card '0'
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1246:(snd_func_refer) error evaluating name
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5220:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM default
[3293:0603/110439.193337:ERROR:alsa_util.cc(204)] PcmOpen: plug:default,No such file or directory
```

## Problem 2 with Electron

This error is listed in the open issue https://github.com/electron/electron/issues/32843

`[2984:0603/110516.880897:ERROR:node_controller.cc(567)] Trying to re-add dropped peer FBF32221E33754E2.4A8A5ACE24A3244`

---
Log attached, showing both problems:
[Electron with errors.zip](https://github.com/corona-warn-app/cwa-website/files/8831951/Electron.with.errors.zip)

MikeMcC399:
@svengabr 

When you ran `act` which of the images did you choose from the ones listed on https://github.com/nektos/act#runners?

MikeMcC399:
@svengabr 

I don't think that [nektos/act](https://github.com/nektos/act) should be the preferred way of testing a Cypress workflow for GitHub. It is better to test directly in GitHub on a copy of the repository.

Cypress is for testing against web browsers. The default `act` Medium Docker Image `ghcr.io/catthehacker/ubuntu:act-20.04` however does not contain any web browsers. The `act` Large Docker Image `ghcr.io/catthehacker/ubuntu:full-20.04` is built from the repository https://github.com/catthehacker/virtual-environments-fork, which is forked from the GitHub repository https://github.com/actions/virtual-environments/. The `catthehacker` fork is however left at a commit level of September 2021, which is 9 months behind GitHub. So, neither `act` image corresponds to the current virtual environment [Ubuntu 20.04.4 LTS](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md) used by GitHub, which **does** contain up-to-date [Browsers and Drivers](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers):

- Google Chrome 102.0.5005.61
- ChromeDriver 102.0.5005.61
- Chromium 102.0.5005.0
- Selenium server 4.2.0
- Mozilla Firefox 100.0.2
- Geckodriver 0.31.0

[nektos/act](https://github.com/nektos/act) also introduces additional dependencies of Docker, WSL2 / Hyper-V or Ubuntu plus other package managers.

[nektos/act: Default runners are intentionally incomplete](https://github.com/nektos/act#default-runners-are-intentionally-incomplete) says: "These default images do **not** contain **all** the tools that GitHub Actions offers by default in their runners. Many things can work improperly or not at all while running those image. Additionally, some software might still not work even if installed properly, since GitHub Actions are running in fully virtualized machines while `act` is using Docker containers (e.g. Docker does not support running `systemd`)."

I have only a small amount of experience with Docker and I found it very difficult to get everything running, with many individual steps, so for me it was a high overhead to set up. I was not able to pull the full Docker image (around 20GB) as the download either hung up or failed with `unexpected EOF` after having downloaded several Gigabytes.

So, although it was very interesting to look at [nektos/act](https://github.com/nektos/act) I suggest to drop its use for Cypress testing. If a Cypress test run fails on `act` it does not mean that it will fail on GitHub (e.g. Chrome works on GitHub and it is not necessary to install additional Ubuntu dependencies). If a Cypress test run is successful on `act` there is no guarantee that it will work on GitHub because the contents and version levels are not aligned between `act` and GitHub.

svengabr:
> I don't think that [nektos/act](https://github.com/nektos/act) should be the preferred way of testing a Cypress workflow

@MikeMcC399 yes I agree with that. I will as well change from act to my own copy. Also, my colleague @brifemu is not available this week so I will take care of cypress in the meantime.

PR #2968, opened 2022-06-20
Include checking of external links from Android and iOS repos in app_to_web Cypress test

brifemu:
I have improved the existing test by checking the external links stored in the cwa-android and cwa-ios repositories.

For the checks, apart from external links to CWA, redirects from /faq/ to /faq/results/ and hashed URLs (for FAQ links as well as for other internal CWA links) are taken into account. 

Currently, it does not detect any errors, just sometimes gets a timeout to visit a specific Apple URL.

Screenshot 2022-06-20 at 13 15 51

Screenshot 2022-06-20 at 13 16 20


MikeMcC399 (changes requested):
I request that this new code by moved to a new test in the https://github.com/corona-warn-app/cwa-website/tree/master/cypress/integration/hybrid directory.

See above for detailed explanation.

MikeMcC399:
It seems that the test always fails, as in the screenshot in https://github.com/corona-warn-app/cwa-website/pull/2968#issue-1276774637.

```
Method: GET
URL: http://apps.apple.com/de/app/corona-warn-app/id1512595757?mt=8

No response was received within the timeout.
```

It is checking an `http` URL although the link specified by the app is `https`.

See https://github.com/corona-warn-app/cwa-app-ios/blob/eac52ea2bcdb11faad382b53f0275f9bfcfb009c/src/xcode/ENA/ENA/Resources/Localization/bg.lproj/Localizable.links.strings#L63
"TicketValidation_CoronaWarnAppLink" = "https://apps.apple.com/de/app/corona-warn-app/id1512595757?mt=8";

MikeMcC399:
I am not really familiar with how the branches of the https://github.com/corona-warn-app/cwa-app-ios are managed, but when I looked at https://github.com/corona-warn-app/cwa-app-ios/tree/main I saw that this branch has not been updated for 2 months and therefore it does not contain the current released code, which would be in https://github.com/corona-warn-app/cwa-app-ios/tree/release/2.23.x.

MikeMcC399:
@brifemu 

I'm happy you agree about my comments! Your new code is a good addition to the test suite and it is complementary to the existing test.

When you move the code to a new test I will take another look and see if I can make some other suggestions.

I used the name `hybrid` to mean a mixture of internal and external link checking. Since the external link checking is often a problem, that is why I wanted to separate the tests. Many sites try to protect themselves from Denial-of-Service attacks, which can cause Cypress probes from cwa fail, although if you access the same URL manually there is no problem.

MikeMcC399:
One other point from the point of view of testing philosophy:

I don't believe it should be the responsibility of the cwa-website testing process to test whether a link from the CWA app (iOS or Android) to an external domain like `apps.apple.com` works or not. This is the responsibility of the CWA app testing process, not the responsibility of the website testing process [cwa-website/.../cypress](https://github.com/corona-warn-app/cwa-website/tree/master/cypress). The cwa-website is not involved if an app user clicks on a link in the app to apple for instance, so if the website isn't involved then the website test should also not be involved.

The principal behind the `app_to_web.js` test was to ensure that all the links on https://coronawarn.app which are expected and used by the CWA apps are actually available and working. This can reasonably be considered as the responsibility of the website testing process. That is why the test `app_to_web.js` did not cover the external links **from the apps** such as:
- https://reopen.europa.eu
- https://play.google.com
- https://www.bundesregierung.de
- https://www.rki.de
- https://apps.apple.com

It's different if there is a page on https://coronawarn.app which is trying to connect to an external site, in which case it should be tested, but not on the basis of testing which links the apps use. Then the test should be done in the `check_links.js` test, not in `app_to_web.js`.

MikeMcC399:
Based on my previous comment, I don't think that this PR should be implemented.

MikeMcC399 (changes requested):
Based on my previous comment, I don't think that this PR should be implemented.

MikeMcC399:
@brifemu 

> I guess we can modify this test to check only the links from the cwa-apps that redirect to cwa-website, to avoid dead links like when we implemented the new FAQ.

If you exclude external links, then your new test will test essentially the same links that the original app_to_web.js tested. So the test would be more or less a duplicate. There are however quite a few differences in the test method, so the results might not be identical.

> In that case, I think we can include it in "app_to_web.js". If you agree, I will start with the modification.

No I don't think this would be good. If you think it is necessary to use also your test methodology to duplicate the original test then I ask you to put your code in a separate *.js file and we can discuss further. There are advantages and disadvantages of the two different test methods, but I really wonder if both are necessary.

MikeMcC399:
As previously noted, this PR breaks the workflow [cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml):

```
      - name: Cypress run
        uses: cypress-io/github-action@v3
        with:
          browser: chrome
          headless: true
          config: baseUrl=https://coronawarn.app
          # Only run tests in top level of integration directory
          spec: cypress/integration/*.js
```

The above is equivalent to:
`npx cypress run -s 'cypress/integration/*.js' -c baseUrl=https://coronawarn.app --headless --browser chrome`

The failing test can also be demonstrated with the simpler command:
`npx cypress run -s 'cypress/integration/app_to_web.js' -c baseUrl=https://coronawarn.app --headed`

where it times out in each of the new tests after waiting 100000ms each time and therefore causes the complete test to fail after a delay of 400000ms.

When I tested the workflow [cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml) with this PR it failed after 18 minutes. Normally the workflow completes successfully in 8 minutes.

---
This PR should not be merged. It will cause failure.

It should be considered whether these changes in general are needed. If they are needed then I suggest closing this PR and creating a new PR with only the code from the new test.

MikeMcC399:
@brifemu 

Are you going to do any more work on this PR? It seems like the best way forward would be a new PR with a new separate Cypress test.

MikeMcC399:
I suggest to close this PR. @brifemu has not given any further feedback in the last three weeks.

It contains multiple problems including the showstopper issue that it breaks the workflow  [cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml). Execute<br>`npx cypress run -s 'cypress/integration/app_to_web.js' -c baseUrl=https://coronawarn.app --headed`<br>to demonstrate the showstopper issue.

dsarkar:
@MikeMcC399 We will check how we will proceed with this.

MikeMcC399:
@dsarkar 
> We will check how we will proceed with this.

Thank you!

The showstopper (breaking the workflow) could be fixed by code changes. If that is done, then the question comes up whether this test should fail if the app includes a non-working link to a website which is **not** https://www.coronawarn.app e.g. some other Federal German or EU website. I don't think this situation should cause this website test to fail, so checks for other websites like https://reopen.europa.eu/ should be excluded.

Now if checks for other websites are excluded, the new code just duplicates the existing test except that it pulls the link data directly from the iOS and Android app repositories. So effectively it would just check the correctness of the [cypress/fixtures/appToWebLinks.json](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/fixtures/appToWebLinks.json) file which drives the URLs to be tested. This test file was created by hand in order to be able to test the different types of links better and more selectively.

I suggest that if any further work is done in this area then there should be a clear functional specification which says what the goal of additional code is to be.

svengabr:
Let me explain the background of this PR:

The app developers are unhappy with our current (live) tests since the tests are missing links that simply are not included in our static link list.

They wanted our cypress tests to read their link configuration files from the iOS and Android repository:
- iOS: https://github.com/corona-warn-app/cwa-app-ios/blob/main/src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.links.strings
- Android: https://github.com/corona-warn-app/cwa-app-android/blob/main/Corona-Warn-App/src/main/res/values/links.xml

Of course, only checking links that point to our website, not checking any external links like `https://www.bundesregierung.de/...`

MikeMcC399:
@svengabr 

Thanks for explaining the background and motivation!

> The app developers are unhappy with our current (live) tests since the tests are missing links that simply are not included in our static link list.

The production version of [cypress/integration/app_to_web.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/app_to_web.js) already checks all links to https://www.coronawarn.app which the iOS and Android apps have defined in their links tables. The Cypress test uses the following data table and checks each of the links in the two different language variations `/en` and `/de`. Please let me know if specific links are missing from this table. I will double-check from my side as well. 

I am also looking at how to automate the check to make sure no links have been missed out. That involves extracting links from the app files, filtering out links which are not https://www.coronawarn.app, consolidating, de-duplicating and sorting them.

https://github.com/corona-warn-app/cwa-website/blob/0ecf5ed9dc20e3c9f6b7f1af69d0e980301cc1fe/cypress/fixtures/appToWebLinks.json#L1-L47

> Of course, only checking links that point to our website, not checking any external links like https://www.bundesregierung.de/...

Thanks for clarifying that point!

MikeMcC399:
@svengabr 

- I checked whether the test [cypress/integration/app_to_web.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/app_to_web.js) is covering all cases, and I found two gaps: 
    - The test was not covering the default FAQ landing page https://www.coronawarn.app/en/faq/
    - Android and iOS are using two different variants to access the blog page https://www.coronawarn.app/en/blog/2021-08-05-statistictiles/, one is with a trailing forward slash character (/) and the other is without the slash character. Only the variant without the slash was being explicitly tested.
- I documented the findings on https://github.com/corona-warn-app/cwa-website/issues/3028 and submitted a solution in PR https://github.com/corona-warn-app/cwa-website/pull/3029 for the above two issues.

MikeMcC399:
@svengabr 

- Since PR #3029 is now merged, I don't see any other gaps in the testing. Please let me know if you think differently.

MikeMcC399:
@svengabr / @brifemu 

I saw some movement again in this PR from last week. Do you still want to go ahead with it?

I have also been looking at the task of dynamically taking the links files from the apps and checking them, so I have a few ideas. Unfortunately I have very little JavaScript experience, so I have a steep learning curve and I'm quite slow.

If you do go ahead, then I ask you to put the new test in a new script (if you need the old code then take a copy of it), leaving the existing `app_to_web.js` as it is. It seems the old code is being used to work around the redirects, but really if the test is to use live data, then it should also use the live redirect file https://github.com/corona-warn-app/cwa-website/blob/master/src/data/faq_redirects.json as well.

svengabr:
@MikeMcC399 we tried around a little but but finally decided to close this PR.

I think https://github.com/corona-warn-app/cwa-website/pull/3029 is good and additional changes are not needed.

MikeMcC399:
@svengabr 

I think that closing this PR was the right decision. The app links are all being checked by the existing test.

Even so, if the link data is taken dynamically from the app repositories it would take away the manual maintenance of the fixtures file. I will carry on looking at this and I will let you know if I get a separate working solution. It would only be worth putting into production if the CWA app has a longer lifetime with several more releases and the links from the app into the website are changing a lot, or if the existing links get different redirections implemented.

Issue #2970, opened 2022-06-21
Per user limit on certificate reissue


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/results/#dcc_replacement_issues "I cannot renew a certificate in the app. What could be the reason?"<br><br>"A certificate can only be reissued a maximum of three times per user."

- https://www.coronawarn.app/de/faq/results/#dcc_replacement_issues "Ich kann ein Zertifikat in der App nicht erneuern. Woran kann das liegen?"<br><br>"Die Anzahl der Neuausstellungen ist auf drei Neuausstellungen pro Nutzer begrenzt."

## Describe the issue

The FAQ article https://www.coronawarn.app/en/faq/results/#dcc_replacement_issues "I cannot renew a certificate in the app. What could be the reason?" lists the following as one reason why a certificate cannot be renewed:

- "A certificate can only be reissued a maximum of three times per user."

The text "per user" is misleading in this context.

In most cases "user" means the user of the device, but here it is really referring to the user as the name on the certificate.

Another device user cannot take the certificate and get it reissued if the certificate has  been reissued three times already.

## Suggested change

Remove the text "per user".

The sentence:

- "A certificate can only be reissued a maximum of three times"

contains a sufficient and correct explanation.

For the German text, similarly remove "pro Nutzer".

Alternatively, use exactly the same text as the CovPass FAQs use (see below):

## Additional information

https://digitaler-impfnachweis-app.de/en/faq says:

- The number of reissues is limited to three per certificate.

https://digitaler-impfnachweis-app.de/faq

- Die Anzahl der Neuausstellungen ist auf drei Neuausstellungen je Zertifikat begrenzt.

MikeMcC399:
This would be a simple change, however I would like confirmation please before I submit a PR.

dsarkar:
Hi @MikeMcC399, thanks for opening this issue. Indeed, your suggestion LGTM, PR will be merged. Best, DS.

PR #2971, opened 2022-06-21
Certificate renewal limit unrelated to user

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/2970 "Per user limit on certificate reissue".

It modifies:

- https://www.coronawarn.app/en/faq/results/#dcc_replacement_issues "I cannot renew a certificate in the app. What could be the reason?"

- https://www.coronawarn.app/de/faq/results/#dcc_replacement_issues "Ich kann ein Zertifikat in der App nicht erneuern. Woran kann das liegen?"

The certificates can be reissued a maximum of three times per certificate, not per user.

The misleading reference to a dependency on the user is removed and the text is aligned to the FAQ text of the CovPass app (see https://digitaler-impfnachweis-app.de/faq).


dsarkar (approved):

dsarkar:
Thanks @MikeMcC399, LGTM.

PR #2986, opened 2022-06-29
added files for new blog post

HannaHei:


MikeMcC399:
There is a problem with GitHub Actions
https://www.githubstatus.com/incidents/qzmhbwp7f6nn

so the blog has not been deployed yet.

larswmh:
> There is a problem with GitHub Actions https://www.githubstatus.com/incidents/qzmhbwp7f6nn
> 
> so the blog has not been deployed yet.

Thanks for the information. The deployment went through now and the blog is available on the website. All other 2.24 changes are now queued to be deployed, as it seems like GitHub actions are still operational with longer wait times.

Issue #2990, opened 2022-06-30
iOS: Layout issue when using the blog search on mobile devices


Ein-Tim:
## Where to find the issue
- https://www.coronawarn.app/en/blog/
- https://www.coronawarn.app/de/blog/

## Describe the issue
Using the blog search on my iPhone XR, the page zooms in to the search field, however, this is not a good thing to do, as the page then cuts of some content. See this video:

https://user-images.githubusercontent.com/67682506/176706022-728f4079-c39e-4a33-a306-a7f9d8fb09f9.MP4

## Suggested change
Do not zoom in to the search field at all OR align the search to the one on https://www.coronawarn.app/de/faq.

---
Internal Tracking ID: [EXPOSUREAPP-13502](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13502)

MikeMcC399:
@Ein-Tim 

Which browser are you using?

I did not see the problem with Google Chrome emulating an iPhone XR.

Ein-Tim:
@MikeMcC399 I'm using Safari for iOS. When you tried to emulate, was the software keyboard shown? I think this is what causes the website to zoom in.

MikeMcC399:
@Ein-Tim 

> When you tried to emulate, was the software keyboard shown? 
No, I did not use a software keyboard. I emulated under Windows 11 using the hardware keyboard.

I expect that your software keyboard comes from iOS, not from the Safari for iOS browser.  

I also tried Google Chrome on Android where there is no issue either. Do you have the ability to test with Google Chrome on iOS?

This issue is discussed in https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone

Ein-Tim:
@MikeMcC399 

> I expect that your software keyboard comes from iOS, not from the Safari for iOS browser.

Yes, that's correct.

> Do you have the ability to test with Google Chrome on iOS?

I will try.

> On your video it did not seem to prevent you using the blog search function.

No, but it zoomed into the page so that it made the blog post teasers cut off.

MikeMcC399:
@Ein-Tim 

This seems to be an issue with iOS. Apart from
https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone which I also mentioned, there is another article 
[...]

PR #3002, opened 2022-07-05
Increase input font size to avoid iOS input zoom

brifemu:
Due to issue #2990, I increased the input's font size just for iOS devices to avoid the input zoom.

Tested in a iPhone 6.


larswmh:
Tested on two devices:
- iPhone 8: it works
- iPhone 11 Pro Max: it doesn't work

brifemu:
@larswmh 
I included a few more iPhones media queries to control it

larswmh (approved):
I can now confirm that this PR fixes the issue for iPhone 8 & iPhone 11 Pro Max

dsarkar (approved):

Ein-Tim (approved):
Tested on iPhone XR. Works!

dsarkar:
@Ein-Tim Thanks for review!

Issue #2981, opened 2022-06-24
Backarrow on FAQ page does not restore view after visiting Glossary


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/en/faq/results/#glossary

## Describe the issue

After following a link to a target in the glossary, where the link opens in the same frame, using the browser backarrow does nothing.

Opening a followed link in the same frame is the default action and equivalent to `target="_self"`.

## Steps to reproduce

1. Open https://www.coronawarn.app/en/faq/results/#glossary_android in Google Chrome
2. Click on [iOS](https://www.coronawarn.app/en/faq/results/#glossary_ios)
3. Note that the iOS glossary article is now displayed
4. Click on the browser back arrow
5. Note that https://www.coronawarn.app/en/faq/results/#glossary_android is shown in the browser address bar, however the Android article is not displayed
6. Press F5 (refresh)
7. Note that the Android article is now displayed

The issue is also reproducible in Mozilla Firefox

## Expected behaviour

The browser backarrow should return to the previous view after following a link to the glossary which has either `target="_self"` or no `target` frame defined in the link.

---
Internal Tracking ID: [EXPOSUREAPP-13506)](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13506)

MikeMcC399:
@svengabr raised the question in https://github.com/corona-warn-app/cwa-website/pull/3009#issuecomment-1185312353 whether it would be better to also use `target="_blank"` in the Glossary.

This would give a more consistent user experience and it would workaround the backarrow issue by avoiding its use in the Glossary.

MikeMcC399:
- PR #3030 works around this issue by avoiding the default use `target='_self'` and replacing it by `target='_blank'` in the FAQs. In this case the back arrow is not activated and therefore cannot cause problems.

PR #3009, opened 2022-07-06
Refresh page on change location hash in faq-results

brifemu:
Due to issue #2981 I'm creating this PR. The problem is that on faq-results page we don't navigate, content is hidden or showed depend on the location hash.

So, when a user go back on his browser history the content doesn't change, because the content is manipulated by our script. To get this functionality, we need refreshing on hash change.

**Pros:**
- We can navigate like in a normal web using back and forward buttons

**Cons:**
- Every time we browse we will consume more resources, because before the web was only refreshed when searching or selecting a topic, now it will refresh every time we change the hash (which are many).
- Visually it is easier to get dizzy because when you refresh the web, all the content loads again and the script have to hide again the content that we do not want to show, creating jumps on the screen.

I will leave this PR open for a time to give you the opportunity to test it and receive feedback from the community, as well as to receive suggestions for other ways to solve the problem.

---
Internal Tracking ID: [EXPOSUREAPP-13506)](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13506)


Ein-Tim:
Although this PR fixes the issue, it introduces graphic glitches caused by the reloading mentioned in the OP all over the FAQ page, so I would vote to not merge this in the current state.

MikeMcC399:
@brifemu 

- Thank you for making an attempt to fix issue #2981!
- I had only noted this problem in the Glossary section, not in other parts of the FAQ. In the rest of the FAQ links in the body of the articles are generally set up to open in a new frame `target='_blank'` in which case the back arrow has no effect in any case.
- There is however an example https://www.coronawarn.app/en/faq/results/#who_can_help which links to https://www.coronawarn.app/en/faq/results/#qr_test without  `target='_blank'` and in this case the back arrow works (without this PR), so this shows that the issue is only in the Glossary.
- In the Glossary the links (`href`) are set up without any `target` defined.
- It has been noted that the refresh has undesirable side-effects.

## Conclusion

Since the problem only affects the Glossary and if the solution produces side-effects which negatively affect the rest of the FAQ then it would be better not to implement this solution.

svengabr:
Agreed, so would it be better to have `target="_blank"` in the glossary so it's the same as in faq?

MikeMcC399:
@svengabr 

> ... so would it be better to have target="_blank" in the glossary so it's the same as in faq?

That would work around the back-arrow issue and provide a more consistent user experience.

If this is changed for the Glossary, then it would be good to also make sure all the corresponding links in the top part of the FAQs also use  `target="_blank"` as currently there are a few internal links which don't have this attribute set.

PR #3030, opened 2022-07-15
Update links: Open all FAQ links in a new tab

### LINKED ISSUE #2981: Backarrow on FAQ page does not restore view after visiting Glossary

larswmh:
Resolves #2981 

With this PR, any link in the FAQ opens up in a new tab. The changes are quite hard to represent in screenshots since there were many changes in this, however I'd like to explain my method of solving this.

Using the regular expression `<a href='[^']*'>`, any links that **only** contain the `href` attribute were highlighted. The attribute `target='_blank'` was added to all matching results, as suggested in the issue and preceding pull request (#3009).

---
Internal Tracking ID: [EXPOSUREAPP-13506)](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13506)

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -274,7 +274,7 @@
                                 "anchor": "rat_tan_hotline",
                                 "active": false,
                                 "textblock": [
-                                    "It is not possible to share a positive <a href='#glossary_rapid_antigen_test'>rapid antigen test</a>, <a href='#glossary_selftest'>self-test</a> or <a href='#glossary_rapid_pcr_test'>rapid PCR test</a> result via the TAN hotline."
+                                    "It is not possible to share a positive <a href='#glossary_rapid_antigen_test' target='_blank'>rapid antigen test</a>, <a href='#glossary_selftest' target='_blank'>self-test</a> or <a href='#glossary_rapid_pcr_test' target='_blank'>rapid PCR test</a> result via the TAN hotline."
                                 ]
                             },
                             {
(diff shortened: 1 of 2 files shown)


dsarkar (approved):

MikeMcC399:
@dsarkar / @larswmh 

There has never been a written policy here about external links (it's never too late though!) but for consistency  `rel='noopener noreferrer'` would need to be added to ensure security for external links, such as 

`<a href='https://tools.rki.de/PLZTool/en-GB' target='_blank'>Robert Koch Institute Tool: Health authority by postal code or place</a>`

which have been converted to use `target='_blank'`.

larswmh:
@MikeMcC399 I've created issue #3031 to keep track of this issue as we're planning to fix this soon. Thanks for bringing this up!

PR #3032, opened 2022-07-18
Update FAQ: Ensure security on external links

larswmh:
Fixes #3031 

`rel='noopener noreferrer'` was added to all external links in the FAQ. 

Some links that were not covered by #3030 due to the fact that they had `target='_self'` in it were corrected to `target='_blank'`, so that they open up in a new tab as intended.

---
Internal Tracking ID: [EXPOSUREAPP-13638](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13638)

PR #3033, opened 2022-07-18
FAQ link fixes for fix/3031

MikeMcC399:
@larswmh 

I managed to make progress on Cypress testing and found some other link issues to be fixed. I've described the changes below and put them into this PR as an add-on to your PR https://github.com/corona-warn-app/cwa-website/pull/3032.

In the English and German FAQs:

- In https://www.coronawarn.app/en/faq/results/#vac_booster_basics there is a link with `target='blank'` instead of` '_blank'`. That opens a new frame using the frame name `blank`, which is probably not what is intended (see https://www.w3schools.com/tags/att_a_target.asp) so I changed it.

- In https://www.coronawarn.app/en/faq/results/#check_in_qr_code_other_codes I changed `_self` to `_blank` to conform with the other links

- In https://www.coronawarn.app/en/faq/results/#who_can_help there was a malformed hyperlink with `<a/>` instead of `</a>` to the community page which I fixed

- In https://www.coronawarn.app/en/faq/results/#where_can_i_get_tested and https://www.coronawarn.app/en/faq/results/#qr_test I added missing `target='_blank'` attributes 

In addition in the German Glossar

- https://www.coronawarn.app/de/faq/results/#glossary_rapid_antigen_test I added a missing `rel='noopener noreferrer'` to the link for an external PDF file. This was already correct in the English Glossary.


Ein-Tim (approved):
Changes looking good to me!

larswmh (approved):
Thanks for these fixes

Issue #3043, opened 2022-07-22
sitemap.xml includes unrounded priority 0.3999999999999999


MikeMcC399:
## Where to find the issue

https://www.coronawarn.app/sitemap.xml 

## Describe the issue

The autogenerated https://www.coronawarn.app/sitemap.xml includes entries with an unrounded `<priority>0.3999999999999999</priority>` entry such as:

```
    <url>
        
        <loc>https://coronawarn.app/de/faq/results/</loc>
    
        
        <lastmod>2022-07-21T11:46:46.438Z</lastmod>
    
        
        <priority>0.3999999999999999</priority>
    
    </url>
```

The current priority calculation yields `1`, `0.8`, `0.6`, `0.3999999999999999`, `0.19999999999999996` and `0` depending on the depth of the URL from the top level, although `0.19999999999999996` does not occur in practice because there are currently no URLs at this depth on `cwa-website`.

## Steps to reproduce

Execute `npm run build`
Examine the file `public/sitemap.xml` in the local cwa-website repository
Note entries with `<priority>0.3999999999999999</priority>`

## Expected behaviour

The priority is expected to go down exactly in 0.2 steps i.e. `1`, `0.8`, `0.6`, `0.4`, `0.2` and `0` according to the depth of the URL from the top level of the web.

Instead of priority `0.3999999999999999` there should be a priority `0.4` shown.

## Suggestion

Change the calculation https://github.com/corona-warn-app/cwa-website/blob/ca65a737858f606bc75be2f7063e5d90e11aba40/gulpfile.js#L488 to

`return (10 - (entry.file.split('/').length - 1) * 2) / 10`

See https://www.w3schools.com/js/tryit.asp?filename=tryjs_numbers_inaccurate3.

## Reference

According to https://www.sitemaps.org/protocol.html

`priority`

optional

The priority of this URL relative to other URLs on your site. Valid values range from 0.0 to 1.0. This value does not affect how your pages are compared to pages on other sites&mdash;it only lets the search engines know which pages you deem most important for the crawlers.

The default priority of a page is 0.5.

MikeMcC399:
- PR #3044 submitted to resolve this issue.

PR #3044, opened 2022-07-22
Correct sitemap priority calculation

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3043 "sitemap.xml includes unrounded priority 0.3999999999999999" by changing the priority calculation to avoid unintended precision inaccuracies.

The priority calculation is first carried out using whole numbers then divided by 10:

`return (10 - (entry.file.split('/').length - 1) * 2) / 10`

## Verification

Execute `npm run build`

Examine the file `public/sitemap.xml` in the local cwa-website repository
Check that there are no longer any entries with priority `0.3999999999999999`
Check that the entry for https://coronawarn.app/de/faq/results/ shows priority `0.4`
Check that only priorities `1`, `0.8`, `0.6`, `0.4`, `0.2` and `0` are used

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -485,7 +485,7 @@ function build_sitemap() {
       siteUrl: "https://coronawarn.app",
       priority: function (siteUrl, loc, entry) {
         // Reduce priority by 0.2 per level
-        return 1.0 - (entry.file.split('/').length - 1) * 0.2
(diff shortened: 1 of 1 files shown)


dsarkar:
Good morning @MikeMcC399,
 Thanks for reporting and providing a PR!

PR #3066, opened 2022-07-31
Improve English of Science 5 blog

MikeMcC399:
This PR contains improvements to the Science 5 blog English translation on https://www.coronawarn.app/en/science/2022-03-03-science-blog-5/ "How many active users does the Corona-Warn-App have?"

Mainly punctuation and grammar are changed.
In case of doubt the English text has been compared to the German text to check the intended meaning.


sapcoder123:
RKI review is done

PR #3083, opened 2022-08-24
update dependencies

thomasaugsten:


MikeMcC399 (changes requested):
This PR reverts to lockfile version 1, which is not desirable. It should stay at  `"lockfileVersion": 2` to be compatible with the LTS version of NPM in use on this website.

https://github.com/corona-warn-app/cwa-website/blob/b186ecc27d5dbfeb0b09195e90927dfef5ab97c1/package-lock.json#L4 

Please  re-run the update with Node 16 installed as specified in https://github.com/corona-warn-app/cwa-website/blob/dependencyUpdates/README.md#requirements

"You need the Node.js 16 Active LTS version of [Node.js](https://nodejs.org/en/) (which includes npm) to build the website."

MikeMcC399:
You need to use npm version >= 7 (see https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion) to ensure `lockfileVersion 2`. Preferably use the current version 8, which is bundled into https://nodejs.org/en/.

`npm -v`

reveals the version in use.

thomasaugsten:
node-gyp 7 or 8 is not compatible with node-sass
I can only update with node-gyp 6 the package-lock
But If you can update the package-lock with npm 8 please update the package-lock as a commit.

MikeMcC399:
@thomasaugsten 

I see you solved the issue with the package-lock version. 👍🏻 So I assume you don't need me to do anything more.

PR #3178, opened 2022-10-27
[FAQ] move text from #vac_status to #vac_children

brianebeling:
New:

de
vac_children 
image

vac_status
image

en
vac_children
image

vac_status
image

Internal Tracking ID: [EXPOSUREAPP-14241](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14241)


Ein-Tim (commented):
`vac_children` is a new FAQ entry. I don't see why there should be an update date for this entry, as this is the first version to go live.

brianebeling:
> `vac_children` is a new FAQ entry. I don't see why there should be an update date for this entry, as this is the first version to go live.

Usually we don't add an "updated at" to new entries. However, it can be valueable to add it. Seeing "updated at X.X.2020" could indicate that the info might not be up-to-date. We are of course trying to keep everything up to date, but this increases transparence and makes it easier for anyone to check.

And also in this case specifically, the text in that "new" entry is from an old one.

dsarkar (approved):

Issue #3184, opened 2022-10-28
Cypress times out checking https://emergentalliance.org/


MikeMcC399:
## Where to find the issue

[cypress/integration/hybrid/check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/hybrid/check_links.js)

## Describe the issue

Running the Cypress test [cypress/integration/hybrid/check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/hybrid/check_links.js) times out with the error message:

```
  1) Check for broken links on entries
       "/en/blog/" entries - Check for broken links:
     CypressError: `cy.request()` timed out waiting `50000ms` for a response from your server.

The request we sent was:

Method: GET
URL: https://emergentalliance.org/calculation-of-the-effective-reproduction-number-germany/

No response was received within the timeout.
```

## Steps to reproduce

`npm run test:open`
select `hybrid/check_links.js` in Cypress test runner

Note error (see above) testing the link https://emergentalliance.org/calculation-of-the-effective-reproduction-number-germany/

## Suggestion

Add https://emergentalliance.org/calculation-of-the-effective-reproduction-number-germany/ to the allowlist of entries in [cypress/integration/hybrid/check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/hybrid/check_links.js)

MikeMcC399:
This issue is no longer reproducible.

MikeMcC399:
I spoke too soon! The issue happens sometimes and sometimes not.

MikeMcC399:
- PR #3185 submitted to solve this issue.
- Running `npm test` is successful, so closing once more.

PR #3185, opened 2022-10-28
Add emergentalliance.org to check links exceptions

MikeMcC399:
This PR resolves the issue https://github.com/corona-warn-app/cwa-website/issues/3184 "Cypress times out checking https://emergentalliance.org/" by adding https://emergentalliance.org/calculation-of-the-effective-reproduction-number-germany/ to the allowlist of entries in [cypress/integration/hybrid/check_links.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/integration/hybrid/check_links.js)

## Verification

`npm run test:open`
select `hybrid/check_links.js` in Cypress test runner

The test should run without errors.


MikeMcC399:
Since this issue is no longer reproducible there is no need for a PR.

MikeMcC399:
This fix is needed after all, since the issue appears sporadically.

dsarkar (approved):

dsarkar:

Issue #3194, opened 2022-10-31
Experimental json import warning from build


MikeMcC399:
## Where to find the issue

https://github.com/corona-warn-app/cwa-website/blob/4bf9ad8704d2982c2cd84c80a2531dbe8edf2b4b/gulpfile.mjs#L33

## Describe the issue

- PR https://github.com/corona-warn-app/cwa-website/pull/3181 migrated

`const analyseConfig = require('./src/data/analyse.json');`

to

`import analyseConfig from './src/data/analyse.json' assert {type: 'json'};`

Executing the command

`npm run build` results in the message
`(node:21804) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time`

## Steps to reproduce

Execute the command

`npm run build`

## Suggestion

Replace

```js
import analyseConfig from './src/data/analyse.json' assert {type: 'json'};
```

with

```js
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
const analyseConfig = require('./src/data/analyse.json');
```

## References

- https://nodejs.org/api/esm.html#import-assertions<br>lists Import assertions with [Stability: 1](https://nodejs.org/api/documentation.html#stability-index) - Experimental

MikeMcC399:
- PR #3195 submitted.

PR #3195, opened 2022-10-31
Fix experimental json import in gulpfile.mjs

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3194 "Experimental json import warning from build"

It changes the method used to import a `json` file in an ESM environment from using an experimental feature to using a supported `Node.js` feature.

Adding the lines:

```js
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
```

before the original code from the CommonJS environment:

```js
const analyseConfig = require('./src/data/analyse.json');
```

resolves the issue.

## Verification

`npm run build` should run successfully and it should not output the following message:
`(node:21804) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time`

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -32,7 +32,13 @@ const pluginSass = gulpSass(npmSass);
 
 import { processBlogFiles, getBlogEntries } from './src/services/blog-processor.js';
 import { processScienceBlogFiles } from './src/services/science-blog-processor.js';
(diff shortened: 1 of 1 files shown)


dsarkar (approved):

MikeMcC399:
Good morning @dsarkar !
Thanks for merging!

dsarkar:
Good morning @MikeMcC399 ! Thanks for providing PR !

PR #3199, opened 2022-11-04
Correct wrong anchor tags in Accessibility

MikeMcC399:
This PR corrects two malformed closing anchor tags in e-mail links in:

- https://www.coronawarn.app/en/accessibility/
- https://www.coronawarn.app/de/accessibility/

## Reference

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a


dsarkar:
Good morning @MikeMcC399! Thanks for the PR!

Issue #3212, opened 2022-11-16
Reserved HTML characters "<" and ">" used in FAQs


MikeMcC399:
## Where to find the issue

FAQs

- https://www.coronawarn.app/en/faq/results/
- https://www.coronawarn.app/de/faq/results/

## Describe the issue

There is a lot of incorrect literal usage of the characters `<` and `>` in the FAQs. These are characters reserved to delimit HTML tags.

These are used in:

1. Mathematical comparisons, such as `>= 63dB`
2. In a sequence of instructions, such as `'Settings' -> 'Reset App'` and `'Security & Location' > 'Location' > 'App level permissions'`

## Suggestion

Replace literal usage of the reserved characters `<` and `>` in the FAQs by their escaped equivalents as follows

| Character | Escaped<br>character |
| --------- | -------------------- |
| `<`       | `&lt;`               |
| `>`       | `&gt;`               |

## References

- https://developer.mozilla.org/en-US/docs/Glossary/Entity#reserved_characters

---
Internal Tracking ID: [EXPOSUREAPP-14343](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14343)

MikeMcC399:
I have not done any work on this yet and I will wait for feedback first. 

If this is something which should be cleaned up, then it would also be an opportunity to standardize on the way that navigating menus should be described. There is a mixture of `>` and `->`. Which of these is the preferred notation? For example:

- `'Settings' -> 'Reset App'` or
- `'Settings' > 'Reset App'`

larswmh:
@MikeMcC399 

> I have not done any work on this yet and I will wait for feedback first.
> 
> If this is something which should be cleaned up, then it would also be an opportunity to standardize on the way that navigating menus should be described. There is a mixture of `>` and `->`. Which of these is the preferred notation? For example:
> 
> * `'Settings' -> 'Reset App'` or
> * `'Settings' > 'Reset App'`

Thanks for opening this issue. I agree that it would be a good opportunity to standardize the way navigation menus should be described and I think we should go with `>` as the preferred notation. Should we prepare a PR for this or are you interested in working on this?

MikeMcC399:
@larswmh 

> Should we prepare a PR for this or are you interested in working on this?

I'm happy to prepare a PR for the changes.

MikeMcC399:
The following PRs have been submitted to resolve this issue:

- #3213 
- #3214 

[...]

PR #3213, opened 2022-11-16
Convert app navigation symbols ">" and "->" to legal HTML

MikeMcC399:
This PR partially resolves issue https://github.com/corona-warn-app/cwa-website/issues/3212 "Reserved HTML characters "<" and ">" used in FAQs" for the case where the symbols are used to describe a chain of menu selections in the app such as in 
`'Settings' -> 'Reset App'` or `'Security & Location' > 'Location' > 'App level permissions'`

Where the symbol set "->" has been used, this is standardized to ">".

">" is then converted to its escaped form: `&gt;`.


GisoSchroederSAP (approved):
Nice beautifier, thanks a lot!
Obviously, approved.

brianebeling (approved):

PR #3214, opened 2022-11-16
Replace / remove lt and gt comparison operators in FAQs

MikeMcC399:
This PR partially resolves issue https://github.com/corona-warn-app/cwa-website/issues/3212 "Reserved HTML characters "<" and ">" used in FAQs" for the case where the symbols are used for a comparison of less than or greater than.

This concerns mainly the FAQ article
https://www.coronawarn.app/en/faq/results/#encounter_19_calc and references to it.

In the case of the reference to version 1.9 in the title of https://www.coronawarn.app/en/faq/results/#encounter_19_calc as  "(version >=1.9)" this problematic comparison has been removed. Version 1.9 was released almost 2 years ago on December 16, 2020 and in the meantime the current released version is 2.28. The text of the FAQ article is clear that it refers to version 1.9 and later, so it is unnecessary at this time to keep this information also in the title of the FAQ article.

In addition, in the German version of https://www.coronawarn.app/de/faq/results/#github there is a comparison operator. This has been changed to plain text instead of a mathematical operator to match the rest of the text.


GisoSchroederSAP (approved):
Changes validated; they are correct and consistent. Thanks a million to pay attention to these tiny 
 but important details.
Approved!

PR #3219, opened 2022-11-17
Fix img width attribute in 2.18 blog

MikeMcC399:
This PR resolves an HTML-compliance issue in:

- https://www.coronawarn.app/en/blog/2022-02-23-cwa-2-18/ and
- https://www.coronawarn.app/de/blog/2022-02-23-cwa-2-18/

`px` is removed from `width` attributes in `img` tags. Pixel as unit is implied and must not be explicitly specified.

## Reference

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width

"width
The intrinsic width of the image in pixels. Must be an integer without a unit."

### DIFF
--- a/blog/2022-02-23-cwa-2-18/index.md
+++ b/blog/2022-02-23-cwa-2-18/index.md
@@ -15,7 +15,7 @@ The project team consisting of the Robert Koch Institute, Deutsche Telekom and S
 
 <div class="right-float">
 <figure>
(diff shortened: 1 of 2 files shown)


brianebeling (approved):

brianebeling:
Perfect - thank you very much for the additional PR!

PR #3220, opened 2022-11-18
Update traffic_monitoring

### LINKED ISSUE #984: Does `traffic_monitoring` need an update?
## Your Question

* Website URL Path: https://www.coronawarn.app/en/faq/#traffic_monitoring & https://www.coronawarn.app/de/faq/#traffic_monitoring
* Line / Paragraph: Whole entry
* Question: Is this information there still up to date?

The entry still says:
>Furthermore, various additional technical measures will be introduced shortly, for example, by randomly creating and sending false notifications that will be discarded on the server side.

AFAIK the so called playbook (which creates these random requests) is already implemented (see for example iOS: https://github.com/corona-warn-app/cwa-app-ios/blob/release/2.14.x/src/xcode/ENA/ENA/Source/AppDelegate%20&%20Globals/PlausibleDeniabilityService.swift)

---
Internal Tracking ID: [EXPOSUREAPP-13563](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13563)

brianebeling:
Fixes #984 

See file changed for the text changes.


src/data/faq.json
```diff
@@ -2763,7 +2763,7 @@
...
-                                    "Furthermore, various additional technical measures will be introduced shortly, for example, by randomly creating and sending false notifications that will be discarded on the server side. These false notifications cannot be distinguished from valid notifications and create a type of background noise. The transmission of TANs and upload of real infection data cannot be distinguished from this background noise. So, even with monitored network traffic, plausible deniability is still possible."
+                                    "Furthermore, various additional technical measures have been taken. For example, randomly creating and sending false notifications that will be discarded on the server side. These false notifications cannot be distinguished from valid notifications and create a type of background noise. The transmission of TANs and upload of real infection data cannot be distinguished from this background noise. So, even with monitored network traffic, plausible deniability is still possible."
```
Ein-Tim, line 2766: I think the "by" has to go back into this sentence, @MikeMcC399?

Ein-Tim (approved):
Better than before! But the text is really complicated and I doubt that non technical users will understand something like "plausible Abstreitbarkeit".

src/data/faq.json
MikeMcC399: @Ein-Tim 

The original English was a little weak grammatically and the changes carry this weakness over. I think it is good enough to publish as is though. If you want to make it better you could change the second sentence to:

"For example, false notifications that will be discarded on the server side are randomly created and sent."

This is a complete sentence and follows on directly as a detailed description of "additional technical measures"..

brianebeling: Thanks for the improvement!

Issue #3223, opened 2022-11-18
Incorrect copyright Division in footer


MikeMcC399:
## Where to find the issue

Footer of each page on https://www.coronawarn.app/

## Describe the issue

The copyright Content Division element in the footer of each page is incorrectly terminated with `<div/>`:

```
<div class="copyright text-center text-sm-left">
<a href="https://github.com/corona-warn-app/cwa-website/graphs/contributors" target="_blank" rel="noopener noreferrer">© 2020-2022 - The authors of the Corona-Warn-App open-source project</a>
<div/>
```

## Steps to reproduce

Run https://validator.w3.org on https://www.coronawarn.app/en/ and note errors

- "End tag footer seen, but there were open elements"
- "Unclosed element div"

## Suggestion

Change the closing tag of copyright to `</div>`

MikeMcC399:
- PR https://github.com/corona-warn-app/cwa-website/pull/3224 submitted to resolve this issue

MikeMcC399:
- Following the merge of PR #3224 running https://validator.w3.org/ on https://www.coronawarn.app/en/ 
no longer outputs the error "End tag footer seen, but there were open elements"


- The remaining reasons for the message: "Unclosed element div" are covered in issue https://github.com/corona-warn-app/cwa-website/issues/3236

Closing, as this issue is now resolved.

PR #3224, opened 2022-11-18
Correct copyright footer closing div

MikeMcC399:
- This PR resolves the issue https://github.com/corona-warn-app/cwa-website/issues/3223 "Incorrect copyright Division in footer".

- It changes the copyright division closing tag to `</div>` in https://github.com/corona-warn-app/cwa-website/blob/master/src/partials/footer.html.

- The incorrect tag was introduced by https://github.com/corona-warn-app/cwa-website/pull/2644

## Verification

Run https://validator.w3.org on https://www.coronawarn.app/en/ and the following errors should no longer produced:

- "Self-closing syntax (/>) used on a non-void HTML element"
- "End tag footer seen, but there were open elements"
- "Unclosed element div"


brianebeling:
Thank you for the PR!

Issue #3225, opened 2022-11-18
Incorrect placing of screen-reader text in navigation


MikeMcC399:
## Where to find the issue

Navigation header of each page on https://www.coronawarn.app/

## Describe the issue

- The screen-reader `<span>` containers for sign language and simple language are inside the related `<svg>` container which causes an HTML error.

-  This item was added by PR https://github.com/corona-warn-app/cwa-website/pull/2644

## Steps to reproduce

Run https://validator.w3.org on https://www.coronawarn.app/en/ and note errors

- "HTML start tag span in a foreign namespace context."
- "Stray end tag svg"

## Suggestion

Move the screen-reader `<span>` down for

- `<span class="sr-only">The Corona-Warn-App project in sign language</span>`
- `<span class="sr-only">The Corona-Warn-App project in simple language</span>`

so that it is outside `svg` and still inside `a`

MikeMcC399:
- PR #3226 submitted to resolve this issue

PR #3226, opened 2022-11-18
Reorder screen-reader title for icons in navigation header

MikeMcC399:
This PR resolves the issue https://github.com/corona-warn-app/cwa-website/issues/3225 "Incorrect placing of screen-reader text in navigation"

For icons in the navigation header it moves the screen reader text outside of the `<svg>` container so that valid HTML is generated by changing
[src/partials/navlist-item.html](https://github.com/corona-warn-app/cwa-website/blob/master/src/partials/navlist-item.html)

## Verification

Run https://validator.w3.org on https://www.coronawarn.app/en/ and check that the following errors are no longer shown:

- "HTML start tag span in a foreign namespace context."
- "Stray end tag svg"


brianebeling (approved):
Awesome find. Thanks!

MikeMcC399:
I checked that the validator no longer shows the related error messages and that the screen reader continues to read out the text for the sign language and simple language icons.

Issue #3227, opened 2022-11-18
Illegal alt attribute for video


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/results/#in_a_nutshell in "Explanatory video"
- https://www.coronawarn.app/de/faq/results/#in_a_nutshell in "Erklärvideo"

## Describe the issue

The `<video>` tag for the CWA Explanatory video contains the attribute `alt='Corona-Warn-App - explanatory film'`.

The `<video>` tag is not allowed to have an `alt` attribute so this is incorrect HTML code: 

`<video src='/assets/video/erklaerfilm_en.mp4' class='w-100' controls='' alt='Corona-Warn-App - explanatory film'></video>`

### Validator

Run https://validator.w3.org on https://www.coronawarn.app/en/faq/results/ and note error:

- "Attribute alt not allowed on element video at this point."

### Screen reader

In Windows 10 / 11

1. Open https://www.coronawarn.app/en/faq/results/#red_card in Google Chrome
2. Click on "In a nutshell" to expand
3. Enable Narrator: Windows key + Ctrl + Enter
4. Press 3 to move to "In a nutshell"
5. Press 3 until "Explanatory video" is reached and read out
6. Press Tab to move to video
7. Note that it reads "Group"

It does **not** read the alt text 'Corona-Warn-App - explanatory film'.

## Suggestion

Convert the `alt` attribute from the `<video>` content tab to an `aria-label` attribute.

## References

- [HTML `<video>` Tag](https://www.w3schools.com/tags/tag_video.asp)
- [Accessible Publishing Knowledge Base - Video examples](http://kb.daisy.org/publishing/docs/html/video.html#ex)

MikeMcC399:
- PR #3228 submitted to resolve this issue

MikeMcC399:
- The same problem in other places outside the FAQs is described in issue #3242.

PR #3228, opened 2022-11-18
Change video alt attribute to aria-label

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3227 "Illegal alt attribute for video" by converting the `alt` attribute from the `<video>` content tab to an `aria-label` attribute in:

- https://www.coronawarn.app/en/faq/results/#in_a_nutshell in "Explanatory video"
- https://www.coronawarn.app/de/faq/results/#in_a_nutshell in "Erklärvideo"

### Validator

Run https://validator.w3.org on https://www.coronawarn.app/en/faq/results/ and check that the following error is no longer produced:

- "Attribute alt not allowed on element video at this point."

### Screen reader

In Windows 10 / 11

1. Open https://www.coronawarn.app/en/faq/results/#red_card in Google Chrome
2. Click on "In a nutshell" to expand
3. Enable Narrator: Windows key + Ctrl + Enter
4. Press 3 to move to "In a nutshell"
5. Press 3 until "Explanatory video" is reached and read out
6. Press Tab to move to video
7. Note that it reads the `aria-label` text "Corona-Warn-App - explanatory film" followed by "Group"
8. Press Tab to access the play control
9. Press Space to play the video


brianebeling:
Hey, I was quickly researching and stumbled upon this https://www.w3.org/WAI/PF/HTML/wiki/Media_Alt_Technologies#1:_Use_.40aria-label_for_the_text_description_of_player

In there, they specifically label the video using the sentence: "hit space to toggle play/pause". 

I think we should either add that (or similar) to our label sentence or additionally use the  aria-describedby, if applicable.
What do you think?

MikeMcC399:
@brianebeling 

Your suggestion to use also `aria-describedby` sounds like a good enhancement. So far this PR was just a bug-fix. Let me try this out with a screen-reader. This is a new area for me.

brianebeling:
Thanks! Looking good otherwise. I think this is a great improvement already.

MikeMcC399:
@brianebeling 

I made a start by checking that the change to `aria-label` works using the built-in [Windows 11 Narrator](https://support.microsoft.com/en-us/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1).

I have never used this before and I found it not so intuitive. It's really necessary to go through the getting started guide.

I checked on other pages on the web site and there are other videos which need correcting as well because of their incorrect use of the `alt` attribute.

The exercise I am going through at the moment is to clean up the HTML only on the FAQ page, so I would like to suggest merging this PR as is (without the addition of `aria-describedby`) and then opening a new request to correct and enhance all video links on the web for accessibility and consistency. The resolution could include adding `aria-describedby` to all the other videos.

MikeMcC399:
@brianebeling 

I have added verification steps using the Windows 10 / 11 screen reader (Narrator) to the original post. 

Unfortunately your suggested text "hit space to toggle play/pause" does not provide the correct instructions for the Narrator in this instance and I don't know enough about this topic to be able to suggest the correct general solution which will work on all screen readers and for the other different videos on this website.

I will open a new issue to describe the incorrect use of the `alt` attribute together with other `<video>` tag usage outside of the FAQs. Again, this will just be a bug report with a suggestion for a bug fix rather than an enhancement.

brianebeling (approved):
Considering that we have a new Issue now, I will merge this PR and we can take a look at the narration and accessibility in the future again.

MikeMcC399:
Verified fixed with:

- https://validator.w3.org/
- Windows 11 Narrator (Windows key + Ctrl + Enter)

(details see original post)

Issue #3233, opened 2022-11-21
The titel Sections is not in the center-position 


MohamedZaghdoudi:
> Where to find the issue: 

https://www.coronawarn.app/en/faq/

> Describe the issue: 

the Position of Sections is not in the Center .

> Suggestion : 

switch in _reboot.scss the position from text-align from "$body-text-align " to "center"

MikeMcC399:
@MohamedZaghdoudi 

- Why is this closed? Normally you would close an issue after it has been resolved through merging a PR.

- I commented on the issue #3234 which you opened. Unfortunately it has some major side-effects.

- I have also noticed that there is a lot of use of the `<center>` tag on many pages, however this tag, which was already deprecated in HTML4, is not supported in HTML5, so in fact this is a general issue on the website.

MikeMcC399:
- See also related issue https://github.com/corona-warn-app/cwa-website/issues/3235

MohamedZaghdoudi:
@MikeMcC399 okay i will see it now

MikeMcC399:
@MohamedZaghdoudi 

Maybe you could also correct a typo in the subject?

 "tittel" should be "title"

MohamedZaghdoudi:
@MikeMcC399 thank you for your feedback . It is done.

MikeMcC399:
@MohamedZaghdoudi 

In English it is spelled "[title](https://dictionary.cambridge.org/dictionary/english/title)" not "titel".

brianebeling:
Closing this issue as it became a bit misleading. Originally we wanted to recenter the content under "Sections". It looks a bit misaligned. We did not want to center the headline itself. 

Please refer to: https://github.com/corona-warn-app/cwa-website/issues/3246

MikeMcC399:
@brianebeling 

I agree that this was a confusing issue. Also the suggested resolution to edit [_reboot.scss](https://github.com/corona-warn-app/cwa-website/blob/master/src/assets/scss/_reboot.scss) did not work and probably was not the right place to make a single layout change anyway.

[_reboot.scss](https://github.com/corona-warn-app/cwa-website/blob/master/src/assets/scss/_reboot.scss) looks like it comes from
https://github.com/twbs/bootstrap/blob/main/scss/_reboot.scss as described in https://getbootstrap.com/docs/4.0/content/reboot/.

PR #3234, opened 2022-11-21
Faq Sections in the center

MohamedZaghdoudi:
This PR resolves the issue #3233 that the Sections-Tittel not find in the center and its resolved by switching "body-text-align" with the position "center".


MikeMcC399:
@MohamedZaghdoudi 

- This change moves all FAQ content to be centered 🙁 

image


- Also it is mixed up with the addition of a Backup section.

MikeMcC399 (changes requested):
Please don't merge this as it has bad side-effects which cause the whole FAQ contents to be centered!

Issue #3236, opened 2022-11-21
Unclosed div sections


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/results/#where_can_i_get_tested
- https://www.coronawarn.app/de/faq/results/#where_can_i_get_tested

- https://www.coronawarn.app/en/faq/results/#in_a_nutshell
- https://www.coronawarn.app/de/faq/results/#in_a_nutshell

## Describe the issue

1. The `<div>` tag in https://www.coronawarn.app/en/faq/results/#where_can_i_get_tested in the second row with the image "Testzentren in Ihrer Nähe" is not closed.
2. The `<div>` tag in https://www.coronawarn.app/en/faq/results/#in_a_nutshell with the low risk image is not closed.

## Steps to reproduce

Run https://validator.w3.org/ on 

- https://www.coronawarn.app/en/faq/results/
- https://www.coronawarn.app/de/faq/results/

and note the error message:

"Unclosed element div"

## Suggestion

1. Add one more `</div>` end tag to the second row in https://www.coronawarn.app/en/faq/results/#where_can_i_get_tested with the image "Testzentren in Ihrer Nähe".

2. Add one more `</div>` end tag to the low risk image in https://www.coronawarn.app/en/faq/results/#in_a_nutshell.

MikeMcC399:
- PR #3237 submitted to resolve this issue

MikeMcC399:
Verified that https://validator.w3.org/ run on

- https://www.coronawarn.app/en/faq/results/
- https://www.coronawarn.app/de/faq/results/

no longer outputs the error message

"Unclosed element div"

following on the merge of PR #3237.

PR #3237, opened 2022-11-21
Close unclosed div in FAQs

MikeMcC399:
This PR resolves the issue https://github.com/corona-warn-app/cwa-website/issues/3236 "Unclosed div sections" in

- https://www.coronawarn.app/en/faq/results/#where_can_i_get_tested
- https://www.coronawarn.app/de/faq/results/#where_can_i_get_tested

- https://www.coronawarn.app/en/faq/results/#in_a_nutshell
- https://www.coronawarn.app/de/faq/results/#in_a_nutshell

where in each case three levels of `<div>`s are opened to display image with text, however only two levels are closed. A third `</div>` end tag is added to each problem area to resolve the unclosed `<div>` element messages.


brianebeling (approved):
Unless I made a mistake counting two times, this should be good to go. Thank you for the PR!

PR #3168, opened 2022-10-24
Show QR code in fullscreen

przybciu:
Issue #1150 
please review
It works in the following way:
after clicking create button to create qr code, the mouse become a pointer cursor and the qr preview can be clicked to enter full screen mode. You can exit by clicking esc button or by clicking "back" on mobile phone.

---
Internal Tracking ID: [EXPOSUREAPP-14227](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14227)

PR #3244, opened 2022-11-24
Show QR code in fullscreen

brianebeling:
Previous PR: https://github.com/corona-warn-app/cwa-website/pull/3168

"after clicking create button to create qr code, the mouse become a pointer cursor and the qr preview can be clicked to enter full screen mode. You can exit by clicking esc button or by clicking "back" on mobile phone." - [przybciu](https://github.com/przybciu)

This PR will exit draft state after https://github.com/corona-warn-app/cwa-website/pull/3168#issuecomment-1293090520 has been implemented.

---
Internal Tracking ID: [EXPOSUREAPP-14227](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14227)


Ein-Tim:
On which environments was this PR tested?

MikeMcC399:
@Ein-Tim 
> On which environments was this PR tested?

My understanding is that this is still a Work-in-Progress. You will probably need to wait for @brianebeling to add his improvements before testing again on Safari.

e
```diff
@@ -0,0 +1,12 @@
...
+* [32mQR-code-fullscreen[m
+  center_tag-refactoring[m
+  change-of-syntax[m
+  faq-backup-updaten[m
+  faq-updat-backup[m
+  format_option_for_downloads[m
+  master[m
+  syntax-chage[m
+  testes[m
+  your-new-Branch[m
```
MikeMcC399, line 12: @MohamedZaghdoudi 

This file should not be part of the PR.

dsarkar:
@Ein-Tim We haven been testing on different devices and browsers. Good news: Android works in general, iOS depends: on a newer iPad it works, but on a newer iPhone it does not. Table will be published here soon. Given that the normal use case are tabled for this functionality, I suggest we could merge. Maybe add a small sentence below to tap on the QR code for fullscreen.

brianebeling:
Can't review my own PR - but LGTM, tested locally. It was tested on different devices as well, confirming the table.

larswmh (approved):

Issue #3246, opened 2022-11-26
FAQ Sections is not centered


brianebeling:
## Where to find the issue
https://www.coronawarn.app/en/faq/

## Describe the issue

Centering:
![image](https://user-images.githubusercontent.com/12231155/204078797-4994bbc1-b610-4e60-86b8-a058d8b410dd.png)

## Suggested change
 Move the content (and title) to the right

Ein-Tim:
I think e.g. "Check-in" & "iOS" are one purpose only slightly visible, to make it clear to the user that there is more to that then only the topics that are shown. I really like this, as it is quite a low-key way to tell the user: "Hey, there's more!".

Centering the whole list makes sense.

Btw, I think in your title "had" should be "has".

MikeMcC399:
The logic seems to be a bit strange. The list can have indexed and non-indexed titles. Only the indexed titles are shown in https://www.coronawarn.app/en/faq/. If there are 7 or more titles (whether they are shown or not) then the last title in the list is faded out. It would be more intuitive to view if the number of displayed titles is limited to the same number. So you would need to count the number of displayed titles instead of the number of all titles.

MikeMcC399:
Please note when making changes to this layout that it is dynamic, according to the browser width.

brianebeling:
> I think e.g. "Check-in" & "iOS" are one purpose only slightly visible, to make it clear to the user that there is more to that then only the topics that are shown. I really like this, as it is quite a low-key way to tell the user: "Hey, there's more!".
> 
> Centering the whole list makes sense.
> 
> Btw, I think in your title "had" should be "has".

Whoops, you are right. I slipped, thanks for letting me know!

> The logic seems to be a bit strange. The list can have indexed and non-indexed titles. Only the indexed titles are shown in [coronawarn.app/en/faq](https://www.coronawarn.app/en/faq/). If there are 7 or more titles (whether they are shown or not) then the last title in the list is faded out. It would be more intuitive to view if the number of displayed titles is limited to the same number. So you would need to count the number of displayed titles instead of the number of all titles.

For now I removed the visibility thing. I wasn't realizing its a feature and not a bug. For some reason it does look off to me. At the moment I can't pintpoint what it is though..

MikeMcC399:
@brianebeling 

[...]

PR #3255, opened 2022-11-28
Center the FAQ lists

MikeMcC399:
This PR resolves the enhancement request https://github.com/corona-warn-app/cwa-website/issues/3246 "FAQ Sections is not centered".

The lower section of https://www.coronawarn.app/en/faq/ is now centered so that it has a consistent look independently of the available browser width.

## Old
---
Previously
image

## New
---
Desktop wide
FAQ wide

---
Desktop narrow
FAQ narrow

---
Mobile
FAQ mobile

### DIFF
--- a/src/partials/page-faq.html
+++ b/src/partials/page-faq.html
@@ -30,7 +30,7 @@
       <h1 class="headline headline-heavy mt-5 mb-3">{{page-contents.section-main.headline.subtitle}}</h1>
       <div class="row d-flex justify-content-center text-left pb-5 pt-5">
         {{#each page-contents.section-main.topics}}
-        <div class="col-lg-2 col-sm-6 col-md-3 col-xs-6">
+        <div class="pl-5 col-lg-2 col-sm-6 col-md-3 col-xs-6">
           <img src="{{icon}}" alt="{{title}}" height="29">
           <p><a href="results/?search=&topic={{id}}" class="no-decoration text-black"><b>{{{title}}}</b></a></p>
           <div class="{{#if sections.[6]}}section-column{{/if}}">
--- a/src/partials/page-faq.html
+++ b/src/partials/page-faq.html
@@ -1,7 +1,7 @@
 {{#if page-contents}}
-  <main class="main">
+  <main class="main text-center">
     <div class="container-fluid pt-5 pb-5">
-      <h1 class="headline headline-heavy text-center pb-3" id="top">{{page-contents.section-main.headline.title}}</h1>
+      <h1 class="headline headline-heavy pb-3" id="top">{{page-contents.section-main.headline.title}}</h1>
       <div class="d-flex justify-content-center">
         <form method="get" id="faq-search-form" class="container">
           <div class="mb-2 row justify-content-center">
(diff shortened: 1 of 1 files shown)


brianebeling:
Hey @MikeMcC399 

Thanks for the change. Unfortunately I'm unable to review it today as I would really like to take a more in-depth look.
I'll try to review and test it tomorrow. 

Specifically I'd like to review the centered title and see if we can find some alternatives. But I also want to take the opinion of others into consideration if possible.

mtb77:
@brianebeling just tell me when its ready to review :)

brianebeling:
@MikeMcC399 This might be a bit of a wild idea, but what if we remove the Sections title? I don't think it adds any value. Just looking at the table already suggests these are the categories or sections available.

brianebeling:
image
So basically this

MikeMcC399:
@brianebeling 

- I'm afraid that removing the "Sections" header is just scratching at the surface of issues on this page. It removes an `<h1>` tag which would then be no longer available for screen reader navigation. 
- I also wonder why the text "View all topics and questions" at the top shows all FAQs articles whereas "View all topics and questions" at the bottom of a column shows only the topics from the column. Why isn't the text at the top more prominent if it is more important? Why doesn't it have an `<hn>` tag? Why is this link closer to the search box when it belongs more to the bottom section? 
- The fact that the display of some titles in the columns is suppressed, but then they show up in the results page is non-intuitive. If they have a lower priority then they should be shown visually with a lower priority. 
- Then there is the issue of how many items are displayed in the column of lists and where it fades out.

I think it would be possible to spend quite a bit of time investigating and optimising these and related points.

For the moment I would restrict this PR to the centering topic only. If you want me to put the "Sections" header back on the left again, let me know and I will update the PR. I would however not want to remove it as an isolated change to this page.

brianebeling:
> Brian-J. Ebeling
> 
> * I'm afraid that removing the "Sections" header is just scratching at the surface of issues on this page. It removes an `<h1>` tag which would then be no longer available for screen reader navigation.
> * I also wonder why the text "View all topics and questions" at the top shows all FAQs articles whereas "View all topics and questions" at the bottom of a column shows only the topics from the column. Why isn't the text at the top more prominent if it is more important? Why doesn't it have an `<hn>` tag? Why is this link closer to the search box when it belongs more to the bottom section?
> * The fact that the display of some titles in the columns is suppressed, but then they show up in the results page is non-intuitive. If they have a lower priority then they should be shown visually with a lower priority.
> * Then there is the issue of how many items are displayed in the column of lists and where it fades out.
> 
> I think it would be possible to spend quite a bit of time investigating and optimising these and related points.
> 
> For the moment I would restrict this PR to the centering topic only. If you want me to put the "Sections" header back on the left again, let me know and I will update the PR. I would however not want to remove it as an isolated change to this page.

Yes, I agree that there are quite some issues. Then also not a lot of urgency to change much. It'd certainly improve the usability and visual appearance of the site, if it was a bit refined and weight of elements reconsidered.
Take the sections h1, why is it a h1? It shouldn't have the same weight or priority as the actual headline "How can we help you?".

The more I look at it, the more of an urge I get to work on this. 😆 

Ok. Back to the PR at hand. I think Sections on the left is the lesser evil at the moment. Layout wise it doesn't make sense to put it at the left, but it isolates the misaligned sections and feels less out of place to me.

Sadly the issue of the visual alignment still remains. If you measure the flexbox or inspect element, it is truly aligned. But it doesn't look aligned. That is because the rightmost column is shorter than the other columns, but they are equal width. As such creating a left-sided balance. 

I fixed it with a `pl-5`. That does the trick and also improves the mobile experience a bit.

image

MikeMcC399:
@brianebeling 

Please go ahead as you wish with this PR!

Issue #3242, opened 2022-11-23
Illegal alt attribute for other videos


MikeMcC399:
This is a follow-on from issue https://github.com/corona-warn-app/cwa-website/issues/3227 which covered a problem with the [Explanatory video](https://www.coronawarn.app/en/faq/results/#in_a_nutshell) in the FAQs.

This issue covers the same problem, but in other places on https://www.coronawarn.app.

## Where to find the issue

- https://www.coronawarn.app/en/science/2021-06-15-science-blog-1/
"What is the purpose of evaluating the CWA and what aspects play a role in the evaluation?"

- https://www.coronawarn.app/de/science/2021-06-15-science-blog-1/
"Über die Wirksamkeit und den Nutzen der Corona-Warn-App"

- https://www.coronawarn.app/de/sign-language/ "Gebärdensprache"

## Describe the issue

`<video>` tags on the above pages contain the attribute `alt=`. The `<video>` tag is not allowed to have an `alt` attribute so this is incorrect HTML code.

An activated screen reader does not read out any description of the video contents.

### Validator

Run https://validator.w3.org on 

- https://www.coronawarn.app/en/science/2021-06-15-science-blog-1/ (3 video errors)
- https://www.coronawarn.app/de/science/2021-06-15-science-blog-1/ (2 video errors)
- https://www.coronawarn.app/de/sign-language/ (1 video error)
  
and note the video related error message as follows:

- "Attribute alt not allowed on element video at this point."

### Screen reader

In Windows 10 / 11

1. Open corresponding web page in Google Chrome
2. Enable Narrator: Windows key + Ctrl + Enter
3. Navigate through document with numerical heading keys and tab
4. Move to video
5. Note that it reads (only) "Group"

It does **not** read the alt text.

## Suggestion

Convert the `alt` attribute from the `<video>` content tab to an `aria-label` attribute.

## References

- [HTML `<video>` Tag](https://www.w3schools.com/tags/tag_video.asp)
- [Accessible Publishing Knowledge Base - Video examples](http://kb.daisy.org/publishing/docs/html/video.html#ex)

MikeMcC399:
- I will wait for the outcome of PR https://github.com/corona-warn-app/cwa-website/pull/3228 before starting any work to resolve this one.

brianebeling:
[...]

PR #2481, opened 2022-02-25
Add missing alts to blog and science blog

larswmh:
Internal Tracking ID: [EXPOSUREAPP-11818](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11818)

PR #3256, opened 2022-11-28
Change video alt to aria-label in science and sign-language

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3242 "Illegal alt attribute for other videos".

For the `<video>` tags on the following pages, the `alt` attributes are changed to an `aria-label` attribute. This enables a screen reader such as Windows Narrator to read out the text property of the `aria-label`.

These are the pages affected:

- https://www.coronawarn.app/en/science/2021-06-15-science-blog-1/
"What is the purpose of evaluating the CWA and what aspects play a role in the evaluation?"

- https://www.coronawarn.app/de/science/2021-06-15-science-blog-1/
"Über die Wirksamkeit und den Nutzen der Corona-Warn-App"

- https://www.coronawarn.app/de/sign-language/ "Gebärdensprache" (also one typo on this page is corrected)

### Validator

Run https://validator.w3.org on 

- https://www.coronawarn.app/en/science/2021-06-15-science-blog-1/ (3 videos)
- https://www.coronawarn.app/de/science/2021-06-15-science-blog-1/ (2 videos)
- https://www.coronawarn.app/de/sign-language/ (1 video)
  
and ensure that the following video-related error message is no longer produced:

- "Attribute alt not allowed on element video at this point."

### Screen reader

In Windows 10 / 11

1. Open corresponding web page in Google Chrome
2. Enable Narrator: Windows key + Ctrl + Enter
3. Navigate through document with numerical heading keys and tab
4. Move to video
5. Note that it reads a description corresponding to the `aria-label` attribute property, followed by the word "Group"


MohamedZaghdoudi (commented):
@MikeMcC399 Very observative, thanks!

MikeMcC399:
This is really a bug.

1. It is a bug because `alt` is not allowed inside `<video>` .
2. Since the alt texts were added as part of PR https://github.com/corona-warn-app/cwa-website/pull/2481 by @larswmh which was an accessibility effort, I assume that this was a mistake, so this current PR corrects a bug in the accessibility implementation, rather than an enhancement.

brianebeling:
> This is really a bug.
> 
> 1. It is a bug because `alt` is not allowed inside `<video>` .
> 2. Since the alt texts were added as part of PR [Add missing alts to blog and science blog #2481](https://github.com/corona-warn-app/cwa-website/pull/2481) by Lars which was an accessibility effort, I assume that this was a mistake, so this current PR corrects a bug in the accessibility implementation, rather than an enhancement.

Ok. I was thinking enhancement because of the added accessibility, e.g. narrator. Will change the label then.

MikeMcC399:
@brianebeling 

> Ok. I was thinking enhancement because of the added accessibility, e.g. narrator. Will change the label then.

Thanks for changing the label! I understand why you classified it as an enhancement. Providing the PR gets merged, the category is not really so important.

brianebeling (approved):

Issue #3262, opened 2022-11-30
Stranded `</span>` tag in /faq/results/ breadcrumbs


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/results/
- https://www.coronawarn.app/de/faq/results/

## Describe the issue

The breadcrumbs / search result count display in `/faq/results/` includes one too many end `</span>` tags.

## Steps to reproduce

Run https://validator.w3.org/ on https://www.coronawarn.app/en/faq/results/ and note error message:

"Stray end tag span."

## Suggestion

Remove wrong additional `</span>` end tag from the end of:

https://github.com/corona-warn-app/cwa-website/blob/5e3765569242be8fb015b08fc5267e65df5bda3c/src/partials/page-faq-results.html#L50

MikeMcC399:
- PR #3263 submitted to resolve this issue

PR #3263, opened 2022-11-30
Remove extra span end tag from faq breadcrumbs

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3262 "Stranded `</span>` tag in /faq/results/ breadcrumbs".

It removes the stranded `</span>` tag from the end of the breadcrumbs / search result count display in `/faq/results/` created through [/src/partials/page-faq-results.html](https://github.com/corona-warn-app/cwa-website/blob/master/src/partials/page-faq-results.html).

## Verification

Run https://validator.w3.org/ on https://www.coronawarn.app/en/faq/results/ and check that the following error message is no longer produced:

"Stray end tag span."


brianebeling (approved):
Looking good, thanks!

Issue #3264, opened 2022-11-30
Nested main element in faq


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/
- https://www.coronawarn.app/de/faq/

## Describe the issue

The FAQ page https://www.coronawarn.app/en/faq/ contains two `<main>` elements. This is not allowed:

```html
    <main id="main" class="main-content" role="main" tabindex="-1">
        <main class="main text-center">
```

## Steps to reproduce

Run https://validator.w3.org/ on https://www.coronawarn.app/en/faq/ and note the error messages:

- "The main element must not appear as a descendant of the main element."
- "A document must not include more than one visible main element"

also

- "Stray end tag div."

## Analysis

The first main definition comes from:
https://github.com/corona-warn-app/cwa-website/blob/5e3765569242be8fb015b08fc5267e65df5bda3c/src/layouts/default.html#L30

- The `class="main-content"` attribute statement does not seem to be used.
- According to https://w3c.github.io/html-aria/#el-main it is not recommended to explicitly assign `role="main"` since it is already implicitly assigned.

The second main definition comes from:
https://github.com/corona-warn-app/cwa-website/blob/5e3765569242be8fb015b08fc5267e65df5bda3c/src/partials/page-faq.html#L2

- The `main` class is not defined. It may not be needed.
- The `text-center` class is from bootstrap.

## Suggestion

Convert the `<main>` tag in [src/partials/page-faq.html](https://github.com/corona-warn-app/cwa-website/blob/5e3765569242be8fb015b08fc5267e65df5bda3c/src/partials/page-faq.html) to a `<div>` tag and combine this with the stray `</div>` end tag.

Review the attribute settings `class="main-content"` `role="main"` in src/layouts/default.html in a separate issue.

## References

- https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element

MikeMcC399:
- PR #3265 submitted to resolve this issue

PR #3265, opened 2022-11-30
Replace child main in faq by div section

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3264 "Nested main element in faq".

The child `main` element is converted to a `div` element. At the same time, a stray `</div>` end tag is cleaned up.

Since the indenting was not correct and therefore hard to read, I added a second commit with a combination of automatic and manual reformatting.

View commit https://github.com/corona-warn-app/cwa-website/commit/20abd0b22e048fb5b512f6b80b4f0befb020cec6 to look at the changes which are not caused by reformatting.

## Verification

Run https://validator.w3.org/ on https://www.coronawarn.app/en/faq/ and the following error messages should no longer appear:

- "The main element must not appear as a descendant of the main element."
- "A document must not include more than one visible main element"
- "Stray end tag div."

### DIFF
--- a/src/partials/page-faq.html
+++ b/src/partials/page-faq.html
@@ -1,5 +1,5 @@
 {{#if page-contents}}
-  <main class="main text-center">
+  <div class="text-center">
     <div class="container-fluid pt-5 pb-5">
       <h1 class="headline headline-heavy pb-3" id="top">{{page-contents.section-main.headline.title}}</h1>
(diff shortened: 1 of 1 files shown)


brianebeling (approved):
Tested locally, no observable errors and validation runs fine. Thanks!

Issue #3276, opened 2022-12-05
Cypress production test runs with older Node.js 16 version


MikeMcC399:
## Where to find the issue

[.github/workflows/cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml)

## Describe the issue

The workflow "Cypress Test Production" [.github/workflows/cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml) runs using the Node.js version v16.13.0. This is not the Node.js 18 `lts/hydrogen` version used in the other workflows, [build.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) and [deploy-master.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/deploy-master.yml), on [cwa-website](https://github.com/corona-warn-app/cwa-website).

## Steps to reproduce

View the action logs https://github.com/corona-warn-app/cwa-website/actions/workflows/cypress-test-prod.yml

## Suggestion

Currently there are no errors caused by using the older Node.js version v16.13.0 to check the production website. The workflow does not need to build the web. It only checks the existing published website.

The workflow "Cypress Test Production" [.github/workflows/cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml) uses the action [cypress-io/github-action@v4](https://github.com/cypress-io/github-action) which is set up to use `node16`. There is a restriction in this action that means it is currently not possible to run it under Node.js 18. See issue https://github.com/cypress-io/github-action/issues/642.

For the moment I suggest no change, however the GitHub repository https://github.com/cypress-io/github-action has quite a backlog of issues and updates necessary. If these become a problem then it would be possible to replace the use of [cypress-io/github-action@v4](https://github.com/cypress-io/github-action) by a direct call to Cypress (`npx cypress`), providing [actions/setup-node](https://github.com/actions/setup-node) and `npm ci` are run before the cypress call.

MikeMcC399:
This is a "for info" issue. I will close it and only re-open it if there is some additional action necessary.

brianebeling:
I think this is still relevant to an extend. Its not a pressing issue, but something we should monitor.

MikeMcC399:
@brianebeling 

[...]

PR #3277, opened 2022-12-05
Cypress test remove dependency on cypress-io/github-action

MikeMcC399:
This PR resolves issue https://github.com/corona-warn-app/cwa-website/issues/3276 "Cypress production test runs with older Node.js 16 version" by removing the dependency of the workflow on the action [cypress-io/github-action@v4](https://github.com/cypress-io/github-action) and replacing it by a direct call of Cypress.

The Cypress call is preceded by a call to [actions/setup-node](https://github.com/actions/setup-node) to install the Node.js `lts/hydrogen` version 18.12.1 and by a call to `npm ci` to install npm dependencies, including Cypress. The environment variable `TERM: xterm` is added to prevent a spurious warning `tput: No value for $TERM and no -T specified` from Cypress running in headless mode.

## Verification

Check the version of Node.js reported by Cypress is as follows:

"Node Version:   v18.12.1 (/opt/hostedtoolcache/node/18.12.1/x64/bin/node)"

and that the Cypress run is successful:

"All specs passed!"

### DIFF
--- a/.github/workflows/cypress-test-prod.yml
+++ b/.github/workflows/cypress-test-prod.yml
@@ -13,15 +13,18 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NO_COLOR: true
+      TERM: xterm
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
 
-      - name: Cypress run
-        uses: cypress-io/github-action@v4
+      - name: Setup Node.js 18 environment
+        uses: actions/setup-node@v3
         with:
-          browser: chrome
-          headed: false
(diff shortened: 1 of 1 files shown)


MikeMcC399:
The action was successfully tested in the fork. See

https://github.com/MikeMcC399/cwa-website/actions/runs/3620471314 => https://github.com/MikeMcC399/cwa-website/actions/runs/3620471314/jobs/6102730118

View Cypress-Test-Production > Cypress run

brianebeling (approved):

Issue #3071, opened 2022-08-04
Cypress visit to FAQ articles times out using short baseUrl


MikeMcC399:
## Where to find the issue

https://coronawarn.app/en/faq/results

## Describe the issue

If two FAQ articles with a non-legacy FAQ-URL in the format `/en/faq/results/#faq_article` are visited in a Cypress test, and the `baseUrl` is set to the short value `https://coronawarn.app`, then the second visit times out with the error:

```
Timed out after waiting 100000ms for your remote page to load.
Your page did not fire its load event within 100000ms.
```

## Steps to reproduce

Save the following code into `cypress/e2e/consecutive_faqs.cy.js` 

```
it("results FAQs", () => {
    cy.visit("/en/faq/results/#android_location");
    cy.visit("/en/faq/results/#cause9002_recovery");
});
```

Run it with
`npx cypress open -c baseUrl=https://coronawarn.app`
and select the test `consecutive_faqs.cy.js` to run

Note the error message:

```
visit /en/faq/results/#cause9002_recovery -> 301: https://www.coronawarn.app/en/faq/results/
CypressError
Timed out after waiting 100000ms for your remote page to load.

Your page did not fire its load event within 100000ms.
```

HTTP 301 means "Moved Permanently".

![Page did not fire load event](https://user-images.githubusercontent.com/66998419/182814393-4ff74229-cfca-4bcb-b421-869f28aa7506.jpg)

## Expected behaviour

Visiting FAQ URLs `/en/faq/results/#article_1` then `/en/faq/results/#article_2` with `baseUrl=https://coronawarn.app` in a Cypress test with `cy.visit` should not result in a page load timeout. It should be possible to visit the pages without errors.

## Additional information

The problem does not occur if:

- `baseUrl` is set to `baseUrl=https://www.coronawarn.app` instead of `baseUrl=https://coronawarn.app`
- `npm run test:open` is used and the test runs on `http://localhost:8000/`
- `/results` is removed from the first or second URL or from both URLs keeping `baseUrl=https://coronawarn.app`

## Workaround

When testing the production server use `baseUrl=https://www.coronawarn.app` instead of `baseUrl=https://coronawarn.app`.

---
Internal Tracking ID: [EXPOSUREAPP-14103](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14103)

PR #3288, opened 2022-12-12
Change Cypress baseURL to https://www.coronawarn.app

larswmh:
This PR implements a workaround proposed in #3071.

When visiting FAQ articles with a non-legacy FAQ-URL in a Cypress test with the baseURL https://coronawarn.app, the second visit times out. If the URL is replaced with https://www.coronawarn.app, it doesn't time out. Steps to reproduce can be found in the aforementioned issue.

---
Internal Tracking ID: [EXPOSUREAPP-14103](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14103)

### DIFF
--- a/.github/workflows/cypress-test-prod.yml
+++ b/.github/workflows/cypress-test-prod.yml
@@ -27,4 +27,4 @@ jobs:
         run: npm ci
 
       - name: Cypress run
-        run: npx cypress run -s 'cypress/e2e/*.js' -c baseUrl=https://coronawarn.app --e2e --headless --browser chrome 
+        run: npx cypress run -s 'cypress/e2e/*.js' -c baseUrl=https://www.coronawarn.app --e2e --headless --browser chrome 
--- a/README.md
+++ b/README.md
@@ -125,9 +125,9 @@ Best practice is to use `data-e2e="your_test_id"` element attributes to select s
 
 To run all tests included in Cypress Test Production execute:
 
-`npx cypress run -s 'cypress/e2e/*.js' -c baseUrl=https://coronawarn.app --headed` => test results are printed in the console, also you can see browser's movements
+`npx cypress run -s 'cypress/e2e/*.js' -c baseUrl=https://www.coronawarn.app --headed` => test results are printed in the console, also you can see browser's movements
 
-`npx cypress run -s 'cypress/e2e/*.js' -c baseUrl=https://coronawarn.app --headless --browser chrome` => test results are printed in the console, the browser is not displayed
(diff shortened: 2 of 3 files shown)


MikeMcC399:
Hi @larswmh 

I don't think that this PR should close the issue https://github.com/corona-warn-app/cwa-website/issues/3071 since it is only a workaround, unless you say that the underlying issue is not going to be fixed.

MikeMcC399:
The Android app links to https://www.coronawarn.app so this PR does make sense to align the testing `baseURL` to the one used by the app.

See https://github.com/corona-warn-app/cwa-app-android/blob/main/Corona-Warn-App/src/main/res/values/links.xml

MikeMcC399:
@larswmh 

Here is some background for your comment ...

> In my research I found something interesting:
> If you copy the URL from the URL bar of your browser to somewhere else, it does not always give the same result. For example:

> Enter github.com: URL turns into https://github.com/
> Enter coronawarn.app: URL turns into https://www.coronawarn.app/en/

That depends on your browser, the browser settings and on the server. Using Google Chrome and with F12 developer showing the Network tab.

Chrome defaults to https: by default (see [blog](https://blog.chromium.org/2021/03/a-safer-default-for-navigation-https.html#:~:text=Starting%20in%20version%2090%2C%20Chrome's,%E2%80%9D%20or%20%E2%80%9Chttps%3A%2F%2F%E2%80%9D)).

so my browser converts `github.com` to  https://github.com/ before it sends a request, which is answered directly and successfully:

```
Request URL: https://github.com/
Request Method: GET
Status Code: 200 
```

If I enter `coronawarn.app` my browser requests https://coronawarn.app/

```
Request URL: https://coronawarn.app/
Request Method: GET
Status Code: 301 Moved Permanently
Remote Address: 80.158.41.247:443
```
with Response Header including

```
Location: https://www.coronawarn.app/
Server: nginx
```

so now my browser sends
```
Request URL: https://www.coronawarn.app/
Request Method: GET
Status Code: 200 OK
```
which is then intercepted by
Request URL: https://www.coronawarn.app/assets/js/static/redirect.js
which redirects to
```
Request URL: https://www.coronawarn.app/en/
Request Method: GET
Status Code: 200 OK
```

So the underlying problem may be caused by the code in `redirect.js`. I think it is OK not to solve the root cause. The benefit is probably marginal and I assume that the website has a limited lifetime.

MikeMcC399 (approved):
@larswmh 

> As I realistically don't see this to be changed since it did not bring up any other issues so far, I think this is a sufficient solution.

That is OK with me.

I also tested that it works!

larswmh:
@brianebeling @MikeMcC399 

Thanks for the reviews

PR #3295, opened 2022-12-20
Update package-lock.json

brianebeling:
Fixes a secrurity issue and also general housekeeping of the package.json/package-lock.json

---
Internal Tracking ID: [EXPOSUREAPP-14494](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14494)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3429,12 +3429,12 @@
       }
     },
     "node_modules/browser-sync": {
-      "version": "2.27.10",
-      "resolved": "https://registry.npmjs.org/browser-sync/-/browser-sync-2.27.10.tgz",
-      "integrity": "sha512-xKm+6KJmJu6RuMWWbFkKwOCSqQOxYe3nOrFkKI5Tr/ZzjPxyU3pFShKK3tWnazBo/3lYQzN7fzjixG8fwJh1Xw==",
+      "version": "2.27.11",
+      "resolved": "https://registry.npmjs.org/browser-sync/-/browser-sync-2.27.11.tgz",
+      "integrity": "sha512-U5f9u97OYJH66T0MGWWzG9rOQTW6ZmDMj97vsmtqwNS03JAwdLVES8eel2lD3rvAqQCNAFqaJ74NMacBI57vJg==",
       "dependencies": {
-        "browser-sync-client": "^2.27.10",
-        "browser-sync-ui": "^2.27.10",
+        "browser-sync-client": "^2.27.11",
+        "browser-sync-ui": "^2.27.11",
         "bs-recipes": "1.3.4",
         "bs-snippet-injector": "^2.0.1",
         "chokidar": "^3.5.1",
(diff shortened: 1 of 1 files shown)


brianebeling:
Can someone run

```
npx cypress run -s 'cypress/e2e/*.js' -c baseUrl=https://www.coronawarn.app --headless --browser chrome
```
for me it fails at the appstore buttons (the very first test) and I can't seem to figure out why. It happens with a clean branch from master as well..

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "corona-warn-app-landingpage",
   "version": "1.1.0",
```
MikeMcC399, line 3: I have not seen this before.

Running Node.js 18 LTS generates lockfileVersion 2, not 3.

MikeMcC399:
@brianebeling 

> It happens with a clean branch from master as well..

That suggests you have a problem with your local system. Good luck with troubleshooting and fixing!

package-lock.json
brianebeling: Here is some documentation

https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#lockfileversion

I suspect some command updated it to 3, I'm not entirely sure when it happened. I will reinvestigate.. or just change it back to 2, as that seems to be the default. Possibly it was `ncu` or the `npm outdated` and `npm update` command.

brianebeling:
> Brian-J. Ebeling
> 
> > It happens with a clean branch from master as well..
> 
> That suggests you have a problem with your local system. Good luck with troubleshooting and fixing!

Well.. as long as it works for you. I will check on another workstation and reinstall some dependencies, in the hope of fixing it. Thanks for the good wished 😆

MikeMcC399:
@brianebeling 

I found that if you use npm 9 you will get `"lockfileVersion": 3`  if you do an `npm install`.

For consistency however we need to use the Node.js 18 LTS version which includes node 18.12.1 and npm 8.19.2. This is what is advised in https://github.com/corona-warn-app/cwa-website/blob/master/README.md#requirements.

If you need to use different versions of node and npm for other purposes and projects then I would recommend installing nvm to switch easily between the versions.

I don't think you should manually edit the "lockfileVersion". Instead, make sure you are working with npm 8 and start again from the master branch of cwa-website.

MikeMcC399 (changes requested):
Any updates to [package-lock.json](https://github.com/corona-warn-app/cwa-website/blob/master/package-lock.json) need to be made using Node.js LTS

See https://nodejs.org/en/download/
Latest LTS Version: 18.12.1 (includes npm 8.19.2)

brianebeling:
> Any updates to [package-lock.json](https://github.com/corona-warn-app/cwa-website/blob/master/package-lock.json?rgh-link-date=2022-12-21T10%3A14%3A45Z) need to be made using Node.js LTS
> 
> See [nodejs.org/en/download](https://nodejs.org/en/download/) Latest LTS Version: 18.12.1 (includes npm 8.19.2)

I found out what went wrong. On this machine I seem to have installed node.js parallel to nvm and it would override the selected version. I fixed the issues that were caused and force pushed after resetting. Should be good to review now.

There is also a follow up PR that updates a lot of dependencies using `npm update` #3296

MikeMcC399:
I would tend to merge the two current PRs into one.

dsarkar:
Good Morning @MikeMcC399, thanks for assisting here and the other PR. Closing this one as you suggested.

PR #3298, opened 2022-12-23
 [R 3.0] Screenshots

dsarkar:
EN:
image
image
image

DE:
image
image
image

---
Internal Tracking ID: [EXPOSUREAPP-14200](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14200)


brianebeling:
Todo:

- [ ] Rename folder from 2.29 to 3.0
- [ ] Check references to 2.29 folder and adjust to 3.0

MikeMcC399:
- Unfortunately this PR will make the issue https://github.com/corona-warn-app/cwa-website/issues/2699 worse, since it adds hundreds of new screenshots to the tens of thousands of existing screenshots.

Ein-Tim:
Wouldn't it make sense to use an approach were screenshots of screens which were not changed by the update are just continued to be used from the versions before? 

I hope you understand what I'm trying to say? 😅

larswmh:
@Ein-Tim @MikeMcC399 
Thanks for sharing your concerns. I can assure you that we are actively working on the problem. See for example #3327, where we tested symlinking duplicate files. We're also taking Git LFS into consideration, but cannot decide that without further internal approvals. For the upcoming release we will stick with the usual procedures.

dsarkar:
LGTM, can be merged

PR #3299, opened 2023-01-01
Bump copyright year in README.md

Ein-Tim:
## Description

This PR bumps the copyright year in the README.md file.

Happy new year! 🎉

### Note

This PR should only be merged once a change to any content of this repository was done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ A list of all public repositories from the Corona-Warn-App can be found [here](h
 
 ## Licensing
 
-Copyright (c) 2020-2022 Deutsche Telekom AG and SAP SE or an SAP affiliate company.
+Copyright (c) 2020-2023 Deutsche Telekom AG and SAP SE or an SAP affiliate company.
 
 Licensed under the **Apache License, Version 2.0** (the "License"); you may not use this file except in compliance with the License.
 


brianebeling (approved):
Same as in the other PRs of the same kind already, thanks for the change. I was unable to find more places that would require a change.

Ein-Tim:
@MikeMcC399 To me, it looks like the NOTICE file of this repository was never correctly established. When comparing the file from this repository to [the file in the documentation repository ](https://github.com/corona-warn-app/cwa-documentation/blob/main/NOTICE), it looks like the contributors are missing. But also there, in the documentation repo, the copyright year has never been changed. I'm unsure if this file is really relevant.

brianebeling:
Honestly, I'm not sure. I will forward it and let you know.

I'd suggest we merge this and make a separate or for the notice.

Also Happy New Year to both of you. I will be on vacation for around two weeks and only check in here irregularly, but @dsarkar is back in action.  🎉

MikeMcC399:
@Ein-Tim / @brianebeling 

https://www.apache.org/licenses/LICENSE-2.0 in the section "HOW TO APPLY THE APACHE LICENSE TO YOUR WORK" says:
"Include a copy of the Apache License, typically in a file called LICENSE, in your work, and consider also including a NOTICE file that references the License."

however since the suggested text is already in the [README, Licensing section](https://github.com/corona-warn-app/cwa-website/blob/master/README.md#licensing) and [NOTICE](https://github.com/corona-warn-app/cwa-website/blob/master/NOTICE) is incorrectly completed, then I suggest that the [NOTICE](https://github.com/corona-warn-app/cwa-website/blob/master/NOTICE) file is deleted as part of any Copyright update review.

larswmh (approved):
Thanks for the PR @Ein-Tim! 

I'd like to merge this now that we've had a change to the FAQ. The concerns about the [NOTICE](https://github.com/corona-warn-app/cwa-website/blob/master/NOTICE) file have been forwarded and we will take action if necessary, thanks for the information @MikeMcC399.

MikeMcC399:
@larswmh 

If the README only refers to itself:

"Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License."

and the NOTICE file refers to "them" without specifying what "them" means as in:

"you may not use them except in compliance with the License."

then this should be cleaned up.

My reading of opinions about Copyright notices says that they are not really necessary, so if they are not necessary, then it doesn't matter if they are clean or not, however this should be sorted out by legal experts.

Thank for passing on the concerns in any case.

Issue #3308, opened 2023-01-05
Improve testing for PR submission


MikeMcC399:
## Current Implementation

The cwa-website can be tested locally and manually using `npm test` which runs a full set of Cypress tests.

When a pull request is opened or changed then the workflow [build](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) is run.

On a daily basis the workflow [cypress-test-prod](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml) is run which executes a subset of Cypress tests.

The advice in [README: Testing](https://github.com/corona-warn-app/cwa-website/blob/master/README.md#testing) is as follows:

"To minimize the occurrence of errors we would ask you to perform all tests when contributing to our repository."

## Problems with current implementation

CI testing only checks that the web builds. It does not run any Cypress tests. PRs which cause the daily Cypress test to fail can be merged without any warning appearing.

`npm test` takes a long time and is flaky due to the test [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js), which can even cause an out-of-memory browser issue when run.

Other test methods need a long string of manual commands which are not easy to remember.

## Suggested Enhancement

1. Add a new script `npm run-script test:short` to make it easier to test before creating a new PR. This should run all tests except the long-running and flaky test [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js) which tests external links. The new test should run in the Electron browser built in to Cypress so the test can run under any of the supported operating systems (Windows, Ubuntu, macOS) without installing additional browsers.
2. Make pull requests cause a Cypress test-run to be executed. This should be the same set of tests as in `npm run-script test:short` (in other words without the external link test), running Cypress headless using the Chrome browser.<br>I suggest to add a new workflow `build-and-test` and leave the current [build](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build.yml) workflow in place, but disable it in the Actions menu under https://github.com/corona-warn-app/cwa-website/actions/workflows/build.yml. That way if there are problems it would be easy to switch back to the original build workflow without testing.
3. Document this in the README file.

## Expected Benefits

[...]

PR #3309, opened 2023-01-05
Add short Cypress test script npm run test:short

MikeMcC399:
This PR implements one suggestion from https://github.com/corona-warn-app/cwa-website/issues/3308 "Improve testing for PR submission".

It provides a new script which can be run with:

`npm run test:short`

This runs all Cypress tests except [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js).

## Verification

Execute

`npm run test:short`

and confirm that there are no errors reported.

---
Internal Tracking ID: [EXPOSUREAPP-14530](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14530)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "corona-warn-app-landingpage",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "description": "Corona-Warn-App website coronawarn.app",
   "main": "gulpfile.js",
   "scripts": {
@@ -11,6 +11,7 @@
     "cypress:open": "cypress open --e2e --browser chrome",
     "cypress:run:chrome": "cypress run --browser chrome --headed",
     "cypress:run:firefox": "cypress run --browser firefox --headed",
+    "cypress:run:short": "cypress run -s 'cypress/e2e/*.js' --e2e --headless",
     "start-server": "gulp start-server",
     "test:open:phase2": "start-server-and-test start-server http://localhost:8000 cypress:open",
     "test:open": "run-s build test:open:phase2",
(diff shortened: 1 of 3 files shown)


MikeMcC399:
Unfortunately there is a problem with the syntax running on Ubuntu. It runs, but gives a warning:

```
⚠ Warning: It looks like you're passing --spec a space-separated list of arguments:

"cypress/e2e/app_to_web.cy.js cypress/e2e/applink.cy.js cypress/e2e/blog.cy.js cypress/e2e/check_anchor_links.cy.js cypress/e2e/check_videos.cy.js cypress/e2e/eventRegistration.cy.js cypress/e2e/faq.cy.js cypress/e2e/faq_link_attr.cy.js cypress/e2e/hotline.cy.js cypress/e2e/mime.cy.js"

This will work, but it's not recommended.

If you are trying to pass multiple arguments, separate them with commas instead:
  cypress run --spec arg1,arg2,arg3
  
The most common cause of this warning is using an unescaped glob pattern. If you are
trying to pass a glob pattern, escape it using quotes:
  cypress run --spec "**/*.spec.js"
```

There is no warning on Windows, which is why I didn't notice it before I submitted the PR.

MikeMcC399:
`npm run test:short` is successful on:

- Ubuntu 22.04
- Windows 11

with Node.js 18.12.1

Perhaps somebody would be kind enough to test it on macOS? @Ein-Tim? @larswmh?

larswmh (approved):
macOS Ventura 13.1, Apple M1 Pro
Node.js 18.12.1

image

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "corona-warn-app-landingpage",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "description": "Corona-Warn-App website coronawarn.app",
   "main": "gulpfile.js",
```
larswmh, line 5: @MikeMcC399 

Since gulpfile.js was renamed to gulpfile.mjs, I think it would be good to rename it here as well. If I have grasped this correctly, this isn't changing anything in functionality and is just for informative purposes, but it would still be good to have the correct file extension in there if we're already making changes to the file

MikeMcC399: @larswmh 

Good catch! I don't believe this is used anywhere but I have corrected it anyway in https://github.com/corona-warn-app/cwa-website/pull/3309/commits/d78aa2d1b2a6ae69daedb5cedb68665b87009bd4.

dsarkar:
@MikeMcC399 Many thanks for this PR. (On my older 2.7 GHz i5 Mac the test still takes about 10 min...)

MikeMcC399:
@dsarkar 
> Many thanks for this PR. 
You're welcome!

> (On my older 2.7 GHz i5 Mac the test still takes about 10 min...)

The run time is quite dependent on the available hardware. On my Windows 11 laptop (SSD drive, intel i5, 8GB) the Cypress test shows "All specs passed 02:47". The screenshot from @larswmh was even faster with 01:24.

Maybe you need to speak to someone nicely to get a hardware upgrade sometime? 😉

PR #3323, opened 2023-01-12
Uniformly use "Risiko-Ermittlung" (instead of "Risikoermittlung") in the German FAQs

Ein-Tim:
## Description

This PR changes all occurrences of "Risikoermittlung" to "Risiko-Ermittlung" in the German FAQ, as this is the name used in the app.

PR #3321, opened 2023-01-11
Release 3.0 | Update TAN FAQs

MohamedZaghdoudi:
The PR is about the updates  of the Release 3.0 in FAQ-EN and FAQ-DE 

The date at the top is a placeholder. This is so that we can replace the dates of all affected FAQs at once when they're in the 3.0 branch.

image

image

---
Internal Tracking ID: [EXPOSUREAPP-14496](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14496)

PR #3324, opened 2023-01-13
Update workflows: Include build job into build-and-test

larswmh:
This PR modifies the `build-and-test.yml` workflow so that it includes two jobs:
- build: the required job to merge PRs into master
- test: the optional job to test PRs with Cypress

---
Internal Tracking ID: [EXPOSUREAPP-](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-)


dsarkar (approved):

MikeMcC399:
@larswmh 

It's interesting to see in https://github.com/corona-warn-app/cwa-website/pull/3323 that the "build" job is required and the "test" job is not required. Again, I'm guessing that this is part of the branch protection rules which are not visible to me.

image

larswmh:
@MikeMcC399 

> It's interesting to see in #3323 that the "build" job is required and the "test" job is not required. Again, I'm guessing that this is part of the branch protection rules which are not visible to me.
> 
> image

In the branch protection rules, the `build` job is set as required. Not the `build.yml` workflow file. This way I could just move it to the build-and-test.yml without problems.
image

MikeMcC399:
@larswmh 

Thank you for the screenshot! I imagined that it must be like this, however I had no way to view it.

It might be good at a later stage to add the `test` job to the required positive status outcomes. It is probably good for the moment not to require it, until there is more experience with the job. Whoever merges a PR can make the decision manually if they want to merge a PR which fails the Cypress test (like https://github.com/corona-warn-app/cwa-website/pull/3321).

PR #3338, opened 2023-01-19
Update blog: Add categorization below search bar

larswmh:
This PR adds a new way to quickly navigate to specific blogs.

EN:
image
image

DE: 
image
image

---
Internal Tracking ID: [EXPOSUREAPP-13933](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13933)


MikeMcC399:
This looks very busy and it seems that it has been generated by hand, and would need to be maintained by hand. 

I don't think I would use it instead of the free-form search.

larswmh:
Thanks for your feedback @MikeMcC399.

You're right, this needs to be maintained by hand. However, considering the frequency of blogs and the future of the app in general, we think this is manageable. 
We think that this feature can be beneficial for some, maybe you should compare it to the FAQ section menu. Of course you could just search for what you're looking for, but it can always be useful to have a compact overview of all relevant contents.

dsarkar (approved):

MikeMcC399:
This PR contains link failures. See issue https://github.com/corona-warn-app/cwa-website/issues/3383.

Only the Cypress `npm run test:short` is run automatically.
Especially for PRs which make link changes it is recommended to additionally run the full  `npm test` when submitting.

dsarkar:
@MikeMcC399 Thanks for pointing out the issue!

MikeMcC399:
@dsarkar 
> Thanks for pointing out the issue!

You're welcome! I was only wanting to check if the Cypress tests still run with the latest version of Cypress (12.5.1) and I was a bit surprised to find the bad links.

I will make a suggestion to add a manual workflow to check these links, so that it can be done more easily.

PR #3343, opened 2023-01-21
Remove stray ul end tag from end of link list on #red_card_how_to_test (EN)

MikeMcC399:
This PR resolves a stray tag `</ul>` in https://www.coronawarn.app/en/faq/results/#red_card_how_to_test.

## Verification

https://validator.w3.org/ on https://www.coronawarn.app/en/faq/results/ should show no error

`Stray end tag ul.`


dsarkar:
@MikeMcC399 Thanks for the PR!

PR #3347, opened 2023-01-25
Dashboard: Disabled stacking for line charts

thorbenkuro:
This PR changes the presentation of line charts on the Dashboard. 
Some charts, like the "Sharing behaviour"-chart for QR codes presented the daily data in a Bar chart, when filtering for "Last 28 days". The presentation changed to a line chart, when filtering the "last 6 months"/"Total", but kept the stacking behaviour of the bar chart.
When using a line chart it is not common to stack the data-sets this way, which might lead to confusion. That is why I removed the stacking behaviour for the aforementioned line charts in this PR.

Current (stacked) line chart:
Line Chart stacked

New (not stacked) line chart:
Line Chart noStacked


thorbenkuro:
Right now GitHub Actions seems to be unavailable, according to the error message of the build pipeline. I will trigger the checks again as soon as possible.

brianebeling (approved):

larswmh (approved):

brianebeling:
The chart in question has been updated in the meanwhile to be more representative (bar chat)

PR #3350, opened 2023-01-25
Correct use of hyphens

dsarkar:
Correct use of hyphens


MikeMcC399:
This PR also corrects German grammar (use of gleich and selber). This is missing from the PR description.

dsarkar:
@MikeMcC399 right, description updated in OP

Issue #3383, opened 2023-02-13
Multiple bad blog "Other topics" links


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/blog/
- https://www.coronawarn.app/de/blog/

## Describe the issue

- The link collections introduced by PR https://github.com/corona-warn-app/cwa-website/pull/3338 contain multiple bad links.

1. Bad link levels. The link to Pandemic Radar for instance points to https://www.coronawarn.app/blog/2023-01-18-cwa-3-0/ instead of https://www.coronawarn.app/en/blog/2023-01-18-cwa-3-0/
1. Some links don't point to any specific blog article. These do not cause a failure, however it is questionable if this what was intended and if the link is useful if it just points to the generic blog.

![image](https://user-images.githubusercontent.com/66998419/218433565-fc456b9e-7b17-4ce0-ac4a-908df239b0aa.png)

## Steps to reproduce

1. Open https://www.coronawarn.app/en/blog/ and hover over each link in the "Other topics" section, noting if it points to a correct blog article or not.
3. `npm run test:open` and select `hybrid/check_links.cy.js` to run

Note the logged errors

```text
==================== Broken links ====================

[RESPONSE 404] http://localhost:8000/blog/2023-01-18-cwa-3-0/ on '/de/blog/'
[RESPONSE 404] http://localhost:8000/blog/2022-10-26-cwa-2-28/ on '/de/blog/'
[RESPONSE 404] http://localhost:8000/blog/2022-02-02-cwa-2-17/ on '/de/blog/'
[RESPONSE 404] http://localhost:8000/blog/2023-01-18-cwa-3-0/ on '/en/blog/'
[RESPONSE 404] http://localhost:8000/blog/2022-10-26-cwa-2-28/ on '/en/blog/'
[RESPONSE 404] http://localhost:8000/blog/2022-02-02-cwa-2-17/ on '/en/blog/'

```

You may also find that the Cypress test runs out of memory and dies.

## Suggestion

Correct the links in [src/data/blog.json](https://github.com/corona-warn-app/cwa-website/blob/master/src/data/blog.json)

- Links with no specific article
- Links with wrong level (`../../` instead of `../`)

https://github.com/corona-warn-app/cwa-website/blob/0c52ddae35073311c9041c01aa04df5e88a13865/src/data/blog.json#L21-L22

[...]

PR #3384, opened 2023-02-13
Add workflow to test links

MikeMcC399:
This PR provides an additional workflow to test links in the website. It fills a gap in assisted testing which is currently missing.

`npm test` will run a full test locally, however there is no workflow which includes Cypress tests for external links.

The workflow is submitted for manually triggering only (`workflow_dispatch`). This could be changed later to trigger it also on `pull_request` for instance.

It tests:

- anchor links
- full links, including external links

The workflow currently fails due to bad links.

---
Internal Tracking ID: [EXPOSUREAPP-](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-)

### DIFF
--- a/.github/workflows/test-links.yml
+++ b/.github/workflows/test-links.yml
@@ -1,0 +1,37 @@
+name: Test Links
+# This workflow executes a subset of Cypress tests
+# concerning
+# - anchor links
+# - external links
+
+on:
+  workflow_dispatch:
+  schedule:
+      # Run every Monday at 03:15 UTC
+    - cron: '15 3 * * 1'
+
+jobs:
+
+  test-links:
+    runs-on: ubuntu-latest
+    env:
+      # See issue https://github.com/cypress-io/cypress/issues/25357
+      ELECTRON_EXTRA_LAUNCH_ARGS: '--disable-gpu'
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v3
+
+    - name: Check anchor links
+      uses: cypress-io/github-action@v5
+      with:
+        start: npm start
+        wait-on: http://localhost:8000
+        spec: cypress/e2e/check_anchor_links.cy.js
+        browser: chrome
+
+    - name: Check external links
+      uses: cypress-io/github-action@v5
+      with:
+        install: false
+        spec: cypress/e2e/hybrid/check_links.cy.js
+        browser: chrome


MikeMcC399:
See failed workflow on https://github.com/MikeMcC399/cwa-website/actions/runs/4163735167/jobs/7204491484

MikeMcC399:
- See also issue regarding bad links https://github.com/corona-warn-app/cwa-website/issues/3383

dsarkar:
@MikeMcC399 Thanks, we will come back to you tomorrow.

larswmh:
Good morning @MikeMcC399, thanks for the PR. This would be a nice addition to our workflows. 

As you already mentioned, this currently only runs when triggered manually. We think it would be good to run this on a schedule, preferably Mondays at a time where it wouldn't consume potentially needed resources. Let's say every Monday at 3 AM. Could you add that to the workflow? We're also interested in your feedback on this, so if you disagree please let us know.

FYI: In the background, I've tried to get [multiple tests running](https://github.com/larswmh/cwa-website/actions/runs/4164943838) at the same time yesterday. 
image
This way, I think it wouldn't be much of a problem anymore if we would add this to more frequently used workflows. However, as of right now, we haven't found a good solution to keep the server from the "build" job running, so that it can be used in all tests in the "test" job. So, in the linked example, every tests builds the site again, causing every test to take an additional two minutes longer. What do you think about changing our workflows to this?

MikeMcC399:
Good morning @larswmh 

> thanks for the PR. This would be a nice addition to our workflows.

I'm happy you agree!

> As you already mentioned, this currently only runs when triggered manually. We think it would be good to run this on a schedule, preferably Mondays at a time where it wouldn't consume potentially needed resources. Let's say every Monday at 3 AM. Could you add that to the workflow? We're also interested in your feedback on this, so if you disagree please let us know.

I have added a schedule of 03:15 UTC Mondays. [GitHub documentation: Schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) recommends avoiding the start of the hour, so I made it 15 minutes past the hour that you suggested.

You should refer to [Notifications for workflow runs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs) to ensure that the right person in the organization receives any notification if this workflow fails due to link problems. I understand that the best way would be to disable and then enable the workflow after it is merged. It is quite likely that the workflow will fail from time to time due to external website issues. This is apart from failures caused by bad links in the cwa-website source.

---

I'll reply separately to the other comments.

MikeMcC399:
@larswmh 
> 
> FYI: In the background, I've tried to get [multiple tests running](https://github.com/larswmh/cwa-website/actions/runs/4164943838) at the same time yesterday. (see ^^ for workflow image) This way, I think it wouldn't be much of a problem anymore if we would add this to more frequently used workflows. However, as of right now, we haven't found a good solution to keep the server from the "build" job running, so that it can be used in all tests in the "test" job. So, in the linked example, every tests builds the site again, causing every test to take an additional two minutes longer. What do you think about changing our workflows to this?

I looked at your workflow and your matrix is running the tests independently in parallel to each other. It would be good for me to understand what you are trying to achieve with this change and what your parameters for success would be. Maybe you could separate out this discussion and put it into a new issue, a new discussion item (if discussions are not going to be killed) or a draft PR based on your proposed changes? If you look at the workflow in this PR the web is only built once, however I don't use a matrix.

The two reasons I had for keeping the [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js) out of the mainstream workflows are:

1. It is flaky and may fail due to reasons out of the control of https://www.coronawarn.app/, for example an external webserver such as RKI, Twitter, Instagram, etc. is failing.
2. It takes a significantly longer time to run compared with all the other tests. In your example it failed after running 18 minutes.

I think those reasons are still valid and that was the motivation for creating this new workflow which only tests links. This complements the existing workflows.

larswmh:
@MikeMcC399 

> I looked at your workflow and your matrix is running the tests independently in parallel to each other. It would be good for me to understand what you are trying to achieve with this change and what your parameters for success would be. Maybe you could separate out this discussion and put it into a new issue, a new discussion item (if discussions are not going to be killed) or a draft PR based on your proposed changes? If you look at the workflow in this PR the web is only built once, however I don't use a matrix.

Yes, sorry for the confusion. This is not related to this PR, I just wanted to give you some insight on what we're working on or thinking of to further improve our Cypress testing in general. 

Ideally, we would edit the `build-and-test` workflow, which is triggered on pull requests, so that it first works on the required `build` job. Once that required job succeeded, the optional jobs (one for each test using matrix) will start by making use of the build done in the previous job to save time. 
Not only would this be an improvement in total time needed to complete all tests, as they would be performed in parallel instead of back to back, but it would also give users a better understanding on what specific test went wrong on the first look.

chrome_MihZTplzlF

In my current draft I linked above, every test builds the page again. This is a problem that is also observable in the `build-and-test` that is currently in use on `master` here, as we also couldn't find a proper solution for the problem described above.

Although GitHub provides these resources for free on open source repositories, we still want to use them responsibly. Going ahead with the draft I currently have wouldn't meet these standards for us, but we're still looking out for a solution for this.

MikeMcC399:
@larswmh 

I did understand that your discussion is about a separate change, which is why I suggested to put that discussion somewhere else, not in this PR! I have some ideas about how to solve your problem.

MikeMcC399:
- Waiting for PR https://github.com/corona-warn-app/cwa-website/pull/3388 to merge before putting this PR up for review.

- Discussion about other workflow continued in https://github.com/larswmh/cwa-website/pull/1

MikeMcC399:
- Unfortunately, even when there are no link errors detected running locally, the check_links Cypress test fails on GitHub when it is run through the official Cypress GitHub action.

image

- I suspect that this is a memory problem in the check_links code, so regrettably I am closing this PR.

- I tried also using the code from https://github.com/corona-warn-app/cwa-website/pull/3389 and putting the check_links test back in again. It failed in multiple places, but not with a crash.

- I suggest to complete PR #3389 and then look at whether this could be cloned to do only check_links. In any case it looks like this could be significant effort to get check_links to work reliably on GitHub.

- In the meantime I recommend to always execute `npm run test` locally before submitting any new PR so that external links are always checked. At the moment this will have to be done manually since we don't have a good automated way of doing the check.

MikeMcC399:
- See also follow-up issue https://github.com/corona-warn-app/cwa-website/issues/3391

PR #2945, opened 2022-06-03
Add NO_COLOR for Cypress

MikeMcC399:
This PR adds the environment variable NO_COLOR to the workflow script `cypress-prod-test.yml` according to https://docs.cypress.io/guides/continuous-integration/introduction#Colors to suppress ANSI color output in Cypress logs.

Without setting NO_COLOR to true, the raw log output contains entries such as the following:

```
0m─[39m[90m─[39m[90m─[39m[90m┐[39m
2022-06-02T10:16:13.2926281Z [90m  │[39m [90mCypress:[39m        9.6.1                                                                          [90m│[39m
```

After setting the environment variable to true, the raw log output becomes readable instead:

```
2022-06-02T16:55:13.3589734Z   (Run Starting)
2022-06-02T16:55:13.3590111Z 
2022-06-02T16:55:13.3668095Z   ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
2022-06-02T16:55:13.3668623Z   │ Cypress:        9.6.1                                                                          │
2022-06-02T16:55:13.3669053Z   │ Browser:        Electron 94 (headless)                                                         │
2022-06-02T16:55:13.3669516Z   │ Node Version:   v16.13.0 (/home/runner/runners/2.291.1/externals/node16/bin/node)              │
2022-06-02T16:55:13.3670049Z   │ Specs:          6 found (applink_spec.js, blog_spec.js, check_anchor_links.js, eventRegistrati │
2022-06-02T16:55:13.3670513Z   │                 on.js, hotline_spec.js, mime.js)                                               │
2022-06-02T16:55:13.3670959Z   │ Searched:       cypress/integration/*.js                                                       │
2022-06-02T16:55:13.3671456Z   └────────────────────────────────────────────────────────────────────────────────────────────────┘
```

This should make detailed troubleshooting easier if it becomes necessary. It improves the outcome of "Download log archive" and "View raw logs" from the GitHub Action tab when viewing an individual run:

image

PR #3389, opened 2023-02-14
Update build-and-test.yml | Run tests in parallel

larswmh:
We want to edit the build-and-test workflow, which is triggered on pull requests, so that it first works on the required build job. Once that required job succeeded, the optional jobs (one for each test using matrix) will start by making use of the build done in the previous job to save time.
Not only would this be an improvement in total time needed to complete all tests, as they would be performed in parallel instead of back to back, but it would also give users a better understanding on what specific test went wrong on the first look.

image

In my current draft I linked above, every test builds the page again. This is a problem that is also observable in the build-and-test that is currently in use on master here, as we also couldn't find a proper solution for the problem described above.

Although GitHub provides these resources for free on open source repositories, we still want to use them responsibly. Going ahead with the draft I currently have wouldn't meet these standards for us, but we're still looking out for a solution for this.

---
Internal Tracking ID: [EXPOSUREAPP-14831](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14831)


larswmh:
@MikeMcC399 brought https://github.com/actions/upload-artifact & https://github.com/actions/download-artifact to my attention, which I will test in this soon. Depending on the time needed to upload / download, this could be a sufficient solution. The documentation for this can be found here https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

In addition, I've remove the long check_links test for now.

MikeMcC399:
@larswmh 

- PR https://github.com/corona-warn-app/cwa-website/pull/3394 solved the browser crash issue in PR https://github.com/corona-warn-app/cwa-website/pull/3384 however I suggest for you to make progress on this PR and then see if it makes sense to extend it to `check_links` later.

larswmh:
Uploading and downloading an artifact seems to be a good solution once good exclusions are figured out, as it would take very long without (see [manually cancelled check](https://github.com/corona-warn-app/cwa-website/actions/runs/4203892629/jobs/7293854146)). One test fails because of a path I excluded where I thought the content wouldn't be used in tests (see [check_videos](https://github.com/corona-warn-app/cwa-website/actions/runs/4204300631/jobs/7294845664))

Some test still seem to fail without giving much information about the cause, for example in [faq_link_attr](https://github.com/corona-warn-app/cwa-website/actions/runs/4204300631/jobs/7294846071) or [check_anchor_links](https://github.com/corona-warn-app/cwa-website/actions/runs/4204300631/jobs/7294845517)

MikeMcC399:
@larswmh 

I see you made some good progress! Well done!

You managed to workaround the very large size of the web by excluding the big files.

> One test fails because of a path I excluded where I thought the content wouldn't be used in tests (see [check_videos](https://github.com/corona-warn-app/cwa-website/actions/runs/4204300631/jobs/7294845664))

I suggest to move any science video files to the specifc video directory /src/assets/video. All the blog videos area hosted there, so the science videos should go there as well. That would allow the test to run.

> Some test still seem to fail without giving much information about the cause, for example in [faq_link_attr](https://github.com/corona-warn-app/cwa-website/actions/runs/4204300631/jobs/7294846071) or [check_anchor_links](https://github.com/corona-warn-app/cwa-website/actions/runs/4204300631/jobs/7294845517)

Those jobs were cancelled because the video job failed. To stop that happening, you need to add `if: always()` I believe.

larswmh:
@MikeMcC399 

> I suggest to move any science video files to the specifc video directory /src/assets/video. All the blog videos area hosted there, so the science videos should go there as well. That would allow the test to run.

I've opened up a PR for your suggestion in #3400

MikeMcC399:
Sorry, the second part was bad advice:

https://docs.github.com/en/actions/examples/using-concurrency-expressions-and-a-test-matrix

"Setting `fail-fast` to `false` prevents GitHub from cancelling all in-progress jobs if any matrix job fails."

larswmh:
Thanks for the advice @MikeMcC399. As expected, all tests are passing. This PR is now ready for review.

MikeMcC399 (changes requested):
I suggest to add

```
    env:
      NO_COLOR: true
      TERM: xterm
```

For explanation of NO_COLOR reason, see PR https://github.com/corona-warn-app/cwa-website/pull/2945

If the environment variable TERM: xterm is added it prevents a spurious warning `tput: No value for $TERM and no -T specified` from Cypress running in headless mode.

MikeMcC399 (approved):
Good improvement to increase readability of test results!

.github/workflows/build-and-test.yml
```diff
@@ -21,10 +21,29 @@ jobs:
...
+        name: public
+        retention-days: 1
+        path: public.zip
 
   test:
     needs: build
     runs-on: ubuntu-latest
+    env:
+      NO_COLOR: true
+      TERM: xterm
```
MikeMcC399, line 40: Since these environment variables are only used by Cypress, it would be more readable to add them to the step which calls Cypress at the bottom of the workflow where `ELECTRON_EXTRA_LAUNCH_ARGS` is defined.

It's an optional change. The workflow works fine the way it is.

larswmh: I moved them down to the Cypress step in https://github.com/corona-warn-app/cwa-website/pull/3389/commits/9bf3a0258a53e938eb11b94c0a3cdf574a285e51, thanks for the suggestion

brianebeling (approved):
Great work, your involvement and knowledge of Cypress really helped a lot. Thanks @MikeMcC399 and @larswmh for the improvement! I saw it in the office already.. looks amazing.

Issue #3391, opened 2023-02-15
Cypress check_links fails on GitHub and locally


MikeMcC399:
When check_links is run on GitHub it may show the error message

"We detected that the Chromium Renderer process just crashed."

and then fail.

On a local system it may fail with the Chrome "Aw, snap!" memory failure message.

## Where to find the issue

[cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js)

## Describe the issue

If an attempt is made to run [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js) in a GitHub workflow, the workflow fails.

Cypress is currently configured with version 10.11.0.

### On GitHub

For test purposes
- modify [.github/workflows/build-and-test.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build-and-test.yml) and change the test selection from `cypress/e2e/*.js` to `cypress/e2e/hybrid/*.js`
- run the workflow

### Locally

Chrome 109 on Windows 11

```bash
nvm use 18
npm ci
npm run test:open
```
select Chrome browser and `hybrid/check_links` test to run

The error message "Error code: Out of Memory" may also appear here, similar to on GitHub.
It is possible to workaround this by logging out (to kill all Chrome processes) and logging in again, however it shows that the memory usage of the test is not robust.

## Analysis

There are two failures:

1.  In the archive blog section
```text
1) "/de/blog/archiv/" - Check for broken links
1) "/en/blog/archive/" - Check for broken links
```
2. Crash after completing check of /en/sitemap and starting check of /de/blog (Check for broken links on entries)

## Cypress logs

```text

  Check for broken links
    ✓ Check if txt results exist (36ms)
    ✓ "/de" - Check for broken links (27736ms)
    ✓ "/en" - Check for broken links (24969ms)
    ✓ "/de/eventregistration/" - Check for broken links (1919ms)
    ✓ "/en/eventregistration/" - Check for broken links (1947ms)
    ✓ "/de/community/" - Check for broken links (3134ms)
    ✓ "/en/community/" - Check for broken links (1430ms)
    ✓ "/de/analysis/" - Check for broken links (8578ms)
    ✓ "/en/analysis/" - Check for broken links (6430ms)
[...]

PR #2891, opened 2022-05-13
Recommend testing with cypress in README.md

larswmh:
In #2478, @MikeMcC399 proposed the definition of Cypress testing policies. With this PR, we're adding a recommendation in the `README.me` file
image

PR #3394, opened 2023-02-16
Check only main section for blog sub-entries

MikeMcC399:
This PR improves the reliability and performance of the Cypress test [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js).

## Condition before PR

The test [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js) has been failing with a browser out-of-memory crash when it was attempting to check blog entries. This started to occur after the introduction of additional links to blog articles placed near the topic of https://www.coronawarn.app/en/blog/ and https://www.coronawarn.app/de/blog/ (see issue https://github.com/corona-warn-app/cwa-website/issues/3391).

About 50 new links were added, which is about a 50% increase in the number of links on the blog main page. These links, which duplicated links on the same page, cause their corresponding linked sub-entries to have their article contents checked a second time.

The headers and footers of the blog sub-entries are identical to each other and to the main blog page. Repeatedly checking these identical sections, which contain about 30 links each, means redundant checks were being made, and in some cases, websites such as Instagram or GitHub blocked the checks because of their rapidity and number, due to rate-limiting or security protection measures.

The four sub-entry tests `/de/blog`, `/en/blog`, `/de/science` and `/en/science` were carrying out 9817 link tests.

## Condition after PR

This PR changes the test strategy of blog and science blog entries by restricting the link check to the `main` section of each blog (sub-)entry. The `main` section is the part which makes each blog entry unique. This means that the `header` and `footer` of a blog sub-entry are not checked, since they are the same as the main blog pages.

The `header` and `footer` continue to be checked on the main blog pages.

## Comparison

The following table gives a comparison of before and after restricting the blog sub-entry test to the `main` section. The following parameters were measured on a Windows 11 system:

| Item                | time            | number of tests |
| ------------------- | --------------- | --------------- |
| Total Test duration | ~19:00 -> 10:06 | 39              |
| /de/blog            |                 | 4434 -> 794     |
| /en/blog            |                 | 4412 -> 772     |
| /de/science         |                 | 485 -> 317      |
| /en/science         |                 | 486 -> 318      |

It was difficult to get a complete performance baseline (before applying the PR) due to the test crashing. It took several attempts before the test ran without crashing.

## Result of PR

- Stops browser crash
- Removes > 7000 unnecessary link checks
- Cuts > 9 minutes off the total time of the test

## Verification

On a local clone of cwa-website

```bash
npm ci
npm run test:open
```

Select `check_links` and confirm that the test runs successfully.

```bash
npm test
```

to run the complete test suite and check that this also runs successfully.

---
Internal Tracking ID: [EXPOSUREAPP-14806](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14806)

### DIFF
--- a/cypress/e2e/hybrid/check_links.cy.js
+++ b/cypress/e2e/hybrid/check_links.cy.js
@@ -51,9 +51,9 @@ context("Check for broken links", () => {
     it(`"${page}" - Check for broken links`, () => {
       cy.visit({log: false, url: page} )
       cy.get("a:not([href*='mailto:'],[href*='tel:'],[href*='#'])").not('.email').each(url => {
-        if (url.prop('href') && !allowlist.includes(url.prop('href'))) {         
+        if (url.prop('href') && !allowlist.includes(url.prop('href'))) {
           cy.request({
-            failOnStatusCode: false, 
+            failOnStatusCode: false,
             log: false,
             url: url.prop('href')
           }).then((response) => {
@@ -65,12 +65,12 @@ context("Check for broken links", () => {
               })
             }
           })
-        }         
+        }
       })
     })
   })
 
-  
+
 })
 
 context("Check for broken links on entries", () => {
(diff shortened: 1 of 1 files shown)


larswmh (approved):
Thanks for the PR @MikeMcC399! We appreciate the effort you put into this and the much shorter testing time. Unfortunately, we didn't notice that when implementing the blog index below the search bar.

MikeMcC399:
@larswmh 

> We appreciate the effort you put into this and the much shorter testing time. Unfortunately, we didn't notice that when implementing the blog index below the search bar.

I think this PR should be OK to merge. I found it working reliably when it is running locally, which is the way it has been used in the past.

There is a set of other problems trying to get the workflow to run on GitHub, which would need to be researched separately. That should not prevent this PR from being merged though, as otherwise it will be very difficult to test external links at all.

Although the [README](https://github.com/corona-warn-app/cwa-website/blob/master/README.md#testing) includes the text: "To minimize the occurrence of errors we would ask you to perform all tests when contributing to our repository." (see PR #2891) this advice has not always been followed. That is a manual process discipline issue. The fact that the full test takes such a long time to run has been a deterrent in the past. Hopefully the test time reduction here will help to reduce this barrier to quality.

PR #3402, opened 2023-02-20
Workflow for check_links

MikeMcC399:
This PR creates a workflow `.github/workflows/test-check_links.yml` which runs the Cypress test [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js) to check hyperlinks which are not anchors.

The test is run weekly every Monday at 03:15 UTC and on demand.

The workflow does not have a `pull_request` trigger because the success of the test relies in many cases on the correct performance of external third-party websites over which cwa-website has no control. Any failure needs to be followed up manually and should not result in a CI failure of any PR. Also the test takes much longer to run than any other of the Cypress tests. (Tests in a fork showed that it typically runs for approximately 18 minutes when successful.)

If the test fails then log files are uploaded as artifacts to aid in troubleshooting.

~~It is currently draft status only. It may need some more tuning to remove any websites which prove to be too unreliable.~~

This is a follow-on from previous PRs https://github.com/corona-warn-app/cwa-website/pull/3384 and https://github.com/corona-warn-app/cwa-website/pull/3399.

image

image

---
Internal Tracking ID: [EXPOSUREAPP-14844](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14844)

PR #3404, opened 2023-02-21
Improve check_links Cypress test to block hostnames not URLs

MikeMcC399:
This PR improves the Cypress test [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js) regarding maintaining the list of link targets which cannot be checked due to their network or security setup.

In the past there were two separate lists, one for the main check and one for the blog checks. The lists were complete URLs.

## Reason for PR

The motivation for this PR is to enable this Cypress test spec to be called successfully from a GitHub workflow. Some websites which allowed testing from a local environment cannot be automatically tested from GitHub, meaning that the list was getting longer and less manageable.

## Changes

- The two lists are merged into one list.
- The list is now based on hostnames e.g. instead of the full URL `https://testbuchen.de/#/?zoom=0&lat=47.71401323721353&lng=8.66960999999999` the hostname `testbuchen.de` is used.
- If an problem is found where the host responds with an error code, the hostname is also logged to the `broken_links_result.txt` log file.
- The reason for each of the entries is added as a comment.
- The source code is also reformatted using standard VSCode settings.

## Benefits

The advantages are:

- One list is easier to maintain and avoids duplication.
- Using hostnames significantly shortens the length of entries. It also makes the list more robust, since in general it is the host server, not a particular URL which causes hard errors in attempting to test, so instead of adding multiple URLs for the same server it is now sufficient to add the one common hostname.
- Logging the hostname to the log file means that in the case of a hard error the hostname can be copied and pasted exactly to the list `noCheckHosts`.
- The comments regarding previous errors and reasons for adding the hostname to the list aid maintaining the list.

### Local test

```bash
npm run test:open
```

select `check_links` and ensure that the test runs successfully. Repeat the test twice more.

### GitHub verification

This PR will be verified as part of PR https://github.com/corona-warn-app/cwa-website/pull/3402. It has already been checked to run on GitHub.

---
Internal Tracking ID: [EXPOSUREAPP-14842](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14842)

### DIFF
--- a/cypress/e2e/hybrid/check_links.cy.js
+++ b/cypress/e2e/hybrid/check_links.cy.js
@@ -1,111 +1,106 @@
 const { softAssert, softExpect } = chai;
 
+const noCheckHosts = [
+  'developer.apple.com',       // HTTP 403 forbidden issue on GitHub
+  'ec.europa.eu',              // HTTP 502 bad gateway
+  'health.ec.europa.eu',       // HTTP 502 bad gateway
+  'onlinelibrary.wiley.com',   // HTTP 503 service unavailable
+  'support.apple.com',         // Timeout - no response
+  'testbuchen.de',             // HTTP 403 permission denied
+  'www.instagram.com',         // SSLV3_ALERT_CLOSE_NOTIFY
+  'www.t-systems.com'          // HTTP 403 forbidden issue on GitHub
+];
 
 context("Check for broken links", () => {
-  const pages = [ '/de',
-                  '/en',
-                  '/de/eventregistration/',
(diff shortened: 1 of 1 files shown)


MikeMcC399:
I ran this yesterday several times successfully and today once, so from my point of view this PR is ready for review and merge.

larswmh (approved):
Thanks for the PR @MikeMcC399!

PR #3417, opened 2023-03-06
Remove dead EcoCare link from blogs

MikeMcC399:
- This PR resolves the issue reported in https://github.com/corona-warn-app/cwa-website/issues/3415#issuecomment-1455528374.

Dead link to https://buergertest.ecocare.center/#c734 in blogs:

- https://www.coronawarn.app/de/blog/2021-05-11-how-to-rapid-test-integration/
image

- https://www.coronawarn.app/de/blog/2021-05-02-corona-warn-app-version-2-1
image

- https://www.coronawarn.app/en/blog/2021-05-02-corona-warn-app-version-2-1
image

It removes the link https://buergertest.ecocare.center/#c734 from being actively used and adds a note to say that the website is no longer in operation.

This allows `npm run test:links` and the [weekly link check](https://github.com/corona-warn-app/cwa-website/actions/workflows/test-check_links.yml) to report success again.

Issue #3415, opened 2023-02-28
Legal changes in TestV affect app and website


MikeMcC399:
## Where to find the issue

FAQs

## Regulation changes

As reported by https://www.kbv.de/html/1150_62389.php

Auf einen Blick: Das fällt weg

- Alle Tests bei Personen ohne Symptome nach Testverordnung (z.B. Bürgertests, Tests vor Krankenhaus- oder Reha-Aufenthalt)
- Bestätigungsdiagnostik nach positivem PoC-Antigentest oder Selbsttest
- Anspruch auf kostenlose Test- und Genesenenzertifikate
- Formular OEGD zur Veranlassung von SARS-CoV-2-Tests nach TestV im Labor
- Formular 10C zur Veranlassung von SARS-CoV-2-Tests im Labor bei symptomatischen Patienten (ab 1. März auf Formular 10)

and https://www.kbv.de/html/1150_62388.php

there are major changes which come into effect on March 1, 2023.

https://www.buzer.de/gesetz/14991/v292062-2023-03-01.htm shows that the following legal paragraphs from the TestV are no longer in force:

§ 1 Anspruch
§ 2 Testungen von nachweislich infizierten Personen, Kontaktpersonen und von Personen mit Voraufenthalt in Virusvariantengebieten
§ 3 Testungen von Personen nach Auftreten von Infektionen in Einrichtungen und Unternehmen
§ 4 Testungen zur Verhütung der Verbreitung des Coronavirus SARS-CoV-2
§ 4a Bürgertestung
§ 4b Bestätigende Diagnostik-Testung
§ 5 Häufigkeit der Testungen
§ 6 Leistungserbringung

## Website impact

- Multiple FAQ articles refer to the forms 10C/OEGD and to the process where test results are transferred to the app. This is no longer available.

- It seems also that the link https://www.coronawarn.app/en/faq/results/#where_can_i_get_tested and the website https://map.schnelltestportal.de/ may no longer be relevant. (I saw a notice on a test container near where I live today which said that they are closing on March 1, 2023.)

## Suggestions

Review FAQ articles to see if they are still applicable after March 1, 2023 and revise them if they are not.

brianebeling:
Hey Mike, thank you for reporting. 

We would like to update the FAQ articles and remove the outdated information where required. 

[...]

PR #3418, opened 2023-03-07
Exclude buergertest.ecocare.center from link test

MikeMcC399:
- This PR addresses the issue reported in https://github.com/corona-warn-app/cwa-website/issues/3415#issuecomment-1455528374. It is an alternative resolution to PR https://github.com/corona-warn-app/cwa-website/pull/3417.

- It adds the hostname `buergertest.ecocare.center` to the list `noCheckHosts` in [cypress/e2e/hybrid/check_links.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/hybrid/check_links.cy.js), since https://buergertest.ecocare.center/#c734 responds with a 404 (Not Found) error.

```text
$ curl -I https://buergertest.ecocare.center/#c734
HTTP/2 404
x-ua-compatible: IE=edge
content-type: text/html; charset=utf-8
date: Tue, 07 Mar 2023 08:20:57 GMT
server: Apache
```

### DIFF
--- a/cypress/e2e/hybrid/check_links.cy.js
+++ b/cypress/e2e/hybrid/check_links.cy.js
@@ -2,6 +2,7 @@ const { softAssert, softExpect } = chai;
 
 const noCheckHosts = [
   'developer.apple.com',       // HTTP 403 forbidden issue on GitHub
+  'buergertest.ecocare.center', // HTTP 404 not found - testing site removed
   'ec.europa.eu',              // HTTP 502 bad gateway
   'health.ec.europa.eu',       // HTTP 502 bad gateway
   'onlinelibrary.wiley.com',   // HTTP 503 service unavailable


MikeMcC399:
@brianebeling 

This would be a quick, easy and relatively clean fix. I don't think it is worth spending too much time on making a decision on this. There are currently more important issues to be resolved.

brianebeling:
> Brian-J. Ebeling
> 
> This would be a quick, easy and relatively clean fix. I don't think it is worth spending too much time on making a decision on this. There are currently more important issues to be resolved.

Hey Mike, its not that I'm spending a lot of time on this, more so that I don't want to be hasty, because there is a lot to do. I already noticed that leads to bad decisions.

I was also waiting for an answer from other participants, if we can make an exception updating the blogs in this case. The answer is yes, we can. I'd say we merge this PR and wait for the banner indicating EOL. Then we don't need to change the blogs at all, for now. (unless another link gets hijacked again, in this case I advocate to just remove it)

PR #3428, opened 2023-03-14
Add End of Dev FAQ Topic and Sections Outline

brianebeling:
This is an outline of what is coming to the FAQ soon.
image
image


GisoSchroederSAP (changes requested):
Please move: "General topics about the end of development..." to the first section with title "General Questions"  and respectively move "Here we will inform about functional impacts..." to the sections "Impacts"
Same switch for the German Translation: "Hier werden bald generelle Themen...publiziert" gehört zur section "Generelle Fragen" und "Hier werden wir darüber informierne welche ... Auswirkungen..." gehört zur section "Auswirkungen".  Das war nur vertauscht.

PR #3423, opened 2023-03-14
Add callouts to FAQ entries for updates

brianebeling:
Where an update text was available, there are now callouts for the updated FAQ entries
image

There are no callouts for entries, where there was no update text already provided, example:
image

There are also no callouts for texts that are structured like this: "Since version 2.5 [..], and since version 2.6" and contain a lot of text. As that would essentially make the whole entry a callout. Callouts are meant to highlight changes to the faq entry, not necessarily the app version.

(merging this soon already, but if there is feedback, please don't hesitate to write it into the already closed pr)

Issue #3416, opened 2023-03-02
EFGS shutdown unclear in FAQs


MikeMcC399:
## Where to find the issue

- https://www.coronawarn.app/en/faq/results/#interoperability_countries
- https://www.coronawarn.app/en/faq/results/#glossary_efgs
- https://www.coronawarn.app/en/faq/results/#anonymous

## Describe the issue

The use of the European Federation Gateway Service (EFGS) has been discontinued by Germany. This is not clearly reflected in the FAQs.

- https://www.coronawarn.app/en/faq/results/#interoperability_countries

In the text there is no mention that there are no longer any active countries participating. In the details however, it lists Germany (from Oct 17, 2020 until Feb 21, 2023) so Germany is no longer participating.

- https://www.coronawarn.app/en/faq/results/#glossary_efgs

"A central infrastructure that is used to exchange and check diagnosis keys across national borders."

This is no longer true. The infrastructure may still be in place, however it is not being used any more.

## Suggestions

Update the text so that it is clear that the EFGS is history as far as Germany is concerned.

## Related

https://github.com/eu-federation-gateway-service/efgs-federation-gateway

---
Internal Tracking ID: [EXPOSUREAPP-14971](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14971)

MikeMcC399:
- This issue was resolved by PR https://github.com/corona-warn-app/cwa-website/pull/3423

PR #3431, opened 2023-03-20
Update FAQ & Glossary regarding EFGS deactivation

larswmh:
This PR solves #3416. It was partially resolved in #3423, but @MikeMcC399 pointed out two more instances that needed change [here](https://github.com/corona-warn-app/cwa-website/issues/3416#issuecomment-1468417438).

[Glossary EN: European Federation Gateway Service (EFGS)](https://www.coronawarn.app/en/faq/results/#glossary_efgs)
image
[Glossary DE: European Federation Gateway Service (EFGS)](https://www.coronawarn.app/de/faq/results/#glossary_efgs)
image

[FAQ EN: How do you guarantee that I remain anonymous when using the Corona-Warn-App?](https://www.coronawarn.app/en/faq/results/#anonymous)
image
[FAQ DE: Wie ist gewährleistet, dass ich bei der Nutzung der Corona-Warn-App anonym bleibe?](https://www.coronawarn.app/de/faq/results/#anonymous)
image

---
Internal Tracking ID: [EXPOSUREAPP-14971](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14971)

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -2809,7 +2809,7 @@
                                 "anchor": "update_faq",
                                 "active": false,
                                 "textblock": [
-                                    "The last content changes to FAQ articles were made on February 21, 2023.",
+                                    "The last content changes to FAQ articles were made on March 20, 2023.",
                                     "<hr>",
                                     "The Corona-Warn-App open source team updates this FAQ regularly and keeps it up to date. You can see when the last updates were made and what changes were made in the change history on <a href='#glossary_github' target='_blank'>GitHub</a>.",
                                     "See: <a href='https://github.com/corona-warn-app/cwa-website/commits/master/src/data/faq.json' target='_blank' rel='noopener noreferrer'>GitHub: English Corona-Warn-App FAQ Change History</a>"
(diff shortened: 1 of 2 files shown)


src/data/faq.json
```diff
@@ -2809,7 +2809,7 @@
                                 "anchor": "update_faq",
                                 "active": false,
                                 "textblock": [
-                                    "The last content changes to FAQ articles were made on February 21, 2023.",
+                                    "The last content changes to FAQ articles were made on March 20, 2023.",
```
MikeMcC399, line 2812: @larswmh 

It seems that this item is often forgotten when FAQs are updated.

It is probably too late now to automate this, but one idea would be to take the last commit date from git and add it somewhere else, like into the footer.

```
$ git log -1 --pretty="format:%cs" src/data/faq_de.json
2023-03-15
```

larswmh: This is indeed often forgotten about and can arguably be seen as a bad solution for this problem in general, but I would still consider it as better than the solution before that just put the date of deployment there (regardless of whether there were changes of the FAQ or not). 

Although I'm not satisfied with the implementation either, I'd suggest to leave it as it is for the remaining timespan. As we usually do not delete FAQ articles for archiving purposes, it was agreed that this article about the latest FAQ update date should remain.

PR #3434, opened 2023-03-22
Update dashboard: Add annotations for Warning without TAN

larswmh:
This PR adds an annotation to the dashboard to show when "Warning without TAN" was implemented into the app. This will help to understand why the statistics in the lower two tiles have significantly changed after this date.

image

---
Internal Tracking ID: [EXPOSUREAPP-14711](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14711)

### DIFF
--- a/src/assets/js/analyse/chart.js
+++ b/src/assets/js/analyse/chart.js
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import ApexCharts from 'apexcharts';
 import { Subject } from 'rxjs';
+import { DateTime } from 'luxon';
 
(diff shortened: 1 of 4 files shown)


MikeMcC399:
@larswmh 
> We agree that there should be another annotation for the legal changes in TestV. We already forwarded your suggestion yesterday and got positive feedback. This will be added in a PR soon, however we would like to merge this PR now.

Many thanks for picking up my suggestion!

I would personally want to add the two annotations together, but of course it's your decision if you want to go ahead with this PR on its own.

PR #3452, opened 2023-03-29
Update FAQ #ramp_down_stores: Remove second sentence

larswmh:
This PR removes the sentence: "Whether the CWA will still be available and/or downloadable on other platforms is not known by us and is the responsibility of the respective operators of these platforms or software distribution channels." or "Ob die CWA danach noch auf anderen Plattformen heruntergeladen werden kann, ist uns nicht bekannt und liegt in der Verantwortung der jeweiligen Plattform-Betreiber." from the [#ramp_down_stores](https://www.coronawarn.app/en/faq/results/#ramp_down_stores) FAQ entry

image
image

---
Internal Tracking ID: [EXPOSUREAPP-15000](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-15000)

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -49,8 +49,7 @@
                                 "anchor": "ramp_down_stores",
                                 "active": false,
                                 "textblock": [
-                                    "The CWA will be available in the app stores of Apple and Google until May 31, 2023.",
-                                    "Whether the CWA will still be available and/or downloadable on other platforms is not known by us and is the responsibility of the respective operators of these platforms or software distribution channels."
+                                    "The CWA will be available in the app stores of Apple and Google until May 31, 2023."
                                 ]
                             },
                             {
--- a/src/data/faq_de.json
+++ b/src/data/faq_de.json
@@ -49,8 +49,7 @@
                                 "anchor": "ramp_down_stores",
                                 "active": false,
                                 "textblock": [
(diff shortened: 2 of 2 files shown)


MikeMcC399:
@larswmh 

> This PR removes the sentence: "The CWA will be available in the app stores of Apple and Google until May 31, 2023." or "Die Corona-Warn-App wird noch bis einschließlich 31. Mai 2023 in den Appstores von Apple und Google verfügbar sein." from the [#ramp_down_stores](https://www.coronawarn.app/en/faq/results/#ramp_down_stores) FAQ entry

It seems that the PR leaves only that sentence and removes the following text :

"Whether the CWA will still be available and/or downloadable on other platforms is not known by us and is the responsibility of the respective operators of these platforms or software distribution channels."

So it looks to me like the change is correct although the description does not match the change.

larswmh:
@MikeMcC399 

> It seems that the PR leaves only that sentence and removes the following text :
> 
> "Whether the CWA will still be available and/or downloadable on other platforms is not known by us and is the responsibility of the respective operators of these platforms or software distribution channels."
> 
> So it looks to me like the change is correct although the description does not match the change.

Thanks for the information. The description should now be correct.

dsarkar (approved):

PR #3454, opened 2023-03-30
Update FAQ: Add note about previous free test entitlement

larswmh:
This PR adds a note to the FAQ where a free test entitlement is mentioned. This was based on the German TestV, which is not in effect anymore. As of right now, this only affects the FAQ entry #rat_differences_pcr

image
image

---
Internal Tracking ID: [EXPOSUREAPP-15004](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-15004)


MikeMcC399:
@larswmh 

- Thanks for adding this PR! It resolves part of the set of issues raised in https://github.com/corona-warn-app/cwa-website/issues/3415.

PR #3455, opened 2023-03-30
Update #ramp_down_apk: Small text and title changes

larswmh:
This PR changes the English FAQ title of #ramp_down_apk to match with the German title by removing the word "still". In addition to that, the last sentence was slightly changed. See https://github.com/corona-warn-app/cwa-website/pull/3455#issuecomment-1490214683

image
image

### DIFF
--- a/src/data/faq.json
+++ b/src/data/faq.json
@@ -61,7 +61,7 @@
                                 ]
                             },
                             {
-                                "title": "Will reproducible builds/APKs still be provided? Is a community-operated backend supported?",
+                                "title": "Will reproducible builds/APKs be provided? Is a community-operated backend supported?",
                                 "anchor": "ramp_down_apk",
                                 "active": false,
                                 "textblock": [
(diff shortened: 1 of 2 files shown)


MikeMcC399:
@larswmh 

Would you also change the grammar in the last sentence?

"Please note, support for this setup will not be provided by BMG and RKI."

That should be 

"Please note, support for this setup will not be provided by BMG or RKI."

It would only be "and" if they were providing support. Since they are not offering support, then "or" is the better choice.

I think the German text with "und" is OK however, since it is worded differently.

larswmh:
@MikeMcC399 

Thanks for your suggestion, I have changed it in https://github.com/corona-warn-app/cwa-website/pull/3455/commits/b2dc762f489633ce2917734e967f8b15842a79bd for both the English and German text. We think this wouldn't make it worse in German.

dsarkar (approved):

PR #3465, opened 2023-04-20
Update FAQ: Remove link to rapid test portal

larswmh:
This PR removes the links to the rapid test portal, following its shutdown stated in boxes above relevant FAQ articles implemented in [here](https://github.com/corona-warn-app/cwa-website/pull/3464). For archiving purposes, we will be leaving the links as text. This way, the Cypress link tests won't fail in the future. See image below for an example.
image

---
Internal Tracking ID: [EXPOSUREAPP-14580](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14580)


MikeMcC399:
@larswmh 

You were too quick for me to comment!

For any hosts which should no longer be checked, you could add them to the list:

https://github.com/corona-warn-app/cwa-website/blob/b38ba0130d60c9ec0e7a32bd477a3c4f5ae71b75/cypress/e2e/hybrid/check_links.cy.js#L3-L15

larswmh:
@MikeMcC399 

> You were too quick for me to comment!
> 
> For any hosts which should no longer be checked, you could add them to the list:
> 
> https://github.com/corona-warn-app/cwa-website/blob/b38ba0130d60c9ec0e7a32bd477a3c4f5ae71b75/cypress/e2e/hybrid/check_links.cy.js#L3-L15

Sorry for the quick merge. I think this particular case wouldn't fit into that list, because it is about to be shut down completely. The ones in the list are generally accessable, but are occassionally causing errors in testing. 

Looking back at it, I might have not worded it understandable in the opening comment. If it was just about the testing, the list would've been the more fitting alternative. 

The original thought was to remove the link (with the text) completely, but with the goal of archiving as best as possible in mind, this seemed like a good trade-off for us. Readers could surely just copy & paste the URL text and check for themselves, but I think the solution implemented in this PR will be way less misleading once it is actually being shutdown.

MikeMcC399:
@larswmh 

I don't suggest that you change anything now. I think the changed article is fine.

I just wanted to point out the alternative if it helps in the future.

>  I think this particular case wouldn't fit into that list, because it is about to be shut down completely. The ones in the list are generally accessable, but are occassionally causing errors in testing.

The list is a mixture, including sites which no longer operate, for example:

```text
'corona-schnelltest-zentren.dm.de', // HTTP 404 not found - testing site removed 
```

Issue #3479, opened 2023-04-29
Flaky Cypress test faq_link_attr.cy.js


MikeMcC399:
Cypress test [cypress/e2e/faq_link_attr.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/faq_link_attr.cy.js) is sometimes failing when running on GitHub using the workflows:

- [.github/workflows/build-and-test.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/build-and-test.yml) or
- [.github/workflows/cypress-test-prod.yml](https://github.com/corona-warn-app/cwa-website/blob/master/.github/workflows/cypress-test-prod.yml)

```text
  1) Check attributes of FAQ links
       Check links are using _blank to open in new frame
         Test links for _blank presence:
     CypressError: Timed out after waiting `100000ms` for your remote page to load.

Your page did not fire its `load` event within `100000ms`.

You can try increasing the `pageLoadTimeout` value in `cypress.config.js` to wait longer.

Browsers will not fire the `load` event until all stylesheets and scripts are done downloading.

When this `load` event occurs, Cypress will continue running commands.
      at timedOutWaitingForPageLoad (http://localhost:8000/__cypress/runner/cypress_runner.js:137295:72)
      at <unknown> (http://localhost:8000/__cypress/runner/cypress_runner.js:138269:16)
      at tryCatcher (http://localhost:8000/__cypress/runner/cypress_runner.js:8914:23)
      at <unknown> (http://localhost:8000/__cypress/runner/cypress_runner.js:4036:41)
      at tryCatcher (http://localhost:8000/__cypress/runner/cypress_runner.js:8914:23)
      at Promise._settlePromiseFromHandler (http://localhost:8000/__cypress/runner/cypress_runner.js:6849:31)
      at Promise._settlePromise (http://localhost:8000/__cypress/runner/cypress_runner.js:6906:18)
      at Promise._settlePromise0 (http://localhost:8000/__cypress/runner/cypress_runner.js:6951:10)
      at Promise._settlePromises (http://localhost:8000/__cypress/runner/cypress_runner.js:7027:18)
      at _drainQueueStep (http://localhost:8000/__cypress/runner/cypress_runner.js:3621:12)
      at _drainQueue (http://localhost:8000/__cypress/runner/cypress_runner.js:3614:9)
[...]

PR #3480, opened 2023-04-29
Fix timeout on faq_link_attr.cy.js

MikeMcC399:
This PR adds a workaround to [cypress/e2e/faq_link_attr.cy.js](https://github.com/corona-warn-app/cwa-website/blob/master/cypress/e2e/faq_link_attr.cy.js) to resolve https://github.com/corona-warn-app/cwa-website/issues/3479.

Before each sub-test the root `\` of the website under test is visited. This seems to work around a timeout issue where web contents was sometimes not displayed.

This is a problem with the website, not with the test, however it is now unlikely that there will be a fix for the website, so the workaround is added to the Cypress test.

### DIFF
--- a/cypress/e2e/faq_link_attr.cy.js
+++ b/cypress/e2e/faq_link_attr.cy.js
@@ -7,6 +7,11 @@ describe("Check attributes of FAQ links", () => {
     const faqResultsUrl = '/faq/results/';
     const languages = ["en", "de"];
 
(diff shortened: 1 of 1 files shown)


larswmh (approved):
Thanks for the PR! Looks good to me

dsarkar:
@MikeMcC399 Thanks!

PR #3490, opened 2023-05-04
Fix failing cypress test regarding faq-links and enable logging

thorbenkuro:
This Pull Request fixes the cypress test, that was causing the GitHub Action, that runs the e2e tests, to fail. The root cause for the failing test seemed to be the forEach loop, that iterated through every link and softExpected the link to have a certain attribute. The test repeatedly timed out at this point, most likely because of the CPU/RAM resources of the GitHub Actions Runner. I have replaced the forEach-loop with an additional filter, and a check for the length of the filtered list, to work-around the stated issue.

I also added the ability to log to the visible console output, to be able to find the root cause of any upcoming issues quicker.

---
Internal Tracking ID: [EXPOSUREAPP-15135](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-15135)


mtb77 (approved):

MikeMcC399:
- This PR fixes https://github.com/corona-warn-app/cwa-website/issues/3479

PR #3526, opened 2023-06-01
One more e-mail-address change

Ein-Tim:
## Description 

This PR changes one more instance where the SAP mail address has been used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ The following channels are available for discussions, feedback, and support requ
 | Type                     | Channel                                                |
 | ------------------------ | ------------------------------------------------------ |
 | **General discussion, issues, bugs**   | <a href="https://github.com/corona-warn-app/cwa-website/issues/new/choose" title="General Discussion"><img src="https://img.shields.io/github/issues/corona-warn-app/cwa-website?style=flat-square"></a> </a>   |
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@Ein-Tim 

What about
https://github.com/corona-warn-app/cwa-website/blob/28ff46ebccdc52c535a0e07dac8c16dafc9547d8/package.json#L30

  "author": "SAP Corona-Warn-App Open Source Team <corona-warn-app.opensource@sap.com>",

Ein-Tim:
@MikeMcC399 Thanks for the hint!

As I'm not familiar with how the information in this file is used, I will need advice from @dsarkar here. As a reminder: We also decided to not change https://github.com/corona-warn-app/cwa-website#licensing, so maybe this is also ok for the package.json file?

MikeMcC399:
@Ein-Tim

According to https://docs.npmjs.com/creating-a-package-json-file#author-field
"email and website are both optional"

so I suggest either to remove the sap e-mail address (since it is no longer reachable) or to completely replace the author by RKI with their e-mail address.

The licensing is OK not to be changed.

dsarkar:
@Ein-Tim As @MikeMcC399 suggests:
- just remove email address and keep the rest

dsarkar (approved):

PR #3529, opened 2023-06-01
Feature/technical hotline shutdown

dsarkar:
- remove hotline cypress test
- remove hotline test from workflow
- remove hotline from footer
- remove entry in [TESTING.md](https://github.com/corona-warn-app/cwa-website/blob/master/docs/TESTING.md)


MikeMcC399:
@dsarkar 

There is also a line in https://github.com/corona-warn-app/cwa-website/blob/master/docs/TESTING.md which could be removed

image

PR #3567, opened 2024-05-08
Feature/website discontinued

svengabr:
This PR will replace the landing pages of the website with a notice that the website was shut down.

In addition to that, all sub-pages except legal pages were also removed from the website.


Ein-Tim (approved):
Tested & works!

